PR #75, opened 2019-09-27
Making lots of changes to get building on OSX to work.

swolter:
I had some time to get this working. It builds and runs now on my machine with XCode 10.2 and OSX 10.14.5. I don't really know CMake, so sorry if it's a bit of a mess. 

The only interesting changes were using Boost filesystem when std's isn't available, passing some hints for GLFW so the window could be created, and I had to modify the cursor fragment shader to remove gl_FragColor. I also set GLFW_COCOA_GRAPHICS_SWITCHING to true so it doesn't force a switch to the discrete GPU while it's running.


christianparpart:
Hey. Many thanks for your contribution. I really respect this. It is currently 08 AM here and I am going to have time to look into it at about 08 PM tonight (including streaming on twitch ;)

PR #108, opened 2020-05-29
Fixing mac ci

uspasojevic96:


.github/workflows/osx.yml
```diff
@@ -20,7 +20,7 @@ jobs:
         run: |
           set -ex
           brew update
-          brew install freetype fontconfig qt5 harfbuzz boost pkg-config
+          brew install freetype fontconfig glew glfw3 glm qt5 harfbuzz boost pkg-config
```
christianparpart, line 23: Why are these 3 new deps needed? (did I forget that somewhere? it shouldn't be needed anymore)

uspasojevic96: I figured to add it because it's in README.md

uspasojevic96: Removed instances of them from yml and readme

PR #128, opened 2020-07-20
Update CODING_STYLE.md

3n16m4:
add 'east const' coding style rule and example

### DIFF
--- a/docs/CODING_STYLE.md
+++ b/docs/CODING_STYLE.md
@@ -23,7 +23,7 @@ namespace org::coding_style::naming_conventions
 {
     void eastConst() {
         int const a = 42;        // a is const
-        int const *p = &a;       // value in p is const, p is not const.
+        int const* p = &a;       // value in p is const, p is not const.
         int const *const p = &a; // both value and p are const.
     }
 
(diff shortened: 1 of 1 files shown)


christianparpart:
Hey. Many thanks. I will give details ASAP. See you soon.

docs/CODING_STYLE.md
```diff
@@ -14,12 +14,19 @@
...
+- east const instead of west const
 
 ### Example
 
 ```cpp
 namespace org::coding_style::naming_conventions
 {
+    void eastConst() {
+        int const a = 42;        // a is const
+        int const *p = &a;       // value in p is const, p is not const.
```
christianparpart, line 26:
```suggestion
        int const* p = &a;       // value in p is const, p is not const.
```

docs/CODING_STYLE.md
```diff
@@ -14,12 +14,19 @@
...
+        int const *const p = &a; // both value and p are const.
```
christianparpart, line 27:
```suggestion
        int const* const p = &a; // both value and p are const.
```

3n16m4: seems reasonable to me, the asterisk operator should be tied to the variable.

PR #135, opened 2020-07-25
Add SGR 9, crossed out, decoration renderer.

gitabaz:
I've added the crossed out feature. The only part that was missing was the actual decoration renderer.
Tested using shell command
`echo -e "Hello \e[9mCrossed\e[0m Out"`

### DIFF
--- a/src/terminal_view/DecorationRenderer.cpp
+++ b/src/terminal_view/DecorationRenderer.cpp
@@ -203,7 +203,23 @@ void DecorationRenderer::rebuild()
             move(image)
         );
     } // }}}
-    // TODO: CrossedOut
+    { // {{{ crossed-out
+        auto const middleCell = screenCoordinates_.cellHeight / 2;
+        auto const thickness = max(lineThickness_ * baseline / 3, 1u);
+        auto image = atlas::Buffer(width * middleCell, 0u);
+
+        for (unsigned y = 0; y < thickness; ++y)
+            for (unsigned x = 0; x < width; ++x)
+                image[y * width + x] = 0xFF;
+
+        atlas_.insert(
+            Decorator::CrossedOut,
+            width, middleCell - thickness / 2,
+            width, middleCell - thickness / 2,
+            GL_RED,
+            move(image)
+        );
+    } // }}}
     // TODO: Framed
     // TODO: Encircle
 }


christianparpart:
Oh many thanks!

src/terminal_view/DecorationRenderer.cpp
```diff
@@ -203,7 +203,23 @@ void DecorationRenderer::rebuild()
...
+        auto const thickness = max(lineThickness_ * baseline / 3, 1u);
+        auto image = atlas::Buffer(width * middleCell, 0u);
+
+        for (unsigned y = 0; y < thickness; ++y)
+            for (unsigned x = 0; x < width; ++x)
+                image[y * width + x] = 0xFF;
+
+        atlas_.insert(
+            Decorator::CrossedOut,
+            width, middleCell,
```
christianparpart, line 217:
```suggestion
            width, middleCell - thickness / 2,
```

src/terminal_view/DecorationRenderer.cpp
```diff
@@ -203,7 +203,23 @@ void DecorationRenderer::rebuild()
...
+            width, middleCell,
```
christianparpart, line 218:
```suggestion
            width, middleCell - thickness / 2,
```

gitabaz:
> Hey,
> 
> many thanks for your PR. I know it wasn't much to do (I didn't remember how much however).
> But it's also a good starter to get comfortable with the source code - I think.
> 
> The only change I'd suggest is that the middle line, to be truely middle, the starting pixels should be `middle - thickness / 2`, otherwise the middle line would sit right above the middle axis.
> 
> p.s.: Framed should be very easy, too. Encircled would require some research or at least a little more mathematical challenge to make it look "good". After that I can also guide you through something slightly more challenging, if you'd want to 1st_place_medal

Thanks, for the suggestion. I've update the code and squashed them into 1 commit.

As a side note: Could you clarify what the difference between the two width and heights are in
```
atlas_.insert(
            Decorator::CrossedOut,
            width, height,
            width, height,
            GL_RED,
            move(image)
        );
```

christianparpart:
The first pair are the actual dimensions and the second one are the dimensions when their are to be rendered.

PR #136, opened 2020-07-26
Add SGR 51, framed.

gitabaz:
Implements SGR 51 and draws a frame around each glyph. Tested using shell command
`echo -e "Hello \e[51mFramed\e[54m" and not`
Currently the result looks like in the image below where each letter is surrounded by a box.
Screenshot from 2020-07-26 11-04-29

Is this the desired result or should it be more like
```
      ----------
Hello | Framed | and not
      ----------
``` 
I'm not sure how to do the second version.


christianparpart:
Many thanks for your contribution.

The second could be achieved using cell attributes grouping. I do that already to optimize performance for background renders as well as for proper text shaping. 

But I think the first version is right. Although it looks like the lines are too thick? What's the pixel line width?

gitabaz:
> Many thanks for your contribution.
> 
> The second could be achieved using cell attributes grouping. I do that already to optimize performance for background renders as well as for proper text shaping.
> 
> But I think the first version is right. Although it looks like the lines are too thick? What's the pixel line width?

I think this is due to drawing the right line of 1 box and the left line of the adjacent box. The thickness of the lines are
`auto const thickness = max(lineThickness_ * baseline / 3, 1u);`

christianparpart:
Hey,

I managed to make it look *slightly* more interesting:

image

with the following diff:

```diff
diff --git a/src/terminal_view/DecorationRenderer.cpp b/src/terminal_view/DecorationRenderer.cpp
index f119c94..cb0ca6b 100644
--- a/src/terminal_view/DecorationRenderer.cpp
+++ b/src/terminal_view/DecorationRenderer.cpp
@@ -188,20 +188,21 @@ void DecorationRenderer::rebuild()
     } // }}}
     { // {{{ framed
         auto const cellHeight = screenCoordinates_.cellHeight;
-        auto const thickness = max(lineThickness_ * baseline / 3, 1u);
+        auto const thickness = max(lineThickness_ * width / 20, 1u);
         auto image = atlas::Buffer(width * cellHeight, 0u);
+        auto const gap = thickness;
 
         // Draws the top and bottom horizontal lines
-        for (unsigned y = 0; y < thickness; ++y)
-            for (unsigned x = 0; x < width; ++x)
+        for (unsigned y = gap; y < thickness + gap; ++y)
+            for (unsigned x = gap; x < width - gap; ++x)
             {
                 image[y * width + x] = 0xFF;
                 image[(cellHeight - 1 - y) * width + x] = 0xFF;
             }
-        
+
         // Draws the left and right vertical lines
-        for (unsigned y = 0; y < cellHeight; y++)
-            for (unsigned x = 0; x < thickness; ++x)
+        for (unsigned y = gap; y < cellHeight - gap; y++)
+            for (unsigned x = gap; x < thickness + gap; ++x)
             {
                 image[y * width + x] = 0xFF;
```

The smaller the font, the lesser obvious it is that those frames are disconnected. I don't think that's an issue at all, because I think that one would want to make only a single cell framed, and multiple connected ones at once (unless it is for TUI debugging or so? :D) 

You can apply that diff on top of your PR.

christianparpart:
Wow nice, you are still awake. If you feel ready, just move out of draft mode.

src/terminal_view/DecorationRenderer.cpp
```diff
@@ -232,7 +233,23 @@ void DecorationRenderer::rebuild()
             move(image)
         );
     } // }}}
-    // TODO: CrossedOut
+    { // {{{ crossed-out
```
christianparpart: I think this came in by accident.

gitabaz: I was adding the framed feature while working on a branch from the old master (before crossed-out was added).
So I'm trying to get my master updated with the crossed out feature and then combine that with the framed feature.

gitabaz: I'm having trouble squashing these 2 commits into 1 since it contains a merge. Perhaps this wasn't the best way of doing things?

christianparpart: Do you happen to remember how you did end up in this state? Just so I understand. Did you do something like: `git merge master` while being in your working branch? I think the right command should have been `git fetch && git rebase origin/master`, ideally.

I think I can provide you a small video (or session tomorrow morning on Twitch) on how to unfuck the situation.

gitabaz: Yes I did a git merge master while in the working branch. Then I had merge conflict on the DecorationRenderer.cpp file. So I opened it up and just deleted the lines that looked like `<<<<<<< HEAD`, `=======`, and `>>>>>>>` so the file looked correct. Then applied your diff `git apply patch.diff` and pushed.

christianparpart:
We've merged it behind the wall of Github.

PR #158, opened 2021-01-07
set min Size of window so it does not crash when the width is near zero

### LINKED ISSUE #157: Crash/freeze when width is extremely low e.g near zero
## Description

When the windows is made really small near zero pixel it just crashes. 
On master I get this error messages "105406 floating point exception (core dumped)". When I test is on wip branch then the window just freezes. 

## Environment

- Contour Version or Git commit hash: (e14560a8bb250109610815898397cd07c2ca6019) wip, (40e766263ec680995c48f626d6b3063b53a06d75) master
- Operating System (name / distribution / version): Manjaro (lastest kernal 5.10.4)
- Contour configuration: default
- Compiler version: gcc 10

## Steps to Reproduce

1. Resize width to the lowest it can get
2. contour crashes/freezes

ResuBaka:
The problem is when the width is near zero then the cell size width will be zero and when that happens there are multible places that can not handle that case.

So to prevent I just have set the minimum Size of the window. 

I don't know it that is the best way to handle the problem, but from my view it is the fastest way to fix it.

Fixes #157.

### DIFF
--- a/src/contour/TerminalWidget.cpp
+++ b/src/contour/TerminalWidget.cpp
@@ -671,6 +671,7 @@ void TerminalWidget::resizeGL(int _width, int _height)
     auto const viewHeight = height();
 
     terminalView_->resize(viewWidth, viewHeight);
+    setMinimumSize(terminalView_->cellWidth() * 3, terminalView_->cellHeight() * 2);
     terminalView_->setProjection(
         ortho(
             0.0f, static_cast<float>(viewWidth),      // left, right


christianparpart:
Oh nice!  I will have a look at soon as possible. Many thanks already

src/contour/Controller.cpp
```diff
@@ -52,6 +52,7 @@ void Controller::newWindow()
         programPath_
     };
     mainWindow->show();
+    mainWindow->setMinimumSize(50, 50);
```
christianparpart, line 55: Why 50? We could use the grid cell size here.

ResuBaka: I have just taken something pretty small that still works.

Is the grid cell size a fixed value of pixel or something that changes base on like zoom level? 

When it is the first one I think I can make that change my self, but for the second case I would a bit of help in it or you need to implement it then in the right place.

uspasojevic96: Does window size take into consideration window decorations from DE?

christianparpart: I think it should be set on the widget (`TerminapWidget`). Then this question should not matter. Can you try this?

Also, the grid cell size is the size in pixels for one single grid cell. The width is the width of one column, and the height is the height of the one line. All in pixels. HiDPI content scaling is already applied. Do you only need to address these attributes and use these as minimum widget hints.

Writing from Smartphone, i hope it helps so far. And many thanks. :)

ResuBaka: Only using setMinimumSize even when we do it in TerminapWidget::resizeGL we run into problems. 

One is we can not directly use the cell size for it. As we always need some more space so a cell size times 3 works in most cases, but when we then start to zoom/resize the font. We run into the same problem/problems. 

We could first use this as a workaround for the time being, maybe with the change to use the cell size for the limit.


The other would be to fix each problem that  needed to fixed so that you can resize the screen you want to.

I know of at least one place where a check needed to be done so that you don't run into a "integer divide by zero" error. An other one looks like a cursor render problem ([line of code that then fails](https://github.com/christianparpart/contour/blob/master/src/terminal/Screen.cpp#L467)). 

So should I push the code for the more or less still just a workaround for resize of the window or how do we want handle this?

Just so the code that I currently have locally would be this:
```c++

void TerminalWidget::resizeGL(int _width, int _height)
{
#if !defined(NDEBUG)
    cout << fmt::format(
        "resizeGL: {}x{}, geometry: {}/{}\n",
        _width, _height,
        terminal::Size{geometry().top(), geometry().left()},
        terminal::Size{geometry().width(), geometry().height()});
#endif

    if (_width == 0 || _height == 0)
        return;

    scrollBar_->resize(scrollBar_->sizeHint().width(), contentsRect().height());
    switch (config_.scrollbarPosition)
    {
        case config::ScrollBarPosition::Left:
            scrollBar_->move(0, 0);
            break;
        case config::ScrollBarPosition::Right:
            scrollBar_->move(size().width() - scrollBar_->sizeHint().width(), 0);
            break;
        case config::ScrollBarPosition::Hidden:
            break;
    }

    auto const viewWidth = width() - scrollBar_->sizeHint().width();
    auto const viewHeight = height();

    terminalView_->resize(viewWidth, viewHeight);
    // Set minimum size so that we always can render the cursor
    setMinimumSize(terminalView_->cellWidth() * 3, terminalView_->cellHeight() * 2);
    terminalView_->setProjection(
        ortho(
            0.0f, static_cast<float>(viewWidth),      // left, right
#if defined(LIBTERMINAL_VIEW_NATURAL_COORDS)
            0.0f, static_cast<float>(viewHeight)      // bottom, top
#else
            static_cast<float>(viewHeight), 0.0f      // bottom, top
#endif
        )
    );

    // if (setScreenDirty())
    //     update();
}
```

ResuBaka: I have changed it to be based on the cell Width and Height and moved it into the TerminalWidget.

christianparpart (approved):
I've tested it and it works great now. Many thanks!

p.s.: minimum row/column count is subjective anyways, so 3x2 is fine I think. Also no need to make that configurable.

Issue #168, opened 2021-02-19
Implement a method to access the buffer rendered inside the current view for fuzzy finding


whisperity:
## Abstract

Have a way (similar to `tmux capture-pane`) to fetch the currently visible text of the current (or any other?) Contour "viewport" (not sure if that is the right word) for text processing.

## Motivation

One of my colleagues created a [Zsh script for FZF](http://github.com/martong/dotfiles/blob/46734e2a3f31b130163a9090a91a6bde1005396e/home/.fzf_tmux_hist.zsh) that through TMux (using the `tmux capture-pane` feature) allows him to do a fuzzy find on either the lines or the words rendered in the current viewport.

I've personally never used this feature as I rarely use TMux locally, but it is *somewhat* useful e.g. when you do something in the terminal and want to reuse a word (e.g. a filename?) or a line from it in a subsequent command without having to reach for the mouse.

Now, that script I linked, and the general idea, requires TMux which he uses extensively, but I don't.
However, we have an amazing terminal emulator at hand which could give us a similar "captured buffer view" for scripts to wrangle. This might depend on #149.

## Specification

No idea. Something like `contour --eval <running_session_id> --get-buffer` (where the session ID is exposed in an env var #167)  could throw back a huge textual buffer (either with all escape sequences or optionall some sort of a view of the current render only, with the escape sequences omitted) on the standard output, which then could be processed by the client. Check my colleague's script, he fetches textual data from `tmux` and then does Bash/Zsh magic on it.

## Additional context?

FWIW, I've *tried* to do something similar with my day-to-day terminal emulator of choice, Konsole. In Konsole, you can set the scrollback buffer between "None", "Fixed size n lines" and "Unlimited". From these options, "None" is obviously useless for the purposes of this issue. "Fixed size" is also useless, because then the backing data is only in RAM, and Konsole exposes no way of access to it. "Unlimited" *might* be useful as it creates the scrollback buffer at `/tmp`... which is a binary file. That was the point I gave up.

~~~~
╰─ fd 'konsole' /tmp | head -n 1 | xargs hexyl | head -n 25
┌────────┬─────────────────────────┬─────────────────────────┬────────┬────────┐
│00000000│ 63 00 00 00 01 00 00 00 ┊ 01 01 00 00 01 55 64 00 │c000•000┊••00•Ud0│
│00000010│ 00 00 01 00 00 00 01 01 ┊ 00 00 01 9c 20 00 00 00 │00•000••┊00•× 000│
[...]

PR #175, opened 2021-02-25
Adds screen buffer capture VT sequence and VT sequence for requesting window/text area's character size.

christianparpart:
Implements #168.

## Checklist

- [x] `contour capture output -` writes screen capture back to stdout (for use with pipes). Maybe allow this only if stout is a pipe.
- [x] Ensure very large screen buffers can be captured, too.
- [x] Add mode to retrieve logical lines (unwrapped lines) instead - update spec respectively.
- [x] the CLI option `-h` and `--help` should also print information about this feature.
- [x] backend logic implementation
- [x] feature detection via `DA1` (identification number `314`)
- [x] frontend hooking into the screen capture event
- [x] permission system
- [x] unit tests for backend logic
- [x] changelog entry
- [x] command line program to capture the current terminal's screen buffer.
- [x] Split `CSI 14 t` VT sequence addition into its own commit.
- [x] Verify permission check integration is working!

## CLI Usage

```txt
    contour capture
        Captures the screen buffer of the currently running terminal.

        Options:

            [logical]        Tells the terminal to use logical lines for counting and capturing.
            [timeout FLOAT]  Sets timeout seconds to wait for terminal to respond. [default: 1]
            [lines UINT]     The number of lines to capture [default: 0]
            output STRING    Output file name to store the screen capture to.
```

## Specification

This feature is purely VT sequence based, so it also works across remote connections.

### Sending the screen capture request:

```sh
CSI > Pm ;  Pc t
```
* **Pm:** how to count the lines, 0 raw mapping, 1 for logical line mapping (unwrapping wrapped lines into a single logical one).
* **Pc:** the number of lines to capture from the bottom of the main page area, counting upwards, up until the history top is reached. If omitted, main page area will be captured.

### Response

The response will be received over `stdin` from the terminal in the given format:

```sh
OSC 314 ; <payload> ST
```

With `<payload>` being the textual data that was saved during screen capture.
The response is chunked into multiples of the above sequence until the `<payload>` does not carry any data.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -24,6 +24,7 @@
 - Adds terminal identification environment variables `TERMINAL_NAME`, `TERMINAL_VERSION_TRIPLE` and `TERMINAL_VERSION_STRING`.
 - Adds config option `profile.*.fonts.TYPE.weight: WEIGHT` and `profile.*.fonts.TYPE.slant: SLANT` options (optional) along with `profile.*.fonts.TYPE.family: STRING`.
 - Adds VT sequence `CSI 18 t` and `CSI 19 t` for getting screen character size. Responds with `CSI 8 ; <columns> ; <rows> t` and  `CSI 9 ; <columns> ; <rows> t` respectively.
+- Adds VT sequence to capture the current screen buffer `CSI > LineMode ; StartLine ; LineCount t` giving the respone back on stdin via `OSC 314 ; <screen buffer> ST`, and feature detection via `DA1` number `314`.
 
 ### 0.1.1 (2020-12-31)
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -23,6 +23,7 @@
 - Adds config option `profile.*.fonts.TYPE.weight: WEIGHT` and `profile.*.fonts.TYPE.slant: SLANT` options (optional) along with `profile.*.fonts.TYPE.family: STRING`.
 - Adds terminal identification environment variables `TERMINAL_NAME`, `TERMINAL_VERSION_TRIPLE` and `TERMINAL_VERSION_STRING`.
 - Adds config option `profile.*.fonts.TYPE.weight: WEIGHT` and `profile.*.fonts.TYPE.slant: SLANT` options (optional) along with `profile.*.fonts.TYPE.family: STRING`.
+- Adds VT sequence `CSI 18 t` and `CSI 19 t` for getting screen character size. Responds with `CSI 8 ; <columns> ; <rows> t` and  `CSI 9 ; <columns> ; <rows> t` respectively.
 
 ### 0.1.1 (2020-12-31)
 
--- a/TODO.md
+++ b/TODO.md
@@ -1,8 +1,8 @@
 # Milestone 0.2.0 checklist
 
-- [ ] CI: use cpack to build linux zip's (and maybe deb files?)
-- [ ] CI: ensure installers contain right version + git sha + prerelease suffix
-- [ ] CI: Fix ubuntu 18.04 dependencies
+- [ ] FIXME: `reset` resets screen size to 80x25, should remain actual one.
+- [ ] debuglog: filter by logging tags (in a somewhat performant way), so the debuglog (when enabled) is not flooding.
+- [ ] CopyLastMarkRange seems not to work (at least for double-line prompts in zsh/p10k)
 
 - [ ] Font: support DirectWrite backend
 - [ ] Font: fix framed underline
--- a/src/contour/CMakeLists.txt
+++ b/src/contour/CMakeLists.txt
@@ -23,6 +23,7 @@ CIncludeMe(contour.yml "${CMAKE_CURRENT_BINARY_DIR}/contour_yaml.h" "default_con
 set(contour_SRCS
     Actions.cpp Actions.h
     BackgroundBlur.cpp BackgroundBlur.h
+    CaptureScreen.cpp CaptureScreen.h
     Config.cpp Config.h
     Controller.cpp Controller.h
     FileChangeWatcher.cpp FileChangeWatcher.h
--- a/src/contour/CaptureScreen.cpp
+++ b/src/contour/CaptureScreen.cpp
@@ -1,0 +1,341 @@
+/**
+ * This file is part of the "libterminal" project
+ *   Copyright (c) 2019-2020 Christian Parpart <christian@parpart.family>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include <contour/CaptureScreen.h>
+
+#include <crispy/utils.h>
+
+#include <fmt/format.h>
+
+#include <algorithm>
+#include <cstdlib>
+#include <fstream>
+#include <functional>
+#include <iostream>
+#include <memory>
+#include <optional>
+#include <ostream>
+#include <string>
+#include <string_view>
+#include <tuple>
+
+#if defined(_WIN32)
+    #include <Winsock2.h>
+    #include <Windows.h>
+#else
+    #include <sys/select.h>
+    #include <termios.h>
+    #include <unistd.h>
+    #include <fcntl.h>
+#endif
+
+#if !defined(STDIN_FILENO)
+    #define STDIN_FILENO 0
+#endif
+
+using std::cerr;
+using std::copy_n;
+using std::cout;
+using std::make_unique;
+using std::nullopt;
+using std::ofstream;
+using std::optional;
+using std::ostream;
+using std::reference_wrapper;
+using std::stoi;
+using std::string;
+using std::string_view;
+using std::tuple;
+using std::unique_ptr;
+
+using namespace std::string_view_literals;
+
+namespace contour {
+
+namespace
+{
+    struct TTY
+    {
+        bool configured = false;
+#if !defined(_WIN32)
+        int fd = -1;
+        termios savedModes{};
+#else
+        DWORD savedModes{};
+#endif
+
+        ~TTY()
+        {
+#if !defined(_WIN32)
+            tcsetattr(fd, TCSANOW, &savedModes);
+#else
+            auto stdinHandle = GetStdHandle(STD_INPUT_HANDLE);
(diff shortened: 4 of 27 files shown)


Changelog.md
```diff
@@ -20,6 +20,8 @@
 - Adds config option `profile.*.fonts.TYPE.weight: WEIGHT` and `profile.*.fonts.TYPE.slant: SLANT` options (optional) along with `profile.*.fonts.TYPE.family: STRING`.
 - Adds terminal identification environment variables `TERMINAL_NAME`, `TERMINAL_VERSION_TRIPLE` and `TERMINAL_VERSION_STRING`.
 - Adds config option `profile.*.fonts.TYPE.weight: WEIGHT` and `profile.*.fonts.TYPE.slant: SLANT` options (optional) along with `profile.*.fonts.TYPE.family: STRING`.
+- Adds VT sequence `CSI 18 t` and `CSI 19 t` for getting screen character size. Responds with `CSI 8 ; <columns> ; <rows> t` and  `CSI 9 ; <columns> ; <rows> t` respectively.
+- Adds VT sequence to capture the current screen buffer `OSC 314 ; Mode ; StartLine ; LineCount ST` giving the respone back on stdin via `OSC 314 ; <screen buffer> ST`, and feature detection via `DA1` number `314`.
```
whisperity, line 24: Response on STDIN?

whisperity:
Will this allow going back in history too? Like, as if someone scrolled up?

Changelog.md
christianparpart: yes, so in the shell you can use the `read` command and in C/C++ the `int read(int fd, char* targetBuf, size_t bufSize)` syscall with `fd` being `STDIN_FILENO` (= `1`).

christianparpart:
> Will this allow going back in history too? Like, as if someone scrolled up?

Yes, with mode=1 you just have access to the main page area (no history) and mode=0 you must specify absolute line coordinates, where startLine=0 is the oldest line in history (so on the very top if you scroll up), startLine=1 the second oldest one, and so on. If you have `historyLines` number of lines in history scrollback buffer and have `startLine=historyLines + 1` then you address the first line in the main page area (just like mode=1 and startLine=1).

whisperity:
Minor nit: this will prevent calling a normal system-wide binary named `capture` because if `argv[1]` is capture it's hard-coded to do this feature, not the normal "execute the specified command in the terminal" kind of stuff.

Suggestion: `contour -X capture`. And this `-X` could be the flag under which all these extra features are implemented. I shamelessly stole the idea from Clang where `-X` is used for "argument forwarding" from the top-level invocation to a lower-level one, skipping the argument's parsing on the top-level.

whisperity:
So does this capture the sequences now? I've run a dummy call and it seems to work nicely.

~~~
╰─ cat FILE.txt 
───────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: FILE.txt
───────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ ╭─ ✔   ~                                                                                            1.37    10:00:56 
   2   │ ╰─ contour capture -o FILE.txt
   3   │ 
   4   │ 
   5   │ 
   6   │ 
   7   │ 
   8   │
   9   │
  10   │
  11   │
  12   │
  13   │
  14   │
  15   │
  16   │
  17   │
  18   │
  19   │
  20   │
  21   │
  22   │
  23   │
  24   │
  25   │
  26   │
  27   │
  28   │
  29   │
  30   │
  31   │
  32   │
───────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
~~~

The hex of the result looks like this:

~~~
╰─ hexyl FILE.txt
┌────────┬─────────────────────────┬─────────────────────────┬────────┬────────┐
│00000000│ e2 95 ad e2 94 80 20 e2 ┊ 9c 94 20 ee 82 b0 20 ef │×××××× ×┊×× ××× ×│
│00000010│ 80 95 20 7e 20 ee 82 b0 ┊ 20 20 20 20 20 20 20 20 │×× ~ ×××┊        │
│00000020│ 20 20 20 20 20 20 20 20 ┊ 20 20 20 20 20 20 20 20 │        ┊        │
│*       │                         ┊                         │        ┊        │
│00000070│ 20 20 ee 82 b2 20 31 2e ┊ 33 37 20 ef 82 80 20 20 │  ××× 1.┊37 ×××  │
│00000080│ ee 82 b2 20 31 30 3a 30 ┊ 30 3a 35 36 20 ef 80 97 │××× 10:0┊0:56 ×××│
│00000090│ 0a e2 95 b0 e2 94 80 20 ┊ 63 6f 6e 74 6f 75 72 20 │_×××××× ┊contour │
│000000a0│ 63 61 70 74 75 72 65 20 ┊ 2d 6f 20 46 49 4c 45 2e │capture ┊-o FILE.│
│000000b0│ 74 78 74 0a 0a 0a 0a 0a ┊ 0a 0a 0a 0a 0a 0a 0a 0a │txt_____┊________│
│000000c0│ 0a 0a 0a 0a 0a 0a 0a 0a ┊ 0a 0a 0a 0a 0a 0a 0a 0a │________┊________│
│000000d0│ 0a 0a                   ┊                         │__      ┊        │
└────────┴─────────────────────────┴─────────────────────────┴────────┴────────┘
~~~

whisperity:
The `capture_buffer: ask` does not work yet, the capture happens no matter what. What's more, if the user is running an older configuration, the loading isn't lazy, and Contour will simply not start if the `permissions.capture_buffer` key is missing from the config file:

~~~
╰─ contour
Configuration failure. Unhandled error caught. invalid node; first invalid key: "capture_buffer"
~~~

christianparpart:
> The `capture_buffer: ask` does not work yet, the capture happens no matter what. What's more, if the user is running an older configuration, the loading isn't lazy, and Contour will simply not start if the `permissions.capture_buffer` key is missing from the config file:
> 
> ```
> ╰─ contour
> Configuration failure. Unhandled error caught. invalid node; first invalid key: "capture_buffer"
> ```

I realized that tonight. That is why this bullet point is still unchecked in there checklist above :)

christianparpart:
> Minor nit: this will prevent calling a normal system-wide binary named `capture` because if `argv[1]` is capture it's hard-coded to do this feature, not the normal "execute the specified command in the terminal" kind of stuff.
> 
> Suggestion: `contour -X capture`. And this `-X` could be the flag under which all these extra features are implemented. I shamelessly stole the idea from Clang where `-X` is used for "argument forwarding" from the top-level invocation to a lower-level one, skipping the argument's parsing on the top-level.

Common practice however is too use `--` to indicate that any following parameter is too be treated verbatim. Maybe this?

christianparpart:
> > Minor nit: this will prevent calling a normal system-wide binary named `capture` because if `argv[1]` is capture it's hard-coded to do this feature, not the normal "execute the specified command in the terminal" kind of stuff.
> > Suggestion: `contour -X capture`. And this `-X` could be the flag under which all these extra features are implemented. I shamelessly stole the idea from Clang where `-X` is used for "argument forwarding" from the top-level invocation to a lower-level one, skipping the argument's parsing on the top-level.
> 
> Common practice however is too use `--` to indicate that any following parameter is too be treated verbatim. Maybe this?

Just verified. You can achieve that via:

```sh
contour -- capture
```

if you want to spawn the process named `capture` via contour (however, usually you pass a fully qualified path).

christianparpart:
Also, this feature cannot be supported on windows until ConPTY implements https://github.com/microsoft/terminal/issues/1173.

whisperity:
> Also, this feature cannot be supported on windows until ConPTY implements [microsoft/terminal#1173](https://github.com/microsoft/terminal/issues/1173).

Does this mean the patch will not go in, or will just Windows users simply get no result, or an error message, if they try to call the special function?

whisperity:
The `mode` setting is not yet wired out to the `contour capture` interface, however. At least wasn't in build 216.

whisperity:
I've tried modifying the script my colleague made to call `contour capture` instead of `tmux capture-pane`, but the captured file seems to always remain empty. Maybe calling the command from inside a subshell or script is a problem? Because if I type the command manually into the command-line, the capture works.

christianparpart:
> Does this mean the patch will not go in, or will just Windows users simply get no result, or an error message, if they try to call the special function?

No, this PR **will** be merged regardless of that fact. I'll have to find a workaround (maybe using WinPTY or looking at how mlterm (and maybe others?) for example is providing more sophisticated VT sequences to the windows terminal land for local processes. But that will then be a topic that needs a whole PR on its own (to generally provide true VT sequence pass-through to contour on Windows.

> The `mode` setting is not yet wired out to the `contour capture` interface, however. At least wasn't in build 216.

It is, via the `-a` flag which tells the capture command to treat the start line position to be an absolute value (0 := oldest history line (scroll-top), 1 := second  oldest line, and so on). If `-a` is not passed, then the start line position is relative to the main page area, with line 1 being the oldest entry. -- OMG, while writing this comment I realize that i should use the same offset-start everywhere, either 0-based or 1-based. I'll change the history addressing to be 1-based too....

> Upgraded to build 234! I've tried modifying the script my colleague made to call `contour capture` instead of `tmux capture-pane`, but the captured file seems to always remain empty. Maybe calling the command from inside a subshell or script is a problem? Because if I type the command manually into the command-line, the capture works.

I'm not so sure about how you invoke this `contour capture ....` but if you do redirect stdin and/or stdout of this process, then it cannot communicate to and with the terminal emulator. Maybe this is the problem? Maybe you can share your edited script on Discord or so, and I can try myself locally?

whisperity:
I'm trying something very simple:

```zsh
local tempfile=$(mktemp)
contour capture -o ${tempfile} 0 200
cat ${tempfile}
```

Also, if I try `contour capture | cat` it'll show empty, together with an error message saying "cannot get current screen size" but I'm not sure if it's contour or cat saying that.

`>file` also stays empty.

But if I just type in `contour capture -o foo.txt`, it writes into the file.

christianparpart:
> I'm trying something very simple:
> 
> ```shell
> local tempfile=$(mktemp)
> contour capture -o ${tempfile} 0 200
> cat ${tempfile}
> ```
> 
> Also, if I try `contour capture | cat` it'll show empty, together with an error message saying "cannot get current screen size" but I'm not sure if it's contour or cat saying that.
> 
> `>file` also stays empty.
> 
> But if I just type in `contour capture -o foo.txt`, it writes into the file.

I did test your above, and it worked for me (bad joke/quote, I know). But if you want to just capture everything, just do this:

```sh
a=`mktemp`
contour capture -o $file -a
```

Now `$file` should contain your terminal history + main page area.

The reason why `contour capture ... > file` does not work is because stdin/stdout must both be pointing to the connected terminal, i.e. not redirected to a pipe, file, or similar.

PR #181, opened 2021-02-27
Feature/build archlinux binary

ResuBaka:
Add Workflow to build an archlinux binary.

Currently it would build vor x86 and arm64, there we could remove the arm part for now when you do not want or let it in here but as a test build wenn someone wanted to test the arm64 version.

What is still open:
* add desktop file
* add icon/logo


ResuBaka:
One thing I want to do is rewrite the history a bit so it cleaner for every body.

christianparpart:
Wow man, i really very much appreciate you doing this for me. Many thanks. I am looking forward to review it whenever you are ready.

ResuBaka:
This should now be ready to review. 

I could check release workflow too so for releases there are builds too or should that be done on a later stage?

christianparpart:
> This should now be ready for review.
> 
> I could check release workflow too so for releases there are builds too or should that be done on a later stage?

Absolutely. This would be really nice for the arch users that wasn't to pick a release archive instead. :+1:

ResuBaka:
I have pushed how a release should work. 

How could that be tested or does it looks like it should work?

ResuBaka:
When you wanted to have code to create PKGBUILD files. For the contour-bin you would need to drop the **-RUN_ID** or add it as the version to the aur package or I guess what could be done is with seed replace like "RUN_ID" with the run id.
 That would be up to you. The main thing is that this wuould need to run after the release as it updates the version in the git case and in the bin version the sha256sums for you.

> What is not in this patch are the changes that would be needed for the CI. 

```patch
diff --git a/3rdparty/libunicode b/3rdparty/libunicode
index a440d09..7983ce3 160000
--- a/3rdparty/libunicode
+++ b/3rdparty/libunicode
@@ -1 +1 @@
-Subproject commit a440d093e3ef57c03a10ea73dd499794e2726a02
+Subproject commit 7983ce340c4574bcde66ee7aad8e8de384116a89
diff --git a/archlinux/Dockerfile b/archlinux/build/Dockerfile
similarity index 89%
rename from archlinux/Dockerfile
rename to archlinux/build/Dockerfile
index f8c3e49..e539912 100644
--- a/archlinux/Dockerfile
+++ b/archlinux/build/Dockerfile
@@ -10,4 +10,4 @@ RUN pacman -Su --noconfirm gcc ninja cmake extra-cmake-modules qt5-base fontconf
 
 RUN ls -la
 
-RUN ./archlinux/dist
+RUN ./archlinux/build/dist
diff --git a/archlinux/dist b/archlinux/build/dist
similarity index 100%
rename from archlinux/dist
rename to archlinux/build/dist
diff --git a/archlinux/release/Dockerfile b/archlinux/release/Dockerfile
new file mode 100644
index 0000000..43564c6
--- /dev/null
+++ b/archlinux/release/Dockerfile
@@ -0,0 +1,22 @@
+FROM lopsided/archlinux:latest
+
+ARG VERSION
+
+WORKDIR /app
+
+COPY . .
+
+RUN pacman -Sy --noconfirm archlinux-keyring
+
+RUN pacman -Su --noconfirm gcc ninja cmake extra-cmake-modules qt5-base fontconfig harfbuzz git base-devel
+
+RUN ls -la
+
+RUN groupadd -g 1000 build && useradd -u 1000 -g 1000 -s /bin/sh -m build
+
+RUN chown -R 1000:1000 archlinux/release
+
+USER 1000:1000
+
+RUN cd archlinux/release && ./update-contour.sh ${VERSION}
+
diff --git a/archlinux/release/contour-bin/PKGBUILD b/archlinux/release/contour-bin/PKGBUILD
new file mode 100644
index 0000000..6f631a4
--- /dev/null
+++ b/archlinux/release/contour-bin/PKGBUILD
@@ -0,0 +1,26 @@
+pkgname=contour-bin
+pkgver=0.1.1
+pkgrel=2
+pkgdesc="Contour - A modern C++ Terminal Emulator"
+arch=('x86_64' 'aarch64')
+url="https://github.com/christianparpart/contour"
+license=('Apache-2.0')
+depends=(
+  'extra-cmake-modules'
+  'qt5-base'
+  'harfbuzz'
+)
+conflicts=('contour-git')
+
+source_x86_64=("https://github.com/christianparpart/contour/releases/download/v${pkgver}/${pkgname/-bin/}_${pkgver}-archlinux_x86_64.tar.gz")
+source_aarch64=("https://github.com/christianparpart/contour/releases/download/v${pkgver}/${pkgname/-bin/}_${pkgver}-archlinux_aarch64.tar.gz")
+
+sha256sums_x86_64=('1bc8f55e8d4b7358fc3206365c909a9f649fddd2a5bb9962d2743207951b5f49')
+sha256sums_aarch64=('86a67bac677d3bc3099ceb766f62ab1383641a01824ae17de897f17ce6c78c82')
+
+package() {
+  _output="${srcdir}/${pkgname/-bin/}_${pkgver}_${CARCH}"
+  install -Dm755 "${_output}/${pkgname/-bin/}" "${pkgdir}/usr/bin/${pkgname/-bin/}"
+  install -D -m644 "${_output}/contour-logo.png" "${pkgdir}/usr/share/pixmaps/contour.png"
+  install -D -m644 "${_output}/contour.desktop" "${pkgdir}/usr/share/applications/contour.desktop"
+}
\ No newline at end of file
diff --git a/archlinux/release/contour-git/.SRCINFO b/archlinux/release/contour-git/.SRCINFO
new file mode 100644
index 0000000..e172fc8
--- /dev/null
+++ b/archlinux/release/contour-git/.SRCINFO
@@ -0,0 +1,22 @@
+pkgbase = contour-git
+	pkgdesc = Contour - A modern C++ Terminal Emulator
+	pkgver = 0.1.1.r0.ga726870
+	pkgrel = 1
+	url = https://github.com/christianparpart/contour
+	arch = x86_64
+	license = Apache-2.0
+	makedepends = gcc
+	makedepends = ninja
+	makedepends = cmake
+	makedepends = extra-cmake-modules
+	makedepends = qt5-base
+	makedepends = fontconfig
+	makedepends = harfbuzz
+	depends = extra-cmake-modules
+	depends = qt5-base
+	depends = harfbuzz
+	source = contour::git+https://github.com/christianparpart/contour#tag=v0.1.1
+	sha256sums = SKIP
+
+pkgname = contour-git
+
diff --git a/archlinux/release/contour-git/PKGBUILD b/archlinux/release/contour-git/PKGBUILD
new file mode 100644
index 0000000..477ec87
--- /dev/null
+++ b/archlinux/release/contour-git/PKGBUILD
@@ -0,0 +1,51 @@
+pkgname=contour-git
+_pkgname=contour
+pkgver=0.1.1.r0.ga726870
+pkgrel=1
+pkgdesc="Contour - A modern C++ Terminal Emulator"
+arch=('x86_64')
+url="https://github.com/christianparpart/contour"
+license=('Apache-2.0')
+depends=(
+  'extra-cmake-modules'
+  'qt5-base'
+  'harfbuzz'
+)
+makedepends=(
+  'gcc'
+  'ninja'
+  'cmake'
+  'extra-cmake-modules'
+  'qt5-base'
+  'fontconfig'
+  'harfbuzz'
+)
+source=("contour::git+https://github.com/christianparpart/contour#tag=v0.1.1")
+sha256sums=('SKIP')
+conflicts=('contour-bin')
+
+pkgver() {
+  cd "$srcdir/$_pkgname"
+  git describe --long --tags | sed 's/^v//;s/\([^-]*-g\)/r\1/;s/-/./g'
+}
+
+build() {
+  export cpuCount=$(awk '/^processor/{n+=1}END{print n}' /proc/cpuinfo)
+  cd "$srcdir/$_pkgname"
+  git submodule update --init --recursive
+  mkdir -p build
+  cd build && cmake .. \
+            -GNinja \
+            -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_COMPILER=g++ \
+            -DYAML_BUILD_SHARED_LIBS=OFF -DYAML_CPP_BUILD_CONTRIB=OFF \
+            -DYAML_CPP_BUILD_TESTS=OFF -DYAML_CPP_BUILD_TOOLS=OFF \
+            -DYAML_CPP_INSTALL=OFF
+  ninja -j$cpuCount
+}
+
+package() {
+  cd "$srcdir/$_pkgname"
+  install -D -m755 build/src/contour/contour "${pkgdir}"/usr/bin/contour
+  install -D -m644 src/contour/res/images/contour-logo.png "${pkgdir}"/usr/share/pixmaps/contour.png
+  install -D -m644 src/contour/contour.desktop "${pkgdir}"/usr/share/applications/contour.desktop
+}
diff --git a/archlinux/release/update-contour.sh b/archlinux/release/update-contour.sh
new file mode 100755
index 0000000..8d3a20f
--- /dev/null
+++ b/archlinux/release/update-contour.sh
@@ -0,0 +1,39 @@
+#!/usr/bin/bash
+
+# Usage Function
+function usage() {
+  cat <<_EOT_
+Usage:
+  $0 version ...
+Description:
+  version tells what pkgver to update to
+_EOT_
+  exit 1
+}
+
+if [ $# = 0 ]; then
+  usage
+fi
+
+newV="$1"
+set -e
+echo "Processing contour-git"
+(
+  cd contour-git || exit
+  makepkg
+  makepkg --printsrcinfo >.SRCINFO
+)
+
+echo "Processing contour-bin"
+(
+  cd contour-bin || exit
+  # git checkout .
+  git clean -fdx
+  # git pull
+  sed -i "s/pkgver=.*/pkgver=$newV/g" PKGBUILD
+  updpkgsums PKGBUILD
+  makepkg
+  makepkg --printsrcinfo >.SRCINFO
+)
+#
+#git commit -a -S -m "$newV"
```

christianparpart:
Hey, many thanks so far. I am busy with family but i will drop a review ASAP. :)

.github/workflows/build.yml
```diff
@@ -298,3 +298,73 @@ jobs:
         if-no-files-found: error
         retention-days: 30
 
+  archlinux:
+    strategy:
+      matrix:
+        arch:
+          [
+            "linux/amd64 x86_64",
+            "linux/arm64/v8 aarch64",
```
christianparpart, line 307: I really like that you are testing more than one architecture. I should do the same for the other OSes too :)

.github/workflows/build.yml
```diff
@@ -298,3 +298,73 @@ jobs:
...
+      uses: docker/setup-buildx-action@v1
+      with:
+        version: latest
+    - name: Read info
+      id: tags
+      shell: bash
+      run: |
+        arch="${{ matrix.arch }}"
+        echo ::set-output name=PLATFORM::${arch%% *}
+        echo ::set-output name=ARCH::${arch##* }
```
christianparpart, line 328: Could these two env vars be maybe also inferred via `uname` too? If so they could be exported via the `set-vars.sh` script, too.
`uname -m` yields `x86_64` for me and `uname -o` yields `GNU/Linux`

.github/workflows/build.yml
```diff
@@ -298,3 +298,73 @@ jobs:
...
+      id: set_vars
+      run: ./scripts/ci-set-vars.sh
+      env:
+        REPOSITORY: ${{ github.event.repository.name }}
+    - name: Build ${{ matrix.arch }} release
+      shell: bash
+      run: |
+        arch="${{ matrix.arch }}"
+        PLATFORM=${arch%% *}
+        ARCH=${arch##* }
```
christianparpart, line 339: `ARCH` should also be available via `steps.tags.outputs.ARCH`, analogous for `PLATFORM`.

.github/workflows/build.yml
```diff
@@ -298,3 +298,73 @@ jobs:
...
+        .
+        docker create --name contour-$ARCH contour:$ARCH
+        docker cp contour-$ARCH:/app/contour.tar.zst contour-$ARCH.tar.zst
+        docker container rm contour-$ARCH
+    - name: "test: libcrispy"
+      shell: bash
+      run: |
+        arch="${{ matrix.arch }}"
+        PLATFORM=${arch%% *}
+        ARCH=${arch##* }
```
christianparpart, line 354: Also here, and also in the task below. Maybe we can reduce the repetitive lines with a little helper script that these run tasks are simply calling? The script could be living in `/scripts/ci-SOME_NAME.sh`.

.github/workflows/release.yml
```diff
@@ -226,11 +226,80 @@ jobs:
...
+        id: buildx
+        uses: docker/setup-buildx-action@v1
+        with:
+          version: latest
+      - name: Read info
+        id: tags
+        shell: bash
+        run: |
+          arch="${{ matrix.arch }}"
+          echo ::set-output name=PLATFORM::${arch%% *}
```
christianparpart, line 255: FYI: The above suggestions apply here and below too. :-)

.github/workflows/release.yml
```diff
@@ -226,11 +226,80 @@ jobs:
...
+        shell: bash
+        run: |
+          arch="${{ matrix.arch }}"
+          PLATFORM=${arch%% *}
+          ARCH=${arch##* }
+          docker run --rm contour:$ARCH build/src/terminal/terminal_test
+      - name: "Uploading Archlinux .tar.zst package"
+        uses: actions/upload-artifact@v2
+        with:
+          name: contour_${{ steps.set_vars.outputs.version }}-${{ steps.set_vars.outputs.RUN_ID }}-archlinux_${{ steps.tags.outputs.ARCH }}
```
christianparpart, line 294: With my limited knowledge of Arch Linux and my rusty VM I once prepared, how do I actually test-install such an `.tar.zstd` package the recommended way? (Googling for that didn't quite yield some satisfying results right now) :)

archlinux/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM lopsided/archlinux:latest
```
christianparpart, line 1: While the release.yml for Debian packages still uses the project root's `/debian/` directory, that one is going away very soon (thanks for reminding me), as I want to reduce redundancies in package creation by using the same logic for release builds that has been used for regular test builds.

Since this directory will be solely used by the CI, what about moving this directory into `/.github/archilinux/` instead?

archlinux/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+WORKDIR /app
+
+COPY . .
+
+RUN pacman -Sy --noconfirm archlinux-keyring
+
+RUN pacman -Su --noconfirm gcc ninja cmake extra-cmake-modules qt5-base fontconfig harfbuzz git base-devel
+
+RUN ls -la
```
christianparpart, line 11: I think this line can be removed.

archlinux/dist
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
+set -o errexit -o pipefail
+
+export cpuCount=$(awk '/^processor/{n+=1}END{print n}' /proc/cpuinfo)
```
christianparpart, line 4:
```suggestion
cpuCount=`grep -c -w ^processor /proc/cpuinfo`
```

Don't `export` unless this env var is needed in sub processes. And avoid `$()` by using backticks unless you need a fully featured subshell inside.

archlinux/dist
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
+set -o errexit -o pipefail
```
christianparpart, line 2: Why did you do this instead of `set -e`?

archlinux/dist
```diff
@@ -0,0 +1,24 @@
...
+
+export cpuCount=$(awk '/^processor/{n+=1}END{print n}' /proc/cpuinfo)
+git submodule update --init --recursive
+mkdir -p build
+cd build && cmake .. \
```
christianparpart, line 7: I think this can be simply split into two lines, one for `cd build` and the other for `cmake ...`, as these commands are all always expected to succeed.

christianparpart (changes requested):
* Will such a package contain the runtime dependency information, so that users get those auto-installed, too? If now, is it possible to be added?
* How do I locally test (install) such an archive, given that I have an Arch Linux at hand?
* FYI: wrt your one question on how to test release, that's what I use the `gh-actions` branch for - up until now. I can do that soon myself (I hope).

.github/workflows/build.yml
ResuBaka: It could be posible I have just taken the ci code from [paru](https://github.com/Morganamilo/paru/blob/master/.github/workflows/build.yml) and I would say not as we do not run on that architect at that point but are using docker cross architect builds for it. So in there we could get those values from.

.github/workflows/release.yml
ResuBaka: This is just are tar that uses Zstandard for the compression so you can just unpack it and try to run the bin in there. 

This file is not for installing it that is what the PKGBUILD files would be fore in the our. The tar.zstd are just the Fileformat that is used for the packages that are in the normal Repositorys they changed it last year to use [zstd from xz](https://archlinux.org/news/now-using-zstandard-instead-of-xz-for-package-compression/).


When I would guess as when you look at the [content for example](https://mirrors.edge.kernel.org/archlinux/community/os/x86_64/aarch64-linux-gnu-gcc-10.2.0-1-x86_64.pkg.tar.zst) event the ones from arch them self they are just tar files and get extractet and pacman then does the rest.


So for having one that can just be installed you need to have the output of the patch they I send as they producte a file like I send in the example.

archlinux/Dockerfile
ResuBaka: Yea I will move it in there as that sounds like a good plan.

archlinux/dist
ResuBaka: I did not now the part of why not to use $() thanks for that. I will remove the export then and change it use backticks.

archlinux/dist
ResuBaka: Look at the dist from [this repo](https://github.com/Morganamilo/paru/blob/master/dist) it basicly just a copy, just with the changes to build countour, instead of rust code.

archlinux/dist
ResuBaka: Will change that to make it cleaner.

.github/workflows/build.yml
ResuBaka: So move this from below into a script and then use it for all other places so we have just one place where we do it? As these values should then be possible to access on every following task I would guess? I did not use github ci so much, that is why I ask.
```
        arch="${{ matrix.arch }}"
        echo ::set-output name=PLATFORM::${arch%% *}
        echo ::set-output name=ARCH::${arch##* }    
```

christianparpart:
Exactly.

ResuBaka:
> * Will such a package contain the runtime dependency information, so that users get those auto-installed, too? If now, is it possible to be added?

That packages that are currently build do not have the info as we do not build packages that are directly for install but for use as our packages. Packages that can directly be used could also be build with the patch I posted before as they have then as a result a fully ready package with all the info that pacman needs, but that is not needed for aur as there a user downloads that PGKBUILD for them selfs and creates a package that gets installed and/or installed directly from the dwnloaded zip/tar/... .

> * How do I locally test (install) such an archive, given that I have an Arch Linux at hand?

You just need to check the content if the files in there are fully working e.g does the contour bin works or not. The only problem for like ubuntu could be that libc is newer for arch. When there should be a full test then you need to make a PGKFILE or change the bin version from the PATCH a bit so it can create a result from an artifact, there I could help when needed and that could then be installed with `pacman -U /path/to/package/package_name-version.pkg.tar.zst`.

> * FYI: wrt your one question on how to test release, that's what I use the `gh-actions` branch for - up until now. I can do that soon myself (I hope).

So I can just create that branch and see if it works, will test that then later. For how you can test it when you do not want to create an arch linux install, you can just test if from the docker image when you just want to test the install or create like a Manjaro or Endeavour OS VM to test it in.

archlinux/dist
ResuBaka: Changed it.

archlinux/dist
ResuBaka: Changed it.

archlinux/dist
ResuBaka: Changed it.

.github/workflows/build.yml
ResuBaka: When I am currently working on this change does it event work and to just to it for the two echo lines it worth it? As the `arch="${{ matrix.arch }}"` needs to be outside I would guess or am i wrong about it?

christianparpart:
Hey good morning. I'm reviewing the modifications soon.

.github/archlinux/build
```diff
@@ -0,0 +1,34 @@
+#!/bin/sh
```
christianparpart, line 1: First things first: I am very sorry for the late response.

Shell scripts should have the file ending `.sh`.

.github/workflows/build.yml
christianparpart: I think there's no need to torture you. I can't await playing around with Arch Linux soon myself, too. :)

christianparpart (approved):
With the only minor thing being the one file rename, I am gladly doing that myself. Many thanks for your contribution, @ResuBaka.

.github/workflows/build.yml
christianparpart: ugh, the answer was yes. since I already merged now, I'll do that on next thursday/friday stream, I hope :)

ResuBaka:
I see I forgot a changelog entry to you add them for me?

PR #186, opened 2021-03-02
[cmake] Fix regex escaping in Version.cmake

avemilia:
```
According to [1], string literals in CMake cannot contain '\', they need
to be escaped. Therefore, if we want to pass '\' to string(REGEX), we
need to use '\\'.

[1]: <https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#quoted-argument>
```


avemilia:
NB: I actually don't know why it worked before, because in my understanding it shouldn't.

christianparpart:
Many thanks for having spotted this @avemilia. :)

PR #187, opened 2021-03-03
[cmake] "unsubmodule" 3rdparty dependencies

avemilia:
```
Summary:

* unsubmodule fmtlib, yaml-cpp, libunicode, catch2
* remove git dependency
* all: add code to select system versions of libraries
* fmtlib: behavior change: remove -DFMT_USE_WINDOWS_H=0: fmt no longer
includes windows.h [1]
* yaml-cpp: add inline patch to:
    * turn off unneeded compilation targets such as tests
    * silence warnings when compiling contour code
* yaml-cpp: attempt to download only a single header, attempt failed,
reverted to downloading full tarball
* catch2: download only the header [2]
* all: add dependency on `patch` binary [3][4]
* all: pin dependencies [5]
* libunicode: temporary inline patch to make it work with this patch

[1]: <https://github.com/fmtlib/fmt/commit/253d63159f79762ebd40aea0c20704cd29acb3c9>

[2]: compilation times can be further improved by creating separate .cpp
file and defining main() there

[3]: perhaps temporary, for the lack of other ideas how to patch
dependencies. A dependency-free way would be to include complete files
that can be hot-swapped with cmake -E copy. However, this requires
careful following of patched dependencies development and manual
backporting "by eye", whereas if patch doesn't apply, it just fails. On
the other hand, it is just as easy to fetch a `patch` binary from the
Internet as it is to fetch any tarball dependency. Googling around,
there is even a native `patch` port for Windows. Not implemented in this
PR due to lack of working Windows machine to test, left as a TODO in
code

[4]: another incovenience of `patch` is that you must apply the patch
exactly once. This is the reason behind guarding it with an if() clause.
Since the patches are provided verbatim and inline, when developer
updates the patch, the clause doesn't update, which means the developer
needs to remember to completely remove the binary dir and cleanly
reconfigure CMake for the new patch to apply. This can come as
unexpected, therefore needs to be addressed, so leaving this as another
TODO for now

[5]: all dependencies (except libunicode) are pinned and
hashsum-checked. This is good for reproducible builds. If cutting edge
dependencies are desired often, such functionality can be added later.
Current implementation does no overly complicated abstraction to support
all possible cases.

Details:

This patch is aimed towards unbundling contour dependencies from git
submodules, or in other words, "unsubmoduling" them. This is a desired
change for a few reasons, as discussed on discord:

* git submodules are _git_ submodules, that is, they require git

* because they require git, tarball archives of code, which github
provides, are useless -- user needs to populate the submodules manually,
which requires git. And guess who downloads tarballs? Definitely not the
people who know how to use git and have it installed.

* even when you clone the repo with git, you need to clone recursively,
or if you didn't read the docs/forgot, you first need to realize that
this the actual problem (since the error message will be bogus), and
then you need to google "git sumbmodule populate", because who the hell
remembers how to do it. This is unpleasant experience from the get-go.
It can be fixed in CMake by probing and populating submodules at
configure time. But it's still inferior to the idea of having no hard
dependency on git.

Therefore, this patch attemtps to replace dependency management via
submodules functionality on best-effort basis.

One such limitation is having a hard dependency on CMake 3.10, which
does not have FetchContent module (CMake 3.11). FetchContent does
exactly what this patch tries to do, which is: download external
dependencies and make them available to the rest of the build at
configure time. As a workaround, ExternalProject_Add is used in a tricky
way.

Tests:

* tested with all pinned dependencies
* tested with all system dependencies (except libunicode) on Manjaro
Linux (Arch Linux derivative)
```

TODO:
* get versions of system libraries via regex
* print more info, e.g. versions of libraries found
* print some info about patch binary
* print some info about downloading
* add to commit later: use http instead of https because cmake is built without openssl by default, and it needs it to be able to download via https
* add to commit later: use UPDATE_DISCONNECTED 1 because it does not go on the net to check for a new version to fetch, it uses internet just once. but this is also a caveat if developer has unpinned deps and expects cmake to fetch new version
* probably make a var to let configure behavior of the above
* where did windows take patch binary? maybe install windows to test it, but I'd much rather have someone else investigate this


avemilia:
Expected to fail on Windows due to `patch` binary. Probably shouldn't fail on macOS.

avemilia:
No idea why it passes on Windows, even after looking at the full log. Perhaps need to add more message()s? Will wait for feedback.

christianparpart:
Wir nice. Give me some time to properly read through this. I will reply as soon as possible.

christianparpart:
@avemilia many thanks for the great work. I have also tested this with using local system dependencies, which is great.

One thing that did not work is using fmtlib as system dependency. I want to merge it nevertheless, since the default is using the embedded fmtlib. Maybe we can improve on that after this merge. :)

avemilia:
> One thing that did not work is using fmtlib as system dependency.

If you want me to look at this, please open a new issue and ping me there, with commands you used and cmake output and cmake error log attached. I have tested `-DCONTOUR_EMBEDDED_FMTLIB=0` on Manjaro (as per commit info) and it works with
```
❯ pacman -Qi fmt
Name            : fmt
Version         : 7.1.3-1
Description     : Open-source formatting library for C++
Architecture    : x86_64
URL             : https://fmt.dev
Licenses        : BSD
Groups          : None
Provides        : libfmt.so=7-64
```

avemilia:
Although I'm not sure if we want to dig into this. I did this with only one thought in mind, to allow distro maintainers to package this software and decouple it from external dependencies. When they will have problems with specific versions, I think they will open an issue themselves or just patch it.

PR #192, opened 2021-03-07
CI: Adds an Ubuntu build test using system dependencies (catch2, yaml-cpp)

christianparpart:
Missing: fmtlib (needs a fix).

maybe also test on Arch Linux?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,6 +54,76 @@ jobs:
     - name: "test: libterminal"
       run: ./build/src/terminal/terminal_test
 
+  ubuntu_2004_sysdep:
+    name: "Ubuntu Linux 20.04 (sysdep)"
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v1
+    - name: "initialize sub modules"
+      run: git submodule update --init --recursive
+    - name: set environment variables
+      id: set_vars
+      run: ./scripts/ci-set-vars.sh
+      env:
+        REPOSITORY: ${{ github.event.repository.name }}
+    - name: "install dependencies"
+      run: |
+        set -ex
+        cat /etc/apt/sources.list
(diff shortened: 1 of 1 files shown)


ResuBaka:
For arch linux you just need to add in the dockerfile "catch2 fmt yaml-cpp", when you want to test it with arch too.

christianparpart:
> For arch linux you just need to add in the dockerfile "catch2 fmt yaml-cpp", when you want to test it with arch too.

Thinking about it, maybe embedded should remain default. I also had trouble enabling system deps on Ubuntu LTS since only meet Ubuntu does have catch2 (but Github CI doesn't have that as build eng).

christianparpart:
While writing this, i realized I could use docker for testing then. Not merging yet then.

avemilia:
> maybe embedded should remain default
Definitely, in my opinion. I honestly see little point in testing with system deps, because it's very hard to know what is the exact version of the dependency if something doesn't compile. Even if it says version x.y.z, you still need to look up if it's patched, how it's packaged, so on and so on. Better leave this tedious work to distro maintainer, who knows better and will open an issue if he so feels. As I said previously, my idea behind it was to just provide an easy way to override defaults, but of course you can repurpose it for whatever you think is reasonable. I just don't think it's in the scope of the project to support bazillion different native distro setups.

christianparpart:
@avemilia I agree. However, As a distribution maintainer you are eager to use system dependencies for packages, so that you do not have to update every package that has yaml-cpp library embedded but that one actually has a CVE that needs to be adressed. Package maintainers love shared libraries for that exact reason. I just wanna make sure it is possible (i.e. not accidentally break that). Other than that, I'd go the embedded route (I remember that nvim is also embedding quite a lot by default).

christianparpart:
hm. not worth it for the time being. closing. can be redone at any time.

PR #196, opened 2021-03-10
Clang Static Analyser + Clang-Tidy in the CI

whisperity:
Add a new CI job that runs the latest CodeChecker with the latest nightly Clang and analyses the project.

If a **push** happens to `master`, the Clang Static Analyser will work in [Cross-Translation-Unit analysis](http://clang.llvm.org/docs/analyzer/user-docs/CrossTranslationUnit.html) mode. This increases the ability to find more bugs and more versatile reports when functions are called that are defined outside of the current TU, but it increases both the analysis time and the ability for Clang to *crash* immensely.
In every other case, e.g. for **pull requests** and to commit to non-`master` branches, the normal (analysis restricted to current translation unit only) analysis will be done.

In both modes, Clang-Tidy is executed too.

The tally of the findings is printed by `CodeChecker parse` at the end, and the results of the analysis, converted to self-contained HTMLs detailing the bugs, is ZIPped and uploaded as a build artefact.

This job can't fail unless the analysis hangs or crashes or times out. No "difference" mechanics are implemented. The only outcome of the analysis is the bug reports in the produced artefact.

### DIFF
--- a/.github/workflows/codechecker-analysis.yml
+++ b/.github/workflows/codechecker-analysis.yml
@@ -1,0 +1,150 @@
+name: CodeChecker C++ Static Analysis
+
+on:
+  push:
+    paths-ignore:
+      - 'docs/**'
+      - '.github/ISSUE_TEMPLATE/**'
+      - '.github/*.yml'
+      - '*.md'
+    branches:
+      - master
+      - codechecker
+      - feature/**
+      - fix/**
+      - improvement/**
+      - wip
+    pull_requets:
+      branches:
+        - master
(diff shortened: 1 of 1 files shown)


whisperity:
Because this is a new CI script it doesn't automatically run in your repo as far as I can see.

Anyways, here's the example output of a [CTU analysis](http://github.com/whisperity/contour/actions/runs/639644921) and a [non-CTU one](http://github.com/whisperity/contour/actions/runs/639731199).

PR #199, opened 2021-03-11
[CodeChecker] Ignore external third party libraries and generated code from analysis

whisperity:
The following improvements were made:

 * Put the common configuration of the analysis into a dedicated config file to deduplicate parts of the CI.
 * Ignore external code sources, like `Build/3rdparty` and `/usr/include` from the analysis.
    * TUs (CPP files) that are in these paths will not be invoked the analysis for, at all.
    * AFAIK (but I might be wrong in this), analysis findings that are reported to files in these directories (e.g. headers in `/usr/include`) will be culled from the analysis. This might be detrimental in case a bogus usage of the STL starts from a core Contour TU but the "null dereference" is in the STL or something like that. Will have to see where this leads us...
 * Added a new action in the CI that shows the result summary to the CI standard output itself

### DIFF
--- a/.github/codechecker_config/config.json
+++ b/.github/codechecker_config/config.json
@@ -1,0 +1,13 @@
+{
+  "analyze": [
+    "--enable=portability",
+    "--enable=security",
+    "--enable=sensitive",
+    "--disable=google",
+    "--enable=google-build-namespaces",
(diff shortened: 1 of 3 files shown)


christianparpart:
Many thanks. :)

PR #201, opened 2021-03-13
Fixes Render glitch

### LINKED ISSUE #194: Weird "rolling" font rendering issue
## Description

There is some odd "rolling shutter" like glitch in rendering letters. Letters that fall into the same column inside the viewport will all be distorted in the same way (around the same "distance" when measured against the width of the letter's block).

Pictures speak better than words here, see below!

## Environment

- Contour Version or Git commit hash: Prelelease 0.2.0, Build 264
- Operating System (name / distribution / version): Ubuntu 18.04.5 LTS
- Contour configuration: JetBrains Mono font.
- TERM environment variable: xterm-256color
- Compiler version: GitHub

## Steps to Reproduce

Type in letters to the terminal, at least 10-20 in a row.

## Details

Look at the `/`s in the picture, noting the `/` that is part of the "current directory". Both `/` above and below it is broken in the same way. Similar to all the other broken glyphs. It seems to happen once roughly every 10 columns. Note also the `l` in **`l`**`oaded`.
image

However, it is not restricted to the `/` letter. It seems the only determining factor on both the existence and the kind of the glitch is which column the glyphs was rendered to.
image

If you check the first `s` below the `/` so in the same column, that `/` is bogus, and so is the `s` in the same column. The `s` under a normal `/` (below the `G` in `SIGINT`) looks normal.
However, the `N` in `SIGINT` also looks broken in the middle, just like how in the same column both `/` and the `l` of `self` is **also** affected similarly.
image

Compare the bottom left corner of the `/` and the left side of the `s` with the other `s`. There is a few pixel wide glitch there, on both.
image

Now the `/` on the top right has a bit of a glitch, and so is the `l` which bottom right part is longer than the other `l`'s (which `/` in the same column isn't glitching)
image

If you check the rightmost `f` here, it has a bit of a glitch to the left of the "trunk" of the `f`. And just in the same position above, the `/` has a bit of a glitch... at around 30% of the total "width" of the glyph.
The previous `f` on the left which has a clear `/` above it doesn't suffer from the glitch.
image

The `9`s in the date are deterministically broken in that column. However, look at the number **just before** the `/` like `82/85` and `83/85`. The `2` and `3` is broken, because they are in the same column.

What's more, in the bottom 5 lines, where the same column is now showing a `/` instead of a number (because the left number is 1 digit long only), it is the `/` that is broken. Look closer at the `e` in `Pr`**`e`**`-analysis` and and the `a` in `St`**`a`**`rting`, which fall onto the same column, they are bonkers too.
Just as how the column of `c` in **`c`**`langsa`, all the `c`s are broken. The `y` in `Pre-anal`**`y`**`sis` in the same column, and all the 1st letters of the filenames printed above, suffer from the glitch, at basically the same position inside the glyph.
image

Here's the previous image zoomed in and some of the affected columns highlighted.
image

christianparpart:
Also improves on logging quite a bit.

Sadly the scrollbar isn't down until the windows gets resized or the screen buffer out into alt mode and then back (WTF Qt....).

Fixes #194

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -1,5 +1,11 @@
 # Milestone 0.2.0 checklist
 
+- [ ] move scrollbar into profile
+- [ ] `ls -l --color=yes /` with wrapping on a bg-colored file (vmlinuz...) will cause the rest of the line to be bg-colored, too. that's wrong. SGR should be empty.This problem only exists when not having resized yet.
+- [ ] altlas debug output on laptop seems blank?
+- [ ] vim's wrap mode with multiline text seems to have rendering issues.
+- [ ] CI: build with parallel support (STL then requires tbb apparently)
+- [ ] `times()` helper type needs random access iterator category so that it can be used for parallel algorithms (try with `std::copy(std::execution::par, ...)`)
 - [ ] FIXME: `reset` resets screen size to 80x25, should remain actual one.
 - [ ] debuglog: filter by logging tags (in a somewhat performant way), so the debuglog (when enabled) is not flooding.
 - [ ] CopyLastMarkRange seems not to work (at least for double-line prompts in zsh/p10k)
--- a/src/contour/CMakeLists.txt
+++ b/src/contour/CMakeLists.txt
@@ -29,6 +29,7 @@ set(contour_SRCS
     FileChangeWatcher.cpp FileChangeWatcher.h
     TerminalWindow.cpp TerminalWindow.h
     TerminalWidget.cpp TerminalWidget.h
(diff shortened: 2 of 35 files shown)


christianparpart:
Merging this PR before it's getting out of hand. :)

whisperity:
Yep, this definitely fixed it, it seems.

christianparpart:
There is one problem if you set the scrollbar to hidden. This is quickly fixed. Will do soon.

PR #209, opened 2021-03-23
[CI] CodeChecker: Speed up analysis

whisperity:
Because the CI does not run a web server, there is no need to download and use NodeJS and the "web" part of CodeChecker.

In addition, this patch makes the presentation of the analysis results better by adding LLVM version to the artefact name.


christianparpart:
Many thanks for the improvement. :-)

PR #211, opened 2021-04-01
ci: use archlinux/archlinux image which is the official archlinux image

ResuBaka:
Instead of an image that is build from an 3party that builds an arch linux based image, we should use the offical image, in my opinion.


uspasojevic96:
Steps need to be updated in the yml file due to not using buildx anymore

ResuBaka:
We still use buildx and why should there be an update to the action task as it still uses this dockerfile as the entry. Can you explain it @uspasojevic96 ?

uspasojevic96:
My bad, I was thinking that GHA supports Docker natively

christianparpart:
FYI: I have been updating the base branch just now. Many thanks for your contribution @ResuBaka - I will see if that becomes green soon.

christianparpart:
It's green! Happy Eastr! :egg: :hatching_chick:

PR #216, opened 2021-04-07
[CI] CodeChecker: Fix build breaking due to exit status signalling "reports found"

whisperity:
Originally, it was the `analyze` command which signalled through the exit status that _"some reports were found"_. Now, Ericsson/CodeChecker#3232 *moved* this weirdness to the `parse` command.

CI script pulled after that, now we're also ignoring the exit status of the `parse` command. Hopefully this will make it so that the analysis artefacts are collected properly...


whisperity:
Right. I had a fear that the `parse` command which does the HTML export will produce exit status too, but it doesn't seem like it, so I guess this is ready to go. See: http://github.com/whisperity/contour/runs/2290633905 for the run.

PR #219, opened 2021-04-23
[contour] Changes CLI parameters `output` to `to` in all cases except for capture command.

christianparpart:
This is how the CLI syntax would look like with this change:

image

/cc @whisperity


whisperity:
Can we call it `to FILE` instead of `to STRING`?

whisperity:
I think the `capture` command can be aligned with this too.

PR #226, opened 2021-06-04
[CodeChecker] Fix CI fail by `parse` command broken again

whisperity:
Turns out that now the `parse` command shows non-zero exit status in case reports are emitted to HTML...

PR #227, opened 2021-06-05
[CI] Fix CodeChecker again - `analyze` needs the exit status ignored

whisperity:
> Follow-up for #226

Sorry, I actually broke one half in the previous patch.

### DIFF
--- a/.github/workflows/codechecker-analysis.yml
+++ b/.github/workflows/codechecker-analysis.yml
@@ -25,11 +25,11 @@ jobs:
     env:
       # The version of LLVM to use. Needed to set the appropriate clang-XX
       # binaries.
-      LLVM_VER: 13
+      LLVM_VER: 12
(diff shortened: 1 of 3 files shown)


.github/workflows/codechecker-analysis.yml
```diff
@@ -14,9 +14,9 @@ on:
       - fix/**
       - improvement/**
       - wip
-    pull_requests:
-      branches:
-        - master
+  pull_request:
```
whisperity, line 17: This was broken in 820f8973. The syntax is `pull_request` and not `pull_request`**`s`**. Also, the indentation was bad. :grin: 

Fun fact: there's a coc plugin for checking JSON schemae or something, because my editor was showing an error:
Property pull_requests is not allowed

christianparpart:
We have a CI label now :)

PR #248, opened 2021-06-24
Workaround for MSVC compilation.

christianparpart:
and clang on OS/X.


WSLUser:
You honestly ought to ditch MSVC altogether, it has so many limitations compared to gcc and clang. The folks in charge of MSVC are improving things, but it's slow going and they still can't achieve the same perf as the other 2 compilers.

PR #260, opened 2021-06-28
Fix invalid detection of source path in autogen script

jcelerier:
In zsh, realpath $0 returns ".", not an absolute path.
Thus it is necessary to obtain the absolute path since otherwise cmake "${ROOTDIR}" will look for a CMakeLists.txt in the build directory.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -1,7 +1,12 @@
 #! /bin/bash
 set -ex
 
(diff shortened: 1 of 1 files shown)


christianparpart:
Oh that is an interesting finding. Many thanks @jcelerier :)

PR #262, opened 2021-06-28
Make offline builds possible

GoldsteinE:
Change `FetchContent_Populate()` to a pair of `FetchContent_Declare()` + `FetchContent_Populate()` to allow offline builds via `FETCHCONTENT_FULLY_DISCONNECTED` variable.


christianparpart:
Many thanks @GoldsteinE
I don't use FetchContent for long, so I am not so confident in using it yet (but it's definitely better than what I had before). Many thanks for your contribution. :)

PR #265, opened 2021-06-29
Update yaml-cpp to latest

data-man:


christianparpart:
💟 many thanks :)

data-man:
What if to use https://github.com/cpm-cmake/CPM.cmake (requires CMake >= 3.14)?

christianparpart:
> 

> What if to use https://github.com/cpm-cmake/CPM.cmake (requires CMake >= 3.14)?

I think fetchcontent is fine for now, isn't it? P.s. didn't check cpm in detail yet.

data-man:
Yes, fetchcontent is fine, but cpm is finest. :)

PR #266, opened 2021-06-29
Update libunicode to latest

data-man:


christianparpart:
That's interesting. the windows build fails because there is no `catch2/catch_all.hpp` (which is only available in the latest Catch2 beta), but it does build on the other platforms? Hmm...

christianparpart:
> Maybe use devel Catch2 in Contour too?

I think it makes sense to ease the use of Catch2 in the sub repos and the parent repo. yeah.

data-man:
Merged to #270. :(

christianparpart:
> Merged to #270. :(

Oops. Did i do a mistake here? If so, i must apologize for that

data-man:
> Did i do a mistake here?

Why? No!

PR #279, opened 2021-06-30
Update decoration.sh

data-man:
Just as a reminder of what features aren't done yet. :)


christianparpart:
Oh. Thanks! That'll be very helpful. :)

PR #284, opened 2021-07-01
Adds config option `mode` to input modifiers for additionally filtering.

### LINKED ISSUE #269: Allow Input Mapping to conditionally apply based on screen buffer (primary | alt)
I just had the case where I wanted to bind `{` and `}` to rel-jump marks up and down, but then I cannot use these in Vim (alt screen) anymore. So let's add a `screen: alt` / `screen: primary` option (defaulting to `alt|primary`) - or similar syntax for input mappings.

This minor feature should probably also still make it into 0.2.0 as it seems super useful (and trivial to add).

christianparpart:
Closes #269.

Example input mappings:

```yaml
- { mods: [Shift], key: '{', action: ScrollMarkUp, mode: "~Alt"}
- { mods: [Shift], key: '}', action: ScrollMarkDown, mode: "~Alt"}
```

this mimmicks Shift+{ and Shift+} from VIM to paragraph-wise jump up/down (between command prompts in this case :D)

### Checklist

- [ ] Add unit tests for testing each mode matching (at least for key press)
- [ ] does mouse input matching based on modes make any sense?


src/contour/Config.cpp
```diff
@@ -456,6 +456,54 @@ optional<terminal::Modifier::Key> parseModifierKey(string const& _key)
...
+                                        YAML::Node const& _node)
+{
+    using terminal::MatchMode;
+    if (!_node)
+        return MatchMode::Default;
+    _usedKeys.emplace(_prefix);
+    if (!_node.IsScalar())
+        return nullopt;
+
+    MatchMode mode;
```
christianparpart, line 470: ` = MatchMode:: Default`

PR #288, opened 2021-07-03
Update fmt library to 8.0.1

data-man:


christianparpart:
Nice. Many thanks.

data-man:
Hmm, some tests are failed. :(

christianparpart:
rebase on top of latest master, I think I've just comitted a fix for that. I really wonder why the HECK I didn't realize that earlier.

PR #301, opened 2021-07-06
Bump GH actions

data-man:


.github/labeler.yml
```diff
@@ -1,22 +1,24 @@
-"frontend":
-    - src/contour/*.cpp
-    - src/contour/*.h
+"CI":
+    - .github/**
+"CMake":
+    - cmake/**
```
christianparpart, line 4: Here too, have `CmakeLists.txt` added. Does this make sense?

data-man: Let's try `**CMakeLists.txt`.

christianparpart (approved):

PR #305, opened 2021-07-07
Building without `LIBTERMINAL_LOG_TRACE` option

data-man:


christianparpart (approved):
Oops. You are right! Many thanks. :)

PR #314, opened 2021-07-10
Add EXCLUDE_FROM_ALL to all third-parties

data-man:
And update yaml-cpp to latest.


christianparpart:
Let's give it a try.

cmake/ThirdParties.cmake
```diff
@@ -52,6 +56,6 @@ CPMAddPackage(
   NAME yaml_cpp
   VERSION ${3rdparty_yaml_cpp_version}
   URL https://github.com/jbeder/yaml-cpp/archive/${3rdparty_yaml_cpp_version}.zip
-  URL_HASH SHA256=73145216ee7a0fd2dd45bf8c62b3d9b805e99f538329620df41457bd9da6e506
+  URL_HASH SHA256=a047cb1266bd994b60980fdc48cf0bb63880dc62a6997849014aebdaaeaf4495
```
christianparpart, line 59: NB: I think the old code did make the SHA hash cmake configurable, too. Just having the URL configurable does not make sense if one cannot also change the expected hash. (not too urgent to fix though, I just wanted to raise that awareness).

Issue #315, opened 2021-07-10
crash caused by CPM?


christianparpart:
As noted in #313, at least on my machine (Ubuntu 21.04) I can reliably reproduce the SEGV in `YAML::LoadFile()` call just by the order of the `CPMAddPackage()` calls in `ThirdParties.cmake`.

What is definitely suspicios is the following:

![image](https://user-images.githubusercontent.com/56763/125177138-b245be00-e1d9-11eb-89a0-d5ee9f3f4d84.png)

Have a look at the line `-- CPM: libunicode: adding package yaml_cpp ...`.
The other CPM lines look different, i.e. they don't mention two different package names.

It might very well be a bug from CPM.cmake? At least the fact that the order changes the behavior does not sound quite right.  /cc @data-man (this is a tracker ticket).

data-man:
Sorry, my system is now broken after the nightly update (I'm writing from the live OS). :(
I will look into the problem as soon as I fix my system.

data-man:
@christianparpart can you check https://github.com/jbeder/yaml-cpp/pull/1011 in your system?

christianparpart:
Hey. I will. Currently our ISP decided to take nature as excuse for taking me offline though. :)

christianparpart:
We do not use CPM anymore

PR #316, opened 2021-07-10
[CMake] Changing the order of CPMAddPackage() calls fixes YAML's SEGV

christianparpart:
Refs #315


data-man (approved):

christianparpart:
@data-man I am merging without thinking about why this fixed it, so that I can continue working on master. :)

PR #321, opened 2021-07-17
Add link to `ANSI escape codes` in Wikipedia

data-man:


christianparpart:
in the long run, we need some kind of structured document that documents all the VT sequences that Contour supports.

Including the extensions and where these extensions are coming from (some from us, some from others). I don't have a perfect idea on how to structure such a document so that it could be also displayed as web page on github pages, but let's see.

data-man:
What if to add a new command?
E.g. `contour list-features`.

christianparpart:
> What if to add a new command?
> E.g. `contour list-features`.

I think i would by far prefer to generate the list of supported VT sequences automatically by contour. It does absolutely make sense to add a CLI command for that to`contour`. So there documentation is also always reflecting exactly what the binary supports.

The `Functions.h` would need to be touched to better support full text documentation per VT sequence. The doc could use markdown and then converting that for the web would also be easy then.

PR #329, opened 2021-07-19
Fixing Qt6 compatibility

data-man:
```c++
#if QT_DEPRECATED_SINCE(5, 15)
    // Actually deprecated since 5.0, in docs
    QT_DEPRECATED_VERSION_X_5_15("Use angleDelta()")
    inline int delta() const  { return qt4D; }
    // Actually deprecated since 5.0, in docs
    QT_DEPRECATED_VERSION_X_5_15("Use angleDelta()")
    inline Qt::Orientation orientation() const { return qt4O; }
```

### DIFF
--- a/src/contour/helper.cpp
+++ b/src/contour/helper.cpp
@@ -159,16 +159,16 @@ void sendWheelEvent(QWheelEvent* _event, TerminalSession& _session)
 {
     auto const yDelta = [&]() -> int
     {
-        switch (_event->orientation())
-        {
-            case Qt::Orientation::Horizontal:
-                return _event->pixelDelta().x() ? _event->pixelDelta().x()
-                                                : _event->angleDelta().x();
-            case Qt::Orientation::Vertical:
-                return _event->pixelDelta().y() ? _event->pixelDelta().y()
-                                                : _event->angleDelta().y();
-        }
-        return 0;
+//        switch (_event->orientation())
+//        {
+//            case Qt::Orientation::Horizontal:
+//                return _event->pixelDelta().x() ? _event->pixelDelta().x()
+//                                                : _event->angleDelta().x();
+//            case Qt::Orientation::Vertical:
+//                return _event->pixelDelta().y() ? _event->pixelDelta().y()
+//                                                : _event->angleDelta().y();
+//        }
+        return _event->angleDelta().y();
     }();
 
     if (yDelta)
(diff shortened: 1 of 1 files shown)


christianparpart:
Oh dear, the removed it.

Apparently we need a CI run for Qt 6 builds.

However, the reason why this test on orientation does exist is the behavior of using your scroll wheel while having the Alt modifier pressed.

If you use your scrollwheel this is usually mapped to `angleDetla.y` EXCEPT for the case when the user also presses Alt modifier. For some **** reason Qt decided (I think it was in Qt version 4.8) to then change vertical scrolling to vertical scrolling. That is why this code path exists, because contour still needs to be able to reliably detect horizontal and vertical scroll wheel moves as-is, regardless of the modifiers being set.

This is needed for the input mapping system, where the user can map `Alt + WheelUp` or `Alt + WheelDown`. But the wheel alt+up/down will never be matched by Qt's wheelEvent() because now the values that were supposed to be in `y` are now in `x`.

I should stop ranting now against Qt. Sorry. :) But maybe there is another way to now reliably know if wheel up/down or wheel left/right was used, regardless of input modifiers?

I think we should wait with a PR merge until this problem is solved (well, we definitely need a Qt 6 CI run).

christianparpart:
p.s.: I think they introduced this in 4.8 because of mobile devices, where one may use the device portrait and landscape modes and changing that frequently.

data-man:
> But maybe there is another way to now reliably know if wheel up/down or wheel left/right was used, regardless of input modifiers?

I'll research. :)

data-man:
> However, the reason why this test on orientation does exist is the behavior of using your scroll wheel while having the Alt modifier pressed.

I don't understand what the problem is. :(
I changed the desktop orientation to 90° clockwise.
Nothing happens when I press `Alt + WheelUp` (with Qt5 and master branch).

christianparpart:
@data-man 

I am merging this now, as it's at least fixing Qt6 build. and I think i have an idea how to fix the fix. :)

As you have been asking what the problem is, I try to explain again:

```yaml
- { mods: [Alt],  mouse: WheelDown,   action: DecreaseOpacity }
- { mods: [Alt],  mouse: WheelUp,     action: IncreaseOpacity }
```

This this without this PR and once with this PR.

The way contour usually worked is, that Alt+WheelUp directly maps to Alt+WheelUp. But since Qt remaps Wheel Up/down events as soon as Alt modifier is pressed, it's actually shown to the caller as wheel left/right instead of up/down.

However, the application *user* expects Alt + WheelUp, when keybinding, to be Alt + WheelUp. And that is not the case with Qt's default behavior.

PR #337, opened 2021-07-22
Update deps

data-man:


christianparpart:
image

I'm getting this in another PR, too. But why. hmm...

data-man:
Maybe we are not using the latest versions.
I'll check it later.

christianparpart:
oopsi. merge conflict :)

christianparpart:
The windows-CI thing is something different and still unknown.
But the OS/X breakage I just pushed a fix onto master. You should rebase for that.

p.s.: I work with rebase instead of repeately merging master into working branches. This keeps the history more linear / clean.

PR #339, opened 2021-07-23
Fixes building on windows, using MSVC.

christianparpart:


cmake/ThirdParties.cmake
```diff
@@ -5,7 +5,7 @@ set(3rdparty_fmt_version "561834650aa77ba37b15f7e5c9d5726be5127df9" CACHE STRING
 set(3rdparty_libunicode_version "a0f72919e4520ee1a02890ea77f19ff16c92d4f8" CACHE STRING "libunicode: commit hash")
```
data-man, line 5: set(3rdparty_libunicode_version "2942d23b441a4df9405227914bed61d0a6d8f26c" CACHE STRING "libunicode: commit hash")

data-man, line 5:
```suggestion
set(3rdparty_libunicode_version "2942d23b441a4df9405227914bed61d0a6d8f26c" CACHE STRING "libunicode: commit hash")
```

data-man: And `URL_HASH SHA256=1cb667bbf3f0e0772364e9ee14fb582d8fc5cbb795dc58ffb1577769409aa590` for `libunicode`.

christianparpart: I can do that tonight.

christianparpart:
it's compiling now again, but not yet running successfully. needs some debugging sessions I guess (not sure i'll include that in this PR here)

PR #347, opened 2021-07-27
[CI][CodeChecker] Add the "swapped/swappable" parameters checkers into the enabled list explicitly

whisperity:
(Also enable using the latest nightly PPA LLVM for the analysis instead of the release version,)


christianparpart (approved):
thanks @whisperity :)

PR #348, opened 2021-07-27
Fix/cpm ppa build

christianparpart:


data-man:
I doubt that changing the CPM version has any effect.
Because the remaining CPM code is unchanged.

christianparpart:
> I doubt that changing the CPM version has any effect.
> Because the remaining CPM code is unchanged.

I fetched the last stable XPM because CPM devs told me it is not wise to have a dev version locally. Turns out it was just a string change ;-)

PR #350, opened 2021-07-29
Improve load_font in DWrite shaper

skyline75489:
* Do not search all the fonts returned by `fontCollection->GetFontFamilyCount`. Use `fontCollection->FindFamilyName` instead.
* Fallback to "Consolas" if no exact font can be found.
* Fallback to "en-US" locale if no exact locale can be found.

### DIFF
--- a/src/text_shaper/directwrite_shaper.cpp
+++ b/src/text_shaper/directwrite_shaper.cpp
@@ -140,70 +140,84 @@ optional<font_key> directwrite_shaper::load_font(font_description const& _descri
     std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> wStringConverter;
     std::wstring familyName = wStringConverter.from_bytes(_description.familyName);
 
-    for (UINT32 i = 0, e = fontCollection->GetFontFamilyCount(); i < e; ++i)
+    UINT32 familyIndex;
+    BOOL familyExists = FALSE;
+    fontCollection->FindFamilyName(familyName.data(), &familyIndex, &familyExists);
+
+    if (!familyExists) {
+        // Fallback to Consolas
+        const wchar_t* consolas = L"Consolas";
+        fontCollection->FindFamilyName(consolas, &familyIndex, &familyExists);
+    }
+
+    Microsoft::WRL::ComPtr<IDWriteFontFamily> fontFamily;
+    fontCollection->GetFontFamily(familyIndex, &fontFamily);
+
+    Microsoft::WRL::ComPtr<IDWriteLocalizedStrings> familyNames;
+    fontFamily->GetFamilyNames(&familyNames);
+
+    BOOL localeExists = FALSE;
+    unsigned index{};
+    familyNames->FindLocaleName(d->userLocale.c_str(), &index, &localeExists);
+
+    if (!localeExists) {
+        // Fallback to en-US
+        const wchar_t* localeName = L"en-US";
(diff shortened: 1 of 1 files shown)


skyline75489:
Tip: use "split" mode for better reviewing experience :)

src/text_shaper/directwrite_shaper.cpp
```diff
@@ -140,70 +140,84 @@ optional<font_key> directwrite_shaper::load_font(font_description const& _descri
...
+    }
+
+    UINT32 length = 0;
 
-            font_weight weight = dwFontWeight(font->GetWeight());
-            if (weight != _description.weight)
-                continue;
+    // TODO: very long font name?
+    wchar_t name[64];
+    familyNames->GetString(index, name, _countof(name));
```
skyline75489, line 178: The `name` is currently unused now. But it *should* be saved into the `FontDescription` somewhere, because this name here might be different from the original name we ask DWrite to find. It's an edge case relatively, so I think it's OK to leave it here. Plus I don't know how to modify `FontDescription` because it seems read-only.

src/text_shaper/directwrite_shaper.cpp
```diff
@@ -140,70 +140,84 @@ optional<font_key> directwrite_shaper::load_font(font_description const& _descri
...
-            names->GetString(i, name, sizeof(name));
+    BOOL localeExists = FALSE;
+    unsigned index{};
+    familyNames->FindLocaleName(d->userLocale.c_str(), &index, &localeExists);
 
-        if (familyName != name)
-            continue;
+    if (!localeExists) {
+        // Fallback to en-US
+        const wchar_t* localeName = L"en-US";
```
skyline75489, line 165: This is quite important because in CJK environment the locale will be something like `zh-Hans-HK` and most programming fonts do not have that locale available.

src/text_shaper/directwrite_shaper.cpp
```diff
@@ -140,70 +140,84 @@ optional<font_key> directwrite_shaper::load_font(font_description const& _descri
...
 
-        IDWriteLocalizedStrings* names;
-        family->GetFamilyNames(&names);
+    Microsoft::WRL::ComPtr<IDWriteFontFamily> fontFamily;
+    fontCollection->GetFontFamily(familyIndex, &fontFamily);
 
-        BOOL exists = FALSE;
-        unsigned index{};
-        names->FindLocaleName(d->userLocale.c_str(), &index, &exists);
+    Microsoft::WRL::ComPtr<IDWriteLocalizedStrings> familyNames;
```
christianparpart, line 156:
```suggestion
    ComPtr<IDWriteLocalizedStrings> familyNames;
```
should work, too. because it got imported on the top of the file.

src/text_shaper/directwrite_shaper.cpp
christianparpart: Interesting. what's the difference of `_countof` to `sizeof`?

christianparpart (approved):
i'm ignoring any coding style differences. I am happy to get some qualified DWrite support much more. It's all minor, and I can't await having a DirectWrite based text shaper & rasterizer on windows. :-)

src/text_shaper/directwrite_shaper.cpp
christianparpart: p.s.: can be changed at any time later. so I can merge early :)

PR #352, opened 2021-07-29
miserable attempt to figure out high CPU load on Ubuntu 18

christianparpart:
Refs #331 

@whisperity can you try out the binary once built by the CI and check the stdout output of the actual executable when reproducing this?

I cannot repro it (will try an Ubuntu 18 VM though).


christianparpart:
Closing until I have a clue again. :)

PR #353, opened 2021-07-29
Always sanitize font description

skyline75489:
This fixes a crash when `_fonts.size.pt` is somehow zero.

### DIFF
--- a/src/contour/helper.cpp
+++ b/src/contour/helper.cpp
@@ -340,6 +340,15 @@ terminal::renderer::PageMargin computeMargin(terminal::ImageSize _cellSize,
     return {leftMargin, bottomMargin};
 }
 
+terminal::renderer::FontDescriptions sanitizeFontDescription(terminal::renderer::FontDescriptions _fonts, crispy::Point screenDPI)
+{
+    if (_fonts.dpi.x <= 0 || _fonts.dpi.y <= 0)
+        _fonts.dpi = screenDPI;
+    if (std::fabs(_fonts.size.pt) <= std::numeric_limits<double>::epsilon())
+        _fonts.size.pt = 12;
+    return _fonts;
+}
+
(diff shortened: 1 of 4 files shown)


christianparpart (approved):
Oh many thanks :)

src/contour/helper.cpp
```diff
@@ -340,6 +340,15 @@ terminal::renderer::PageMargin computeMargin(terminal::ImageSize _cellSize,
     return {leftMargin, bottomMargin};
 }
 
+terminal::renderer::FontDescriptions sanitizeFontDescription(terminal::renderer::FontDescriptions _fonts, crispy::Point screenDPI)
```
uspasojevic96, line 343: `crispy::Point screenDPI` -> `crispy::Point _screenDPI`

src/contour/helper.h
```diff
@@ -173,6 +173,8 @@ terminal::renderer::PageMargin computeMargin(terminal::ImageSize _cellSize,
                                              terminal::PageSize _charCells,
                                              terminal::ImageSize _pixels) noexcept;
 
+terminal::renderer::FontDescriptions sanitizeFontDescription(terminal::renderer::FontDescriptions _fonts, crispy::Point screenDPI);
```
uspasojevic96, line 176: same here

src/contour/helper.cpp
skyline75489: This f**ks me my really hard. 😅  Because in Windows Terminal the naming convention is exactly the opposite. Every member of a class is `_member` and every paramter is `parameter`.

christianparpart: haha. oh sorry. but to be honest, I am already in progress of changing that (in my head). Let's talk about it in Discord :)

christianparpart (approved):

PR #355, opened 2021-07-29
Fix opengl shaper on Win32

skyline75489:


src/terminal/pty/ConPty.cpp
```diff
@@ -84,6 +84,8 @@ ConPty::ConPty(PageSize const& _windowSize) :
 
     if (hr != S_OK)
         throw runtime_error{ GetLastErrorAsString() };
+
+    buffer_.resize(10240);
```
skyline75489, line 88: The buffer_ here had size of 0, which prevents any read operation from happening.

src/terminal_renderer/TextRenderer.cpp
```diff
@@ -550,7 +550,8 @@ void ComplexTextShaper::endSequence()
 text::shape_result const& ComplexTextShaper::cachedGlyphPositions()
 {
     auto const codepoints = u32string_view(codepoints_.data(), codepoints_.size());
-    if (auto const cached = cache_.find(TextCacheKey{codepoints, style_}); cached != cache_.end())
+    auto const cached = cache_.find(TextCacheKey{codepoints, style_});
+    if (cached != cache_.end())
```
skyline75489, line 554: I think this is logic error, unless this is some C++ thing I don't understand.

src/text_shaper/open_shaper.cpp
```diff
@@ -348,6 +348,11 @@ namespace // {{{ helper
         {
             if (_fd.familyName != "monospace")
                 FcPatternAddString(pat.get(), FC_FAMILY, (FcChar8 const*) "monospace");
+#ifdef _WIN32
+            // On Windows FontConfig can't find "monospace". We need to use "Consolas" instead.
+            else
+                FcPatternAddString(pat.get(), FC_FAMILY, (FcChar8 const*)"Consolas");
```
skyline75489, line 354: Without "Consolas" as fallback, the code will crash because of incorrect font metrics.

src/terminal/pty/ConPty.cpp
christianparpart: Ah yes! I just recently refactored that part on Pty in general, adding this `buffer_`, but didn't actually test on windows. Thanks!

src/terminal_renderer/TextRenderer.cpp
christianparpart: Not sure what exactly you meant but that's what it is doing:

check if `TextCacheKey{codepoints, style_}` is in cache, and if so, return that.
otherwise, create cache entry and return that.

The syntax i used is:

```cpp
if (initializer; expression)
    body;
```
just like for-loop initializer, you have that now (C++11) also on if-expression.

christianparpart (approved):
Many thanks for taking care of the windows part, @skyline75489 - I really appreciate that. :)

PR #357, opened 2021-07-30
Refine monospace font search for macOS

skyline75489:


skyline75489:
The reason why I noticed this on both Windows & macOS is that, on these two platforms "monospace" can't be correctly found by font config. Also the default font name inside `config.yml` is precisly "monospace". Failing to find "monospace", FontConfig will choose some other font. This causes trouble here:

https://github.com/contour-terminal/contour/blob/aa14e374c6fc705e783a0b48b42bf14cc10e655a/src/text_shaper/open_shaper.cpp#L460-L469

Some font (how "lucky" I am) doesn't report `horiAdvance` correctly. So `computeAverageAdvance` will return 0(!). This then leads to `gridMetrics_().cellSize` has `width = 0`, which finally leads to "divided by zero" exception somewhere I don't remember.

In conclusion, this PR is more of a naive fix. If the users delibrately choose a font that doesn't exit, the same crash will happen. A better solution is needed.

src/text_shaper/open_shaper.cpp
```diff
@@ -346,12 +346,17 @@ namespace // {{{ helper
...
             // On Windows FontConfig can't find "monospace". We need to use "Consolas" instead.
-            else
+            if (_fd.familyName == "monospace")
                 FcPatternAddString(pat.get(), FC_FAMILY, (FcChar8 const*)"Consolas");
+#elif __APPLE__
+            // Same for macOS, we use "Menlo" for "monospace".
+            if (_fd.familyName == "monospace")
+                FcPatternAddString(pat.get(), FC_FAMILY, (FcChar8 const*)"Menlo");
+#else
+            if (_fd.familyName != "monospace")
```
skyline75489, line 358: I also don’t quite understand this. This is the original logic without all my fixes. Is this line force the user to use “monospace” provided by system? What if the user choose a different font? Will this override that? I didn’t really try. Just curious.

christianparpart: To be fair, fontconfig documentation wasn't really that insightful to me. And while having had a look at quite some other source places I came up with this one. So I think it is "okay" to merge what you think works better especiall on Win32 for now. and we can always reiterate and improve. :)

christianparpart (approved):

PR #358, opened 2021-07-30
Draft implementation of DWrite shaper

skyline75489:


src/text_shaper/directwrite_shaper.cpp
```diff
@@ -279,58 +284,137 @@ void directwrite_shaper::shape(font_key _font,
...
+            gpos.offset.x = static_cast<int>(cellWidth);
+            //gpos.offset.y = static_cast<int>(static_cast<double>(pos[i].y_offset) / 64.0f);
+
+            gpos.advance.x = static_cast<int>(cellWidth);
+            //gpos.advance.y = static_cast<int>(static_cast<double>(pos[i].y_advance) / 64.0f);
+            _result.emplace_back(gpos);
+        }
+    }
+    else {
+        // TODO
```
skyline75489, line 341: This is a BIG TODO here. No complex shaping is supported right now. Consolas will report most ASCII text as "simple", but other fonts might not do so. Hard-code `_isEntireTextSimple` if you need it for testing.

src/text_shaper/directwrite_shaper.cpp
```diff
@@ -279,58 +284,137 @@ void directwrite_shaper::shape(font_key _font,
...
+    }
 }
 
 std::optional<rasterized_glyph> directwrite_shaper::rasterize(glyph_key _glyph, render_mode _mode)
```
skyline75489, line 345: This is largely based on my previous experience and [crossfont](https://github.com/alacritty/crossfont).

One thing I don't understand is that, Alacritty doesn't do shaping at all. All it does in crossfont is `load_font` and then `rasterize`, which is why it doesn't support complex Emoji & ligatures. Here the `rasterize` in contour uses similar interface as in crossfont`(`glyph_key` -> `bitmap`). How does contour handles shaping? Sorry if this is a dumb question. I'm quite an amateur in this area.

src/text_shaper/directwrite_shaper.cpp
christianparpart: When this function is called, the text is already segmented by the font properties needed to do proper text shaping, i.e.

* the text sequence contains only one script (e.g. Latin, or Greek but never both)
* either emoji emoji presentation **or** emoji text presentation
* split up by words (space-delimited)

src/text_shaper/directwrite_shaper.cpp
christianparpart: Yes. Alacritty does not text shaping at all because it is a very expensive task and does cost significant CPU (if done naive). I think one can at least save a bit via caching, which we do. :)

This rasterize function already takes a glyph ID regardless of whether text shaping was involved or a simple 1:1 translation has happened.

### How does contour handle shaping:

It does support two modes, basic shaping (just codepoint to glyph index mapping) and complex text shaping.

primitive codepoint-to-glyph-index translation is done via https://github.com/contour-terminal/contour/blob/master/src/text_shaper/shaper.h#L128-L129

whereas complex textshaping can be implemented via:https://github.com/contour-terminal/contour/blob/master/src/text_shaper/shaper.h#L109-L126

the open_shaper implementation does exactly that. the smaller of those two function where meant to bypass (complex) text shaping in order to gain performance (for the sake of losing features), but turns out in my naive tests it's actually not that much gained (due to caching) :-)

christianparpart: WRT text shaping in terminals I can highly recommend reading https://github.com/contour-terminal/contour/blob/master/docs/text-stack.md - I've put a lot of afford into this document so I won't forget (or if I do, i have a place to quickly recap)

src/text_shaper/directwrite_shaper.cpp
```diff
@@ -279,58 +284,137 @@ void directwrite_shaper::shape(font_key _font,
...
+        renderingMode = DWRITE_RENDERING_MODE_NATURAL_SYMMETRIC;
+    }
+
+    ComPtr<IDWriteGlyphRunAnalysis> glyphAnalysis;
+    rasterized_glyph output{};
+
+    d->factory->CreateGlyphRunAnalysis(
+        &glyphRun,
+        pixelPerDip,
+        NULL,
```
christianparpart, line 386:
```suggestion
        nullptr,
```

src/text_shaper/directwrite_shaper.cpp
skyline75489: OK hard-code `_isEntireTextSimple ` won't work because `_glyphIndices` isn't filled with the correct indices...Just use Consolas for testing 😅

christianparpart (commented):
I am not complaining about different coding style nor coding style inconsistencies as the actual code is much more important, and the minors can be done later at any time / willingness :)

src/text_shaper/directwrite_shaper.cpp
```diff
@@ -279,58 +284,137 @@ void directwrite_shaper::shape(font_key _font,
...
-    //                     featureRangeLengths,
-    //                     featureRanges,
-    //                     maxGlyphCount,
-    //                     clusterMap,
-    //                     textProps,
-    //                     glyphIndices,
-    //                     glyphProps,
-    //                     &actualGlyphCount);
-
+    std::wstring_convert<std::codecvt_utf8<char32_t>, char32_t> conv1;
```
skyline75489, line 287: This is also dumb. I failed to convert `u32string` directly to `wstring`, so I had to do it like this. Hope there is a way to avoid such dumbness.

christianparpart: Maybe the following should help here?

https://github.com/contour-terminal/libunicode/blob/master/src/unicode/convert.h#L332-L339

Mind, I have never done anything with UTF-16. So if the suggested way does not work, that this bug should be fixed (Out of my head, I do not know if I have unit-tested UTF-16, but if not, I should add some).

[libunicode](https://github.com/contour-terminal/libunicode/) is a Unicode library I specifically created to aid Contour's Unicode needs (especially the containing algorithms).

skyline75489: If libunicode doesn’t handle “surrogate pair” correctly in UTF16, then the conversion won’t work. The better UTF-16 support is obviously not the focus of this PR so I think I will probably just leave it here, if you are OK with it. 

I do wonder why utf32 is used in this method. Does utf32 make things easier in harfbuzz or something? Just curious.

src/text_shaper/directwrite_shaper.cpp
skyline75489: Wonderful writeup about text shaping & rendering!

PR #359, opened 2021-07-30
random stuff i encountered while being on Windows

christianparpart:


src/contour/CMakeLists.txt
```diff
@@ -88,7 +88,7 @@ if(WIN32)
     if (NOT ("${CMAKE_BUILD_TYPE}" STREQUAL "Debug"))
         set_target_properties(contour PROPERTIES
             # Tell MSVC to create a GUI application.
-            WIN32_EXECUTABLE ON
+            #WIN32_EXECUTABLE ON
```
christianparpart, line 91: this should not have made it in. That however is the reason why I did not see console output in the first place. Even though I was in `CMAKE_BUILD_TYPE=Debug` mode? hmm

src/contour/ContourApp.cpp
```diff
@@ -231,6 +235,18 @@ ContourApp::ContourApp() :
...
 
+#if defined(_WIN32)
+    // Enable VT output processing on Conhost.
+    HANDLE stdoutHandle = GetStdHandle(STD_OUTPUT_HANDLE);
+    DWORD savedModes{}; // NOTE: Is it required to restore that upon process exit?
+    if (GetConsoleMode(stdoutHandle, &savedModes) != FALSE)
+    {
+        DWORD modes = savedModes;
+        modes |= ENABLE_VIRTUAL_TERMINAL_PROCESSING;
+        SetConsoleMode(stdoutHandle, modes);
```
christianparpart, line 246: I assumed one has to explicitly enable VT output processing, but for some reason it also worked without on my host. I think it should still stay in here just to be sure it is enabled? 

NOTE: I am not sure i should restore previous modes upon exit, or whether conhost does not care.

skyline75489: To make it work reliably, this is needed. However this is some situation that it works without manually setting `ENABLE_VIRTUAL_TERMINAL_PROCESSING`, which maybe considered bug or design? There's some issue in WT repo discussing this but I can't find them now.

You don't need to restore previous mode. Conhost will survive.

PR #360, opened 2021-07-31
Emoji rendering in DWrite shaper & DWrite font locator

skyline75489:


src/text_shaper/directwrite_shaper.cpp
```diff
@@ -66,6 +66,60 @@ namespace
...
+        );
+
+        for (auto i = 0; i < height; i++)
+            for (auto j = 0; j < width; j++)
+            {
+                const auto base = ((height - 1 - i) *width + j) * 3;
+                const auto r = tmp[base];
+                const auto g = tmp[base + 1];
+                const auto b = tmp[base + 2];
+                if (r > 0)
```
skyline75489, line 92: This is my naive implementation to composite multiple layers of bitmap.

src/text_shaper/directwrite_shaper.cpp
```diff
@@ -66,6 +66,60 @@ namespace
...
+                if (b > 0)
+                {
+                    *_it++ = b;
+                }
+                else
+                {
+                    _it++;
+                }
+
+                if (_targetFormat == bitmap_format::rgba)
```
skyline75489, line 117: Seems that I have to create a `rgba` format bitmap for colorAtlas.

src/terminal_renderer/TextRenderer.cpp
```diff
@@ -251,12 +251,12 @@ optional<TextRenderer::DataRef> TextRenderer::getTextureInfo(text::glyph_key con
             if (optional<DataRef> const dataRef = ta->get(_id); dataRef.has_value())
                 return dataRef;
 
-    bool const colored = textShaper_.has_color(_id.font);
-
     auto theGlyphOpt = textShaper_.rasterize(_id, fontDescriptions_.renderMode);
     if (!theGlyphOpt.has_value())
         return nullopt;
 
+    bool const colored = textShaper_.has_color(_id.font);
```
skyline75489, line 258: This is a bit awkward, because DWrite engine needs to first `rasterize` then have the `has_color` available.

src/text_shaper/directwrite_shaper.cpp
```diff
@@ -671,42 +727,76 @@ std::optional<rasterized_glyph> directwrite_shaper::rasterize(glyph_key _glyph,
...
+            output.bitmap.resize(*height * *width * 4);
+            output.format = bitmap_format::rgba;
 
-    if (FAILED(hr)) {
-        return nullopt;
+            d->fontsHasColor.at(_glyph.font) = true;
+            while (true)
+            {
+                BOOL haveRun;
+                glyphRunEnumerator->MoveNext(&haveRun);
```
skyline75489, line 765: See https://docs.microsoft.com/en-us/windows/win32/directwrite/color-fonts#using-color-fonts-with-directwrite-and-direct2d.

This *should* give us colored bitmap, unless i am doing something wrong.

src/terminal_renderer/TextRenderer.cpp
christianparpart: has_color is an architectural work around that i have to really address and fix in the upper layer. I am sorry for this inconvenience. I will try to get rid of the need of this member function ASAP (maybe tonight).

src/contour/Config.cpp
```diff
@@ -1115,7 +1115,11 @@ TerminalProfile loadTerminalProfile(UsedKeys& _usedKeys,
     profile.fonts.boldItalic.slant = text::font_slant::italic;
     softLoadFont(_usedKeys, fontBasePath, _doc["profiles"][_name]["font"], "bold_italic", profile.fonts.boldItalic);
 
+#if _WIN32
+    profile.fonts.emoji.familyName = "Segoe UI Emoji";
```
christianparpart, line 1119: I rember that name, yeah. And did you succeed seeing colored emoji now?

src/text_shaper/directwrite_shaper.cpp
```diff
@@ -66,6 +66,60 @@ namespace
         }
         return font_slant::normal;
     }
+
+    void renderGlyphRunToBitmap(IDWriteGlyphRunAnalysis* _glyphAnalysis, RECT& _textureBounds, bitmap_format _targetFormat, std::vector<uint8_t>::iterator& _it)
```
christianparpart, line 70: I am on smartphone, how do I trigger a code change propose when not being able to press Ctrl+g? Hmm...

The parameter `_textureBounds` can be passed as const reference here.

As of https://docs.microsoft.com/en-us/windows/win32/api/dwrite/nf-dwrite-idwriteglyphrunanalysis-createalphatexture

src/text_shaper/directwrite_shaper.cpp
```diff
@@ -66,6 +66,60 @@ namespace
...
+    {
+        const auto width = _textureBounds.right - _textureBounds.left;
+        const auto height = _textureBounds.bottom - _textureBounds.top;
+
+        std::vector<uint8_t> tmp;
+        tmp.resize(height *width * 3);
```
christianparpart, line 76: Whitespace around binary operators. Probably just a typo. :)

src/text_shaper/directwrite_shaper.cpp
```diff
@@ -66,6 +66,60 @@ namespace
...
+
+        auto hr = _glyphAnalysis->CreateAlphaTexture(
+            DWRITE_TEXTURE_CLEARTYPE_3x1,
+            &_textureBounds,
+            tmp.data(),
+            tmp.size()
+        );
+
+        for (auto i = 0; i < height; i++)
```
christianparpart, line 85: Would you want to give `range::views::iota(0, N)` a try? Ah example use would be here: 
https://github.com/contour-terminal/contour/blob/abdae0192006275b31a6e2e5222755e5b1176924/src/terminal/Grid.cpp#L557-L258

It avoid the need of having the iterator variable being noon-const (mutable) even though it is only used read-only in the loop body.

src/text_shaper/directwrite_shaper.cpp
christianparpart: Oh this is certainly an interesting part of code that i have too sirens some more time in trying to understand ASAP. :)

src/text_shaper/directwrite_shaper.cpp
christianparpart: Yes. Colored emoji are placed into the scene as RGBA texture. In Freetype i have to convert from BGRA to RGBA too (i think OpenGL ES had Limited pixel format support). On Linux OpenGL ES (core Profile) is used by default.

christianparpart: BTW i just realize, that is a perfect opportunity for  providing a SIMD implementation for that task (you don't have to do that,l. I am just writing that one down here so it won't be forgotten).

src/text_shaper/directwrite_shaper.cpp
```diff
@@ -151,6 +206,7 @@ struct directwrite_shaper::Private
 
         auto key = create_font_key();
         fonts.emplace(pair{ key, move(fontInfo) });
+        fontsHasColor.emplace(pair{ key, false });
```
christianparpart, line 209: I will be trying to eliminate the need of the has_color member tonight

src/text_shaper/directwrite_shaper.cpp
skyline75489: I don't honestly know if this is the correct way to do so 😅  I don't have much graphic programming experience.

src/contour/Config.cpp
skyline75489: no i'm still seeing black and white, which confuses me. Could it be the OpenGL windows used is different?

src/text_shaper/directwrite_shaper.cpp
skyline75489: Ah thanks. It should.

src/text_shaper/directwrite_shaper.cpp
skyline75489: Totally a typo.

src/text_shaper/directwrite_shaper.cpp
skyline75489: OK this is not the correct way to composite the different layers. This is the correct way: https://github.com/qt/qtbase/blob/615a9cf99150cfee34c14cfed0aa9583f5993934/src/gui/text/windows/qwindowsfontenginedirectwrite.cpp#L849

But I don't really understand the process. Also I still can't get it working correctly...

src/text_shaper/directwrite_locator.cpp
```diff
@@ -0,0 +1,227 @@
...
+            UINT32 keySize;
+            fontFile->GetReferenceKey(&key, &keySize);
+            ComPtr<IDWriteLocalFontFileLoader> localLoader;
+            loader.As(&localLoader);
+
+            UINT32 pathLen;
+            localLoader->GetFilePathLengthFromKey(key, keySize, &pathLen);
+            std::wstring path;
+            path.resize(pathLen);
+            localLoader->GetFilePathFromKey(key, keySize, path.data(), pathLen + 1);
```
skyline75489, line 100: This returns the absolute path of the font. Same as font config locator.

src/text_shaper/directwrite_locator.cpp
```diff
@@ -0,0 +1,227 @@
...
+    {
+    }
+
+    font_source_list directwrite_locator::locate(font_description const& _fd)
+    {
+        font_source_list output;
+
+        // TODO: use libunicode for that (TODO: create wchar_t/char16_t converters in libunicode)
+        std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> wStringConverter;
+        std::wstring familyName = wStringConverter.from_bytes(_fd.familyName);
```
skyline75489, line 144: The conversion between UTF8 <-> UTF16 is just everywhere.....

src/text_shaper/directwrite_shaper.cpp
```diff
@@ -448,88 +369,18 @@ void directwrite_shaper::shape(font_key _font,
...
-            fontInfo.font->GetStyle(),
-            fontInfo.font->GetStretch(),
-            fontInfo.size.pt,
-            d->userLocale.c_str(),
-            &format);
-
-        ComPtr<IDWriteTextFormat1> format1;
-
-        if (SUCCEEDED(hr) && SUCCEEDED(format->QueryInterface(IID_PPV_ARGS(&format1))))
+        font_source_list sources = d->locator_->resolve(gsl::span(_text.data(), _text.size()));
```
skyline75489, line 375: Note that only DWrite locator implements `resolve`. So basically you'll see zero font fallback for CJK if using DWrite shaper + Font config locator. 

Also I think resolve here need to also pass `fontDescription` for bold/italic/etc.

src/text_shaper/open_shaper.cpp
```diff
@@ -477,7 +475,7 @@ namespace // {{{ helper
     }
 } // }}}
 
-struct FontInfo
+struct HbFontInfo
```
skyline75489, line 478: The rename is here is a must, because under `dwrite_shaper` there's struct with exact same name & namespace, which ruins the stack with mixing usage of shaper & engine.

src/text_shaper/directwrite_shaper.cpp
christianparpart: I will catch up with them font config "resolve" implementation ASAP :)

src/terminal_renderer/Renderer.cpp
```diff
@@ -81,7 +85,13 @@ unique_ptr<text::font_locator> createFontLocator(FontLocatorEngine _engine)
     switch (_engine)
     {
         // TODO: GDI / DirectWrite needs to be hooked in here
-
+        case FontLocatorEngine::DWrite:
+            #if defined(_WIN32)
+            return make_unique<text::directwrite_locator>();
+            #else
+            debuglog(TextRendererTag).write("Font locator CoreText not supported on this platform.");
```
christianparpart, line 92:
```suggestion
            debuglog(TextRendererTag).write("Font locator DirectWrite not supported on this platform.");
```

src/terminal_renderer/Renderer.cpp
```diff
@@ -107,7 +117,7 @@ unique_ptr<text::shaper> createTextShaper(TextShapingEngine _engine, crispy::Poi
             #if defined(_WIN32)
             debuglog(TextRendererTag).write("Using DirectWrite text shaping engine.");
             // TODO: do we want to use custom font locator here?
-            return make_unique<text::directwrite_shaper>(_dpi);
+            return make_unique<text::directwrite_shaper>(_dpi, std::move(_locator));
```
christianparpart, line 120:
```suggestion
            return make_unique<text::directwrite_shaper>(_dpi, move(_locator));
```

As `_locator` is of type `std::unique_ptr<>` I think ADL should take effect and `std::` is not needed. :)

src/text_shaper/directwrite_analysis_wrapper.h
```diff
@@ -0,0 +1,137 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include<string>
```
christianparpart, line 16:
```suggestion
#include <string>
```

src/text_shaper/directwrite_analysis_wrapper.h
```diff
@@ -0,0 +1,137 @@
...
+            text(_text),
+            userLocale(_userLocale)
+        {
+        }
+
+#pragma region IDWriteTextAnalysisSource
+        HRESULT GetTextAtPosition(
+            UINT32 textPosition,
+            _Outptr_result_buffer_(*textLength) WCHAR const** textString,
+            _Out_ UINT32* textLength)
```
christianparpart, line 41:
```suggestion
            _Out_ UINT32* textLength) override
```

I think you are implementing a set of interface(s) here. Every function that is being implemented should be marked as `override`. (Or is there some Win32-specific thing that prevents that?)

src/text_shaper/directwrite_locator.cpp
christianparpart: very nice!

src/text_shaper/directwrite_locator.cpp
```diff
@@ -0,0 +1,227 @@
...
+            factory->GetSystemFontFallback(&systemFontFallback);
+        }
+    };
+
+    directwrite_locator::directwrite_locator() :
+        d{ new Private(), [](Private* p) { delete p; } }
+    {
+    }
+
+    directwrite_locator::~directwrite_locator()
```
christianparpart, line 134: i think we don't need that dtor. :)

src/text_shaper/directwrite_locator.cpp
christianparpart: The layer above, that is using the <text_shaper> API is caching, so only uncache requests come here.

While that is good, I think we should try to get rid if unnecessary conversions, I agree. But that can be easily done in a separate PR that can be used for performance optimizations (such as this).

Also I have created a ticket in libunicode to fix UTF16/32 conversion. (the C++ API is actually deprecated in the C++ standard, I also do not know why, but yeah).

src/text_shaper/directwrite_locator.cpp
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    font_source_list directwrite_locator::resolve(gsl::span<const char32_t> codepoints)
+    {
+        font_source_list output;
+
+        std::wstring_convert<std::codecvt_utf8<char32_t>, char32_t> conv1;
+        std::string bytes = conv1.to_bytes(std::u32string{ codepoints.data(), codepoints.size() });
+        std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> conv2;
+        std::wstring wText = conv2.from_bytes(bytes);
```
christianparpart, line 193: um, it's converting from UTF-32 to UTF-8 to UTF-16? haha. That is hilarious!

No need to change that in this PR, as mentioned above, I'll fix libunicode first so we can do it in *one* pass.

src/text_shaper/directwrite_locator.cpp
```diff
@@ -0,0 +1,227 @@
...
+            &mappedLength,
+            &mappedFont,
+            &scale);
+
+        if (mappedFont)
+        {
+            ComPtr<IDWriteFontFace> fontFace;
+            mappedFont->CreateFontFace(&fontFace);
+
+            // TODO: use libunicode for that (TODO: create wchar_t/char16_t converters in libunicode)
```
christianparpart, line 219: :+1:

src/text_shaper/directwrite_shaper.cpp
```diff
@@ -178,204 +265,38 @@ struct directwrite_shaper::Private
...
-        *numberSubstitution = nullptr;
-        *textLength = text.size() - textPosition;
 
-        return S_OK;
-    }
-#pragma endregion
-
-#pragma region IDWriteTextAnalysisSink
-    HRESULT SetScriptAnalysis(UINT32 textPosition, UINT32 textLength, _In_ DWRITE_SCRIPT_ANALYSIS const* scriptAnalysis)
+    double ptToEm(double pt)
```
christianparpart, line 271:
```suggestion
    constexpr double ptToEm(double pt)
```

src/text_shaper/directwrite_shaper.cpp
```diff
@@ -448,88 +369,18 @@ void directwrite_shaper::shape(font_key _font,
...
-                std::wstring resolvedFamilyName;
-                resolvedFamilyName.resize(length);
-
-                familyNames->GetString(index, resolvedFamilyName.data(), length + 1);
-
-                _font = d->add_font(resolvedFamilyName, fontInfo.description, fontInfo.size, mappedFont);
-                fontInfo = d->fonts.at(_font);
-                fontFace = fontInfo.fontFace.Get();
+                _font = fontKeyOpt.value();
+                fontInfo = d->fonts.at(fontKeyOpt.value());
```
christianparpart, line 382:
```suggestion
                fontInfo = d->fonts.at(_font);
```

src/text_shaper/directwrite_shaper.cpp
```diff
@@ -616,9 +467,9 @@ void directwrite_shaper::shape(font_key _font,
 
 std::optional<rasterized_glyph> directwrite_shaper::rasterize(glyph_key _glyph, render_mode _mode)
 {
-    FontInfo const& fontInfo = d->fonts.at(_glyph.font);
-    IDWriteFontFace5* fontFace = fontInfo.fontFace.Get();
-    const float fontEmSize = _glyph.size.pt * (96.0 / 72.0);
+    DxFontInfo const& fontInfo = d->fonts.at(_glyph.font);
+    IDWriteFontFace5* fontFace = fontInfo.fontFace;
+    const float fontEmSize = d->ptToEm(_glyph.size.pt);
```
christianparpart, line 472: I just realize that ptToEm is a member of `d` (Private). I think that doesn't have to be as it's just transforming the input without any dependency. Maybe make it a free function and put it into the anonymous namespace?

src/text_shaper/directwrite_shaper.cpp
```diff
@@ -671,42 +522,83 @@ std::optional<rasterized_glyph> directwrite_shaper::rasterize(glyph_key _glyph,
 
     auto const width = output.size.width;
     auto const height = output.size.height;
```
christianparpart, line 524:
```suggestion
    auto const [width, height] = output.size;
```
since you use `auto` type deduction, using tuple decomposition to get straight access to the members isn't too far off. What do you think? :-)

src/text_shaper/directwrite_shaper.cpp
```diff
@@ -671,42 +522,83 @@ std::optional<rasterized_glyph> directwrite_shaper::rasterize(glyph_key _glyph,
...
 
-    hr = glyphAnalysis->CreateAlphaTexture(
-        DWRITE_TEXTURE_CLEARTYPE_3x1,
-        &textureBounds,
-        tmp.data(),
-        tmp.size()
-    );
+        if (hr == DWRITE_E_NOCOLOR)
+        {
+            output.bitmap.resize(*height * *width * 3);
```
christianparpart, line 542: Does this line generate any compiler warning about signedness? Currently the wrapped primitive types are using the element type `int`, but `vector::resize(..)` accepts `size_t` so it should, shouldn't it?

In that case, maybe like that?

```suggestion
            output.bitmap.resize(height.as<size_t>() * width.as<size_t>() * 3);
```

src/text_shaper/directwrite_shaper.cpp
```diff
@@ -671,42 +522,83 @@ std::optional<rasterized_glyph> directwrite_shaper::rasterize(glyph_key _glyph,
...
+            renderGlyphRunToBitmap(glyphAnalysis.Get(), textureBounds, DWRITE_COLOR_F{}, output.format, t);
 
-            *t++ = tmp[base];
-            *t++ = tmp[base + 1];
-            *t++ = tmp[base + 2];
+            return output;
         }
+        else
+        {
+            output.bitmap.resize(*height * *width * 4);
```
christianparpart, line 553: same here, if changed above.

src/text_shaper/directwrite_shaper.cpp
```diff
@@ -671,42 +522,83 @@ std::optional<rasterized_glyph> directwrite_shaper::rasterize(glyph_key _glyph,
...
+            d->fontsHasColor.at(_glyph.font) = true;
+            while (true)
+            {
+                BOOL haveRun;
+                glyphRunEnumerator->MoveNext(&haveRun);
+                if (!haveRun)
+                    break;
+
+                DWRITE_COLOR_GLYPH_RUN  const* colorRun;
+                glyphRunEnumerator->GetCurrentRun(&colorRun);
```
christianparpart, line 565: is it guaranteed that this function always writes to its output parameter?

src/text_shaper/directwrite_shaper.h
```diff
@@ -26,7 +26,7 @@ namespace text {
  */
 class directwrite_shaper : public shaper {
   public:
-    explicit directwrite_shaper(crispy::Point _dpi);
+    explicit directwrite_shaper(crispy::Point _dpi, std::unique_ptr<font_locator> _locator);
```
christianparpart, line 29:
```suggestion
    directwrite_shaper(crispy::Point _dpi, std::unique_ptr<font_locator> _locator);
```

src/text_shaper/font.h
```diff
@@ -119,6 +119,7 @@ struct font_description
 #if _WIN32
```
christianparpart, line 119:
```suggestion
#if defined(_WIN32)
```

At least that would be more consistent to the rest of the ifdef's to _WIN32 :)

src/text_shaper/fontconfig_locator.h
```diff
@@ -34,7 +34,7 @@ class fontconfig_locator: public font_locator
 
     font_source_list locate(font_description const& description) override;
     font_source_list all() override;
-    font_source_list resolve(gsl::span<char32_t> codepoints) override;
+    font_source_list resolve(gsl::span<const char32_t> codepoints) override;
```
christianparpart, line 37: oh thanks. of course it's const. sorry.

src/text_shaper/open_shaper.cpp
christianparpart: absolutely fine with me.

christianparpart (changes requested):
I was a _little_ more strict in reviewing. Most of it I think you can just take as suggestions and/or thoughts. :)

src/text_shaper/directwrite_shaper.cpp
skyline75489: yea it should be outside `Private`.

src/text_shaper/directwrite_locator.cpp
skyline75489: ah, good point.

src/text_shaper/font.h
skyline75489: Sure,

src/contour/Config.cpp
christianparpart: This is resolved now, right? I mean, then we should click on resolve here, I think. :)

src/text_shaper/directwrite_locator.h
```diff
@@ -0,0 +1,46 @@
...
+
+#include <text_shaper/font_locator.h>
+#include <text_shaper/font.h>
+
+class IDWriteFontFace;
+
+namespace text {
+
+    /**
+     * Font locator API implementation using `fontconfig` library.
```
christianparpart, line 25: s/fontconfig/DirectWrite/

src/text_shaper/directwrite_locator.h
```diff
@@ -0,0 +1,46 @@
...
+     *
+     * This should be available on all platforms.
```
christianparpart, line 27: Only on windows

src/text_shaper/directwrite_shaper.cpp
skyline75489: Added check for failed HR

src/text_shaper/directwrite_shaper.cpp
skyline75489: There's a lot of unsigned warnings actually... I plan to remove them in future PRs and turn more "warning as error" switch for MSVC build.

src/text_shaper/directwrite_shaper.cpp
skyline75489: I am too dumb to actually make it work 😅 . The compile hates me. I'll give it a try in future PRs.

PR #362, opened 2021-07-31
text shaper cleanup (DO remove shaper::has_color())

christianparpart:


christianparpart:
@skyline75489 the windows build is breaking here, but I think you have fixed that already in your PR. so I am merging here. :)

PR #364, opened 2021-08-02
Fix DWrite shaper fallback logic

skyline75489:
It should try to find fallback fonts *only* if the current font doesn't contain the correct glyphs.


christianparpart (approved):
Many thanks. :)

PR #366, opened 2021-08-02
the little things while trying to debug down has_color-elimination introduced emoji rendering

christianparpart:


src/terminal/pty/ConPty.cpp
```diff
@@ -95,6 +95,8 @@ ConPty::~ConPty()
 
 void ConPty::close()
 {
+    auto const _ = std::lock_guard{mutex_};
```
christianparpart, line 98: This fixes a crash at shutdown on Windows.

PR #370, opened 2021-08-03
Added dedicated bundle plist for macOS

skyline75489:
This will bring HiDPi window chrome in contour on macOS. 

See https://doc.qt.io/qt-5/highdpi.html


uspasojevic96:
I guess that I will define all these variables and remove unnecessary ones

PR #374, opened 2021-08-04
[text_shaper] DWrite minor cleanups.

christianparpart:


src/text_shaper/directwrite_shaper.cpp
```diff
@@ -361,25 +361,25 @@ void directwrite_shaper::shape(font_key _font,
             gpos.presentation = _presentation;
             gpos.glyph = glyph_key{ _font, fontInfo.size, glyph_index{glyphIndices.at(i)} };
             // gpos.offset.x = static_cast<int>(cellWidth);
-            //gpos.offset.y = static_cast<int>(static_cast<double>(pos[i].y_offset) / 64.0f);
+            // gpos.offset.y = static_cast<int>(static_cast<double>(pos[i].y_offset) / 64.0f);
```
skyline75489, line 364: maybe just remove these? if in the future this is needed, we can always add them back.

src/text_shaper/directwrite_shaper.cpp
```diff
@@ -587,23 +587,15 @@ std::optional<rasterized_glyph> directwrite_shaper::rasterize(glyph_key _glyph,
                     &colorGlyphsAnalysis);
 
                 auto t = output.bitmap.begin();
-                if (colorRun->paletteIndex == 0xFFFF)
-                {
-                    renderGlyphRunToBitmap(colorGlyphsAnalysis.Get(), textureBounds, DWRITE_COLOR_F{}, output.format, t);
```
skyline75489, line 592: Actually the correct color here should theortically be the current *text* color, but I don't have that information here. 

However, an empty `DWRITE_COLOR_F` also just works...

christianparpart: Hm, that is a good point. That information could be made available. But would lead to a more contended texture atlas as each glyph would be stored with all its individual colors, which might be expensive. The color is blended in at the fragment shader level. (An empty `DWRITE_COLOR_F{}` most likely means black, as all components will be set to zero, that is why it works).

skyline75489: >  The color is blended in at the fragment shader level

This is one of the reason why DWrite is super slow. It can't do mixed color drawing. The run has to break if there's color change. That is why termbench/refterm is killing the renderer in WT, because it prints a full screen of characters and each of them has a different color. That leads to 7000+ draw calls per frame and severely slows down the performance.

Just some fun fact 😄

christianparpart: Most of the performance in refterm though comes because of the broken design flaw in the scrollback buffer handling. But you mention a very interesting point. I won't aim for 7k fps in rendering in a terminal though. This is not even worth thinking about that :)

skyline75489:
Now that `has_color` is gone, all things `fontsHasColor` related can be safely removed, I assume? :)

christianparpart:
> Now that `has_color` is gone, all things `fontsHasColor` related can be safely removed, I assume? :)

yes, exactly. this can be included in a separated PR as it is touching more than just DWrite.

PR #418, opened 2021-09-19
Experimental: embedded FreeType and HarfBuzz

data-man:


CMakeLists.txt
```diff
@@ -91,6 +92,7 @@ message(STATUS "Enable with code coverage:          ${CONTOUR_CODE_COVERAGE_ENAB
 message(STATUS "Build contour frontend GUI:         ${CONTOUR_FRONTEND_GUI}")
 message(STATUS "Build contour using Qt 6:           ${CONTOUR_BUILD_WITH_QT6}")
 message(STATUS "Build contour using mimalloc:       ${CONTOUR_BUILD_WITH_MIMALLOC}")
+message(STATUS "Buildin FreeType and HarfBuzz:      ${CONTOUR_BUILD_WITH_HARFBUZZ}")
```
christianparpart, line 95: This is a little confusing. Contour if already using freetype and harfbuzz. 
So the steering `CONTOUR_BUILD_WITH_HARFBUZZ` reads as if you can also compile this project without using harfbuzz.

In the other hand: what is the benefit of ending the harfbuzz/freetype source code?

data-man: > This is a little confusing.

What about "Embedded FreeType and HarfBuzz"?

> what is the benefit of ending the harfbuzz/freetype source code?

Performance. I don't know what options were used to compile the system libraries.
I think `march=native` might improve something.

christianparpart: probably more something like "Build contour with system harfbuzz/freetype" `CONTOUR_USE_SYSTEM_HARFBUZZ` defaulting to `ON`, and if off, it'll use the embedded version to be statically linked.

System libraries are never built with `-march=native` as this is not portable. for local builds, sure. I mean, it used to be heavily buggy (just like O3), but my experience on that one is also 20 years old, I'm sure it has changed since then. :)

I am however also not sure about the place you've put this. contour itself doesn't care about freetype/harfbuzz, but the `terminal_renderer` library does (check `src/terminal_renderer/CMakeLists.txt`). So this should be moved then to that place, too.

christianparpart: p.s.: to keep it in line with the other messages, above, maybe have indeed the string like "Build contour using embedded harfbuzz & freetype" (then defaulting to off of course). :)

data-man:
Good night! :)

cmake/CPM.cmake
```diff
@@ -28,7 +28,7 @@
 
 cmake_minimum_required(VERSION 3.14 FATAL_ERROR)
 
-set(CURRENT_CPM_VERSION 0.33.0)
+set(CURRENT_CPM_VERSION 0.34.0)
```
christianparpart, line 31: I don't wanna be nit-picky but here a single commit is doing multiple things. In the long run, I prefer changes to be isolated from each other (one commit for CPM version upgrade, another commit for e.g. a feature addition, another commit for e.g. cmake options changes)

cmake/ThirdParties.cmake
```diff
@@ -1,7 +1,9 @@
 include(CPM)
 
 set(3rdparty_catch2_version "10fb93cce8cb878fdfe8faa118b531258db96a26" CACHE STRING "catch2: commit hash")
-set(3rdparty_fmt_version "d9fd695ac737f84f7de2d0a2aa346b25efb9afbf" CACHE STRING "fmt: commit hash")
+set(3rdparty_fmt_version "2742611cad4aee6b1a5638bd1ebf132908f4a3d9" CACHE STRING "fmt: commit hash")
```
christianparpart, line 4: also fmtlib is updated?

CMakeLists.txt
```diff
@@ -38,6 +38,7 @@ option(CONTOUR_FRONTEND_GUI "Enables GUI frontend." ON)
 option(CONTOUR_COVERAGE "Builds with codecov [default: OFF]" OFF)
 option(CONTOUR_SANITIZE "Builds with Address sanitizer enabled [default: OFF]" OFF)
 option(CONTOUR_STACKTRACE_ADDR2LINE "Uses addr2line to pretty-print SEGV stacktrace." ${ADDR2LINE_DEFAULT})
+option(CONTOUR_BUILD_WITH_EMBEDDED_HARFBUZZ "Builds with embedded FreeType and HarfBuzz [default: OFF]" OFF)
```
christianparpart, line 41:
```suggestion
option(CONTOUR_BUILD_WITH_EMBEDDED_FT_HB "Builds with embedded FreeType and HarfBuzz [default: OFF]" OFF)
```

I think it is best to have **both** libraries somehow referenced with their well known shortcuts (FT, HB). Otherwise it feels like magic that it's indicating not as much as it's inducing. :)

christianparpart (changes requested):
Hey, many thanks. I think the option name change is the only thing that is a little more important to me then the other change suggestions I did, they are more for future recommendations. :)

CMakeLists.txt
data-man: Done.

cmake/ThirdParties.cmake
data-man: I'll update it in other PR. :)

cmake/CPM.cmake
data-man: I'll update it in other PR. :)

PR #430, opened 2021-10-01
[CI] added KDE Neon options for install deps

uspasojevic96:


scripts/install-deps.sh
```diff
@@ -44,6 +44,9 @@ main_linux()
         Ubuntu)
```
christianparpart, line 44:
```suggestion
        Ubuntu|Neon)
```

christianparpart (approved):
LGTM

PR #439, opened 2021-10-03
Update deps

data-man:


christianparpart (approved):

data-man:
So CODEOWNERS dosn't works as I expected. :)

christianparpart:
> So CODEOWNERS dosn't works as I expected. :)

Such a file in a repo still looks like we know what we are doing haha.

What do you expect it to do?

data-man:
> What do you expect it to do?

That your approvals will not be required for CMake's files. :)

data-man:
Qt installation in Windows is frozen. :(

christianparpart:
> > What do you expect it to do?
> 
> That your approvals will not be required for CMake's files. :)

AFAICT, `CODEOWNERS` file let's you auto-assign reviewers automatically based on file path/prefix.

And contour currently requires review count of at least 1 for PR merges to be allowed and nothing can go into `master` without a PR.

I think it is good practice to require a review for code changes.

On the other hand, no one can seriously review my code changes. But I hope that can change - given enough time in the future ;)

christianparpart:
> Qt installation in Windows is frozen. :(

CI restarted.

data-man:
> let's you auto-assign reviewers automatically based on file path/prefix.

But I can't approve my own PRs, if I'm not mistaken. :)

christianparpart:
> > let's you auto-assign reviewers automatically based on file path/prefix.
> 
> But I can't approve my own PRs, if I'm not mistaken. :)

Exactly. Reviews wouldn't make sense at all if you (anyone) could.

Issue #376, opened 2021-08-04
How to set custom WM_CLASS


uunnxx:
Hi! Thank you for this awesome terminal emulator.
I've searched all closed and open issues for `WM_CLASS` and just for `class` but can't able to find it.
Let me know how can I set custom `WM_CLASS` for getting new instance?

Like `urxvt` :
```
urxvt -name float 
urxvt -name ranger
urxvt -name translator
```
or like `alacritty` : 
```
alacritty --class float
alacritty --class ranger
alacritty --class translator
```

Example `i3wm` config:
```
set $trans "alacritty --class translator -e trans :en jp: -I"
bindsym $mod+t exec --no-startup-id $trans


for_window [class="Alacritty" instance="translator"] floating enable resize set 880 400
for_window [class="Alacritty" instance="translator"] move position 1040 674
```

christianparpart:
Hey @uunnxx 

Many thanks for your interest. I think am am.... sure that we do not support that yet.

So let's use this ticked to implement that. :-)

PR #446, opened 2021-10-05
Add option to set a custom `WM_CLASS`

jubnzv:
Add the CLI option `class` and the configuration option `wm_class` to set the user-defined `WM_CLASS` property of the terminal window.

This could be tested using `xprop` utility, to test if we really modified `WM_CLASS`:
```
WM_CLASS(STRING) = "contour", "contour_ex"
```

I additionally tried to use this feature with my `i3wm` as it described in #376, using the following configuration:
```
for_window [class="contour_ex"] floating enable
```

It works as expected in both cases: when set using the CLI option or using a configuration file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ for power users with a modern feature mindset.
   Usage:
 
     contour [terminal] [config STRING] [profile STRING] [debug STRING] [live-config]
-                       [working-directory STRING] [PROGRAM ARGS...]
+                       [working-directory STRING] [class STRING] [PROGRAM ARGS...]
     contour help
     contour version
     contour parser-table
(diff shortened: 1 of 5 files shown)


christianparpart (commented):
Wow. Many thanks. That will definitely help the tiling WM friends :-)

LGTM 👍 But before I press approve, would you mind adding a Changelog entry for that? This is certainly worth mentioning.

christianparpart (approved):
To not waste time, i can also do touch the changelog when being home.

jubnzv:
Okay, thank you. Then I won't change it, so as not to don't break the CI pipeline.

uunnxx:
Many thanks to you guys <3

PR #447, opened 2021-10-05
Update FreeType and HarfBuzz

data-man:
Many important fixes.


christianparpart (approved):
I however wonder what "many important" all means. Maybe for the next one we could mention the jump from version X to version Y.

PR #459, opened 2021-10-14
Improvement/logging

christianparpart:


christianparpart:
> Wow, without 3rd-parties! :)

it is in fact not much. it's basically a rewrite of the older API (`<debuglog>`) but with the goal of being more customizable, especially with regards to reusing for non-debug logging, such as error messages and a hopefully more intuitive (aka. convenient) API :)

PR #497, opened 2021-11-14
Improving new window behavior

uspasojevic96:


src/contour/TerminalSession.cpp
```diff
@@ -819,15 +821,23 @@ void TerminalSession::executeAction(actions::Action const& _action)
 
 void TerminalSession::spawnNewTerminal(string const& _profileName)
 {
-    ::contour::spawnNewTerminal(
-        programPath_,
-        config_.backingFilePath.generic_string(),
-        _profileName,
-        [this]() -> string {
+    auto const wd = [this]() -> string {
```
christianparpart, line 824: I like that you actually adopted this style here ;)

src/contour/TerminalSession.cpp
```diff
@@ -819,15 +821,23 @@ void TerminalSession::executeAction(actions::Action const& _action)
...
             auto const _l = scoped_lock{terminal_};
             return terminal_.screen().currentWorkingDirectory();
-        }()
-    );
+        }();
+
+    std::cout << wd <<std::endl;
```
christianparpart, line 829: debug print should go away *or* being replaced with the debuglog API

src/contour/TerminalSession.cpp
```diff
@@ -819,15 +821,23 @@ void TerminalSession::executeAction(actions::Action const& _action)
...
+    if (config_.spawnNewProcess) {
+        ::contour::spawnNewTerminal(
+            programPath_,
+            config_.backingFilePath.generic_string(),
+            _profileName,
+            wd
+        );
+    } else {
+        config_.profile(profileName_)->shell.workingDirectory = FileSystem::path(wd);
```
christianparpart, line 838: This will change the WD also for the current instance. I'm not sure that's the best approach. What would work is that you create a copy of `config_`, then mutate and use that instead.

src/contour/TerminalSession.h
```diff
@@ -13,6 +13,7 @@
  */
 #pragma once
 
+#include "contour/Controller.h"
```
christianparpart, line 16:
```suggestion
#include <contour/Controller.h>
```

src/contour/TerminalSession.h
```diff
@@ -49,6 +50,7 @@ class TerminalSession: public terminal::Terminal::Events
                     bool _liveconfig,
                     std::string _profileName,
                     std::string _programPath,
+                    Controller &_controller,
```
christianparpart, line 53:
```suggestion
                    Controller& _controller,
```

src/contour/TerminalSession.h
```diff
@@ -166,6 +168,7 @@ class TerminalSession: public terminal::Terminal::Events
     std::string profileName_;
     config::TerminalProfile profile_;
     std::string programPath_;
+    Controller *controller_;
```
christianparpart, line 171:
```suggestion
    Controller& controller_;
```

I though highly recommend that to be a reference instead to give the guarantee that this will never be NULL.

src/contour/TerminalWindow.cpp
```diff
@@ -205,11 +205,12 @@ void ScrollableDisplay::onValueChanged()
 }
 #endif // }}}
 
-TerminalWindow::TerminalWindow(config::Config _config, bool _liveConfig, string _profileName, string _programPath) :
+TerminalWindow::TerminalWindow(config::Config _config, bool _liveConfig, string _profileName, string _programPath, Controller &_controller) :
```
christianparpart, line 208:
```suggestion
TerminalWindow::TerminalWindow(config::Config _config, bool _liveConfig, string _profileName, string _programPath, Controller& _controller) :
```

src/contour/TerminalWindow.cpp
```diff
@@ -205,11 +205,12 @@ void ScrollableDisplay::onValueChanged()
...
     config_{ std::move(_config) },
     liveConfig_{ _liveConfig },
     profileName_{ std::move(_profileName) },
-    programPath_{ std::move(_programPath) }
+    programPath_{ std::move(_programPath) },
+    controller_{ &_controller }
```
christianparpart, line 213:
```suggestion
    controller_{ _controller }
```
because of the pointer-to-ref suggestion above.

src/contour/TerminalWindow.cpp
```diff
@@ -251,6 +252,7 @@ TerminalWindow::TerminalWindow(config::Config _config, bool _liveConfig, string
         liveConfig_,
         profileName_,
         programPath_,
+        *controller_,
```
christianparpart, line 255:
```suggestion
        controller_,
```
same

src/contour/TerminalWindow.h
```diff
@@ -104,6 +105,7 @@ class TerminalWindow :
     const bool liveConfig_;
     std::string profileName_;
     std::string programPath_;
+    Controller *controller_;
```
christianparpart, line 108:
```suggestion
    Controller& controller_;
```

christianparpart (changes requested):
LGTM. but some suggestions to also improve code quality a bit. :)

src/contour/TerminalSession.cpp
uspasojevic96: Well in theory it won't change, because it's getting the working directory from that instance

christianparpart:
@uspasojevic96 Also - no need to force you upon this in this PR - but what about not having the proposed config flag globally but as part of the `NewTerminal` action in the input mapping at the bottom of the config file?

If so, then a config entry could look like this:

```yaml
- { mods: Meta, key: N, action: NewTerminal, shared_process: true }
```
with that flag defaulting to `false` for as long as we can't guarantee production-grade stability.

Have a look at the `NewTerminal` action, as it already takes one parameter, it shouldn't be too hard to add a second one. See Actions.h for extending the action, then Config.cpp for adding the extended parsing, then TerminalSession.cpp line 587 for making use of that added boolean.

The pro-side of that proposal is, that we can provide both functionalities in the same running process.

christianparpart (approved):

christianparpart (approved):
LGTM

PR #498, opened 2021-11-14
Adds process CWD on macOS

uspasojevic96:


Changelog.md
```diff
@@ -7,6 +7,7 @@
 - Adds new config option `spawn_new_process` to define new terminal window behavior.
 - Adds action `CancelSelection` to allow actively canceling selection via input mappings.
 - Adds key bindings to default configuration to allow simply pressing Ctrl+C/Ctrl+V (without Shift modifier) when an active selection is present.
+- Adds process CWD on macOS.
```
christianparpart, line 10: I think we should spell CWD out to have it more human-non-nerd readable

src/contour/TerminalSession.cpp
```diff
@@ -19,6 +19,10 @@
 #include <terminal/Terminal.h>
 #include <terminal/pty/Pty.h>
 
+#if defined(__APPLE__)
+#include <terminal/pty/PtyProcess.h>
+#include <terminal/pty/UnixPty.h>
+#endif
```
christianparpart, line 25:
```suggestion
#endif

```

src/contour/helper.cpp
```diff
@@ -264,14 +264,19 @@ void spawnNewTerminal(string const& _programPath,
 {
     auto const wd = [&]() -> QString {
         auto const url = QUrl(QString::fromUtf8(_cwdUrl.c_str()));
+#if defined(__APPLE__)
```
christianparpart, line 267: I don't understand this change. Why is this needed?

src/contour/helper.cpp
```diff
@@ -264,14 +264,19 @@ void spawnNewTerminal(string const& _programPath,
...
         if (url.host() == QHostInfo::localHostName())
             return url.path();
         else
             return QString();
+#endif
     }();
 
     QString const program = QString::fromUtf8(_programPath.c_str());
     QStringList args;
+    std::cout << "asd " << _cwdUrl <<std::endl;
```
christianparpart, line 279:
```suggestion
```
debug prints using `cout` :)

src/terminal/Process.cpp
```diff
@@ -427,15 +427,17 @@ vector<string> Process::loginShell()
...
+#if defined(__APPLE__)
+        auto shell = string(pw->pw_shell);
+        auto index = shell.rfind('/');
         return {
-            "/usr/bin/login",
-            "-fp",
-            pw->pw_name,
+            "/bin/bash",
+            "-c",
+            fmt::format("exec -a -{} {}", shell.substr(index + 1, 5), pw->pw_shell)
```
christianparpart, line 436: quickly afk. let me get a drink before accepting the existence of such a hack. 🍺

src/terminal/Process.cpp
```diff
@@ -457,8 +459,11 @@ FileSystem::path Process::homeDirectory()
         return FileSystem::path("/");
 #endif
 }
-
+#if defined(__APPLE__)
+string Process::workingDirectory(UnixPty const& _pty) const
```
christianparpart, line 463:
```suggestion
string Process::workingDirectory(UnixPty const* _pty = nullptr) const
```
shouldn't the parameter not made optional instead? such as above? So that the code that attempts to call this function doesn't need to have too many `#ifdef`'s?

src/terminal/Process.cpp
```diff
@@ -472,8 +477,24 @@ string Process::workingDirectory() const
 		// ignore failure, and use default instead.
 		return "."s;
 	}
+#elif defined(__APPLE__)
+    try {
+        struct proc_vnodepathinfo vpi;
```
christianparpart, line 482:
```suggestion
        auto  vpi = proc_vnodepathinfo{};
```
I don't mind not using the auto-way here, but I'd prefer to have at least the `struct` keyword removed. using `{}` in the end ensures initialization of all fields to their default values.

src/terminal/Process.cpp
```diff
@@ -472,8 +477,24 @@ string Process::workingDirectory() const
...
+
+        auto pid = tcgetpgrp(_pty.masterFd());
```
christianparpart, line 484:
```suggestion
        auto const pid = tcgetpgrp(_pty.masterFd());
```

src/terminal/Process.cpp
```diff
@@ -472,8 +477,24 @@ string Process::workingDirectory() const
...
+
+        auto ret = proc_pidinfo(pid, PROC_PIDVNODEPATHINFO, 0, &vpi, sizeof(vpi));
```
christianparpart, line 486:
```suggestion
        auto const ret = proc_pidinfo(pid, PROC_PIDVNODEPATHINFO, 0, &vpi, sizeof(vpi));
```
or simply `if (functionCall(...) < 0) ...`

christianparpart (changes requested):
actual logic LGTM, however.

src/contour/helper.cpp
uspasojevic96: because magically `url.host()` is always empty

christianparpart (approved):

PR #502, opened 2021-11-15
Don't use absolute path to contour in contour.desktop

data-man:
Many apps uses this way.
E.g. I **always** install my compiled binaries to `/usr/local/bin`.

### DIFF
--- a/src/contour/contour.desktop
+++ b/src/contour/contour.desktop
@@ -5,7 +5,7 @@
 [Desktop Entry]
 Type=Application
 TryExec=contour
-Exec=/usr/bin/contour
+Exec=contour
 Icon=contour
 Terminal=false
 Categories=Qt;System;TerminalEmulator;
@@ -193,5 +193,3 @@ Name[uk]=Відкрити нове вікно
 Name[x-test]=xxOpen a New Windowxx
 Name[zh_CN]=打开新窗口
 Name[zh_TW]=開啟新視窗
-Icon=window-new
-Exec=konsole


christianparpart:
> Many apps uses this way. E.g. I **always** install my compiled binaries to `/usr/local/bin`.

Exactly, and that's why both are wrong, because there might be a executable X in /bin, /usr/bin/, /usr/local/bin, and in $HOME/bin (I have a lot in $HOME/usr/bin for example).

I think the best solution is to use a generator and have it point to an absolute path but based on`${CMAKE_INSTALL_PREFIX}/bin`. This way we've fixed them all. ;)

src/contour/contour.desktop
```diff
@@ -5,7 +5,7 @@
 [Desktop Entry]
 Type=Application
 TryExec=contour
-Exec=/usr/bin/contour
+Exec=contour
```
christianparpart, line 8:
```suggestion
Exec=@CMAKE_INSTALL_PREFIX@/bin/contour
```
and then using a preprocessor / generator / `m4` to generate the actual .desktop file.

christianparpart (changes requested):

data-man:
> and that's why both are wrong

I strongly disagree, because hardcoded paths are evil! ;)
BTW, the recently added `contour-run.desktop` also uses only `contour`, so we have some inconsistency.

christianparpart:
> I strongly disagree, because hardcoded paths are evil! ;)

Is it really hard-coded if you generated it based on CMake configuration variables? :thinking:

christianparpart:
I just checked my .desktop files on my system. It's a mixed experience. So there's no common consensus either. But if you make it depending on CMAKE_INSTALL_PREFIX, you'd at least not depend on the users' correctly configured PATH env var.

data-man:
> Is it really hard-coded if you generated it based on CMake configuration variables?

Yes.
E.g. `PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/opt/sbin:/opt/bin` and `contour` installed to `/usr/bin`.

So I can create simple shell script `contour` in `/usr/local/bin with some `LD_PRELOAD` trick. It's flexible. :)

uspasojevic96:
> > Is it really hard-coded if you generated it based on CMake configuration variables?
> 
> Yes. E.g. `PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/opt/sbin:/opt/bin` and `contour` installed to `/usr/bin`.
> 
> So I can create simple shell script `contour` in `/usr/local/bin` with some `LD_PRELOAD` trick. It's flexible. :)

I am confused here, if you specify the install path and it generates it in `.desktop` how would that be hardcoded?

data-man:
>  it generates it in .desktop how would that be hardcoded?

Just like now:
```
...
Exec=/usr/bin/contour
...
```

christianparpart (approved):
Maybe I'm not clearly expressing myself. But I'm fine for not fully qualifying for now.

PR #535, opened 2021-12-09
feat(ci): add RPM packaging

### LINKED ISSUE #420: Feature request: RPM packaging
Hey, hope you're doing well!

As the title says I want to have a RPM package for contour. It would benefit users from Fedora/RHEL because there is people _like me_ who doesn't want to or can't manually build contour from source.

I know all the necessary build dependencies (including the extra ones) and I have a bit of experience packaging software using RPM. Did a tiny [spec file](https://gist.github.com/NTBBloodbath/4ffe36fe66efa79c93ce438407beb187) for compiling contour for testing purposes, here are the produced package files.

And the package while installing it

Note that there's still some work to be done, in Fedora we have a similar build system like PPAs in Ubuntu but we call them COPRs. I never uploaded a package to COPR so I don't know much about how it works so [here](https://docs.fedoraproject.org/en-US/quick-docs/publish-rpm-on-copr/) are the docs in case someone wants to maintain it before me haha.

I'll be glad to maintain the COPR repository for uploading the RPM packages later, but it's gonna take a bit because I need to learn how COPR tools works (e.g. [tito](https://github.com/rpm-software-management/tito)). However if someone wants to do it and already knows how to use COPR then feel free to upload it :)

Regards

NTBBloodbath:
Hey, hope you're doing well! :)

This PR aims to resolve #420 by adding RPM packaging for Fedora (didn't tried in openSUSE which also uses RPM packages but it should work _in theory_). I'm actually using a RPM package produced from my tests and it's working really good btw.

Regarding to structure I tried to keep the same hierarchy as ArchLinux builds, made a directory in `.github/` called `fedora` that contains the following files:
- `Dockerfile` - our Fedora docker container image used to build contour RPM package.
- `contour.spec` - our spec file used to compile contour as a RPM package (contains all the packaging logic).

Also I'm only packaging for `x86_64` arch at the moment, if it's requested then I could also add packaging for `arm64`/`aarch64`.

Please let me know if you want me to change something :)

Cheers

### DIFF
--- a/.github/fedora/Dockerfile
+++ b/.github/fedora/Dockerfile
@@ -1,16 +1,15 @@
-FROM fedora:35
+FROM fedora:latest
 
 WORKDIR /app
 
-# Update packages and install RPM development tools
-RUN dnf upgrade -y
+# Install Git and RPM development tools
 RUN dnf install -y \
-        gcc rpm-build rpm-devel make coreutils diffutils patch rpmdevtools
+        git gcc rpm-build rpm-devel make coreutils diffutils patch rpmdevtools desktop-file-utils
 
 # Install contour dependencies
 RUN dnf install -y \
         cmake extra-cmake-modules ninja-build gcc-c++ pkgconf freetype-devel harfbuzz-devel \
-        qt5-qtbase-devel qt5-qtbase-gui kf5-kwindowsystem-devel
+        qt5-qtbase-devel qt5-qtbase-gui kf5-kwindowsystem-devel fontconfig-devel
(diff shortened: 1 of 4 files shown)


christianparpart:
Amazing work, @NTBBloodbath . Many thanks for saving me some time. I really appreciate that. :)

I will look into it ASAP.

christianparpart:
As agreed on Discord, we're going to merge this even though the last commands don't work yet. We can take care of this in a secondary PR.

PR #538, opened 2021-12-10
make 3rdparty management more package maintainer friendly

christianparpart:
by decoupling fetching the sources from cmake.

The plan is to largely depend on system packages.
And where not possible, fallback to the current stat, but w/o CPM (maybe).


data-man:
> The plan is to largely depend on system packages.

https://github.com/cpm-cmake/CPM.cmake#cpm_use_local_packages

christianparpart:
There seemed to be issues with CPM on FreeBSD sadly though. I am yet to
install FreeBSD in a VM for testing out myself though.

But having decoupled download from config a little more prominently, in
whatever way, wouldn't be a too bad idea.

It can't be all installed via system packages anyways, so singer mechanism
has to still exist. But things like yaml-cpp for example to exist on all
platforms in a decent enough version.

The motivation is that CMake simply searches via find_package, and it would
be up to the package maintainer to provide the dependencies up-front.

herrhotzenplotz:
> > The plan is to largely depend on system packages.
> 
> https://github.com/cpm-cmake/CPM.cmake#cpm_use_local_packages

I may have forgot to mention that this feature appeared to be broken when I wrote the port. It attempted downloading dependencies during the configure stage which failed because 1) no network connectivity 2) no git. Despite the fact that the system wide dependencies were installed.

christianparpart:
> > > The plan is to largely depend on system packages.
> > 
> > https://github.com/cpm-cmake/CPM.cmake#cpm_use_local_packages
> 
> I may have forgot to mention that this feature appeared to be broken when I wrote the port. It attempted downloading dependencies during the configure stage which failed because 1) no network connectivity 2) no git. Despite the fact that the system wide dependencies were installed.

Thanks for your feedback :)
I will be trying to lay out as much as possible to system dependencies.

For libunicode + termbech-pro it won't be possible instantly at least. Catch2 I may need to downgrade the requirements maybe. But the rest IIRC should work (tested with Fedora and OSX/brew so far. Will evaluate Ubuntu next).

scripts/install-deps.sh
```diff
@@ -20,76 +18,201 @@ install_deps_ubuntu()
...
+
+    FULL_DISTFILE="$SYSDEPS_DIST_DIR/$DISTFILE"
+
+    if ! test -f "$FULL_DISTFILE"; then
+        if which curl &>/dev/null; then
+            curl -L -o "$FULL_DISTFILE" "$URL"
+        elif which wget &>/dev/null; then
+            wget -O "$FULL_DISTFILE" "$URL"
+        else
+            echo "Don't know how to fetch from the internet." 1>&2
```
herrhotzenplotz, line 122: For the future in case of FreeBSD, you can use `fetch(1)` which is included in the base system:

```
$ fetch <url> -o <target-file>
```

christianparpart: Nice. Thanks.

christianparpart: Yes. That's there plan. That this script ISD about to support FreeBSD, too.

christianparpart: done.

Issue #400, opened 2021-08-29
Missing contour-terminfo dependency in latest Ubuntu 20.04 package?


j4james:
## Description

I'm trying to install the latest release on Ubuntu 20.04 and it's failing with the error:
```
contour depends on contour-terminfo: however:
  Package contour-terminfo is not installed.
```

## Environment

- Contour Version or Git commit hash: 0.2.0.173
- Operating System (name / distribution / version): Ubuntu 20.04

## Steps to Reproduce

1. Download contour_0.2.0.173-ubuntu_20_04_amd64.deb from the [releases](https://github.com/contour-terminal/contour/releases) page.
2. Run `dpkg -i ./contour_0.2.0.173-ubuntu_20_04_amd64.deb`

It does basically install, but it reports the error mentioned above. Then any time I do something with apt-get it complains about the broken installation, and "repairing" it just uninstalls Contour.

I'm a bit clueless when it comes to Linux, so it's possible I'm just not installing it correctly, but I had no problems with the previous release version (0.1.1).

christianparpart:
Oh, many thanks. I will look into it. :) Did you try the PPA from https://launchpad.net/~christianparpart/+archive/ubuntu/contour-dev ? There the dependency should be present.

j4james:
> Did you try the PPA from https://launchpad.net/~christianparpart/+archive/ubuntu/contour-dev ? There the dependency should be present.

That worked. Thank you!

PR #539, opened 2021-12-11
debian: Adds missing terminfo file to .deb file.

### LINKED ISSUE #531: Can't install latest release
Tried installing on Ubuntu 20.4 with `apt install ...`. Apt informed me that I had an unmet dependency, `contour-terminfo`.

christianparpart:
Fixes #531


j4james:
Would this also be a fix for #400?

christianparpart:
@j4james Oh yes! Thanks for reminding me. I always had you in mind when thinking about the terminfo dependency issue.

I was just about to give another minor release and wanted to address this with it. :)

PR #551, opened 2021-12-20
[Github CI] Adds Spell checking.

christianparpart:


christianparpart:
closing until i came up with a better idea.

Issue #548, opened 2021-12-18
Linker failures


herrhotzenplotz:
First of all, thank you for the big build system refactors.

However, it doesn't quite do the job yet:

- GSL doesn't have a shared library to link against, thus providing it in `target_link_libraries` causes linker failures on FreeBSD at least.
- `yaml-cpp` doesn't seem to give us the absolute path to the shared object which causes it to not link properly since when it is installed from ports/binary packages it resides in `/usr/local/` and *not* in `/usr/`. Solution would be to use `find_library` to figure out the correct absolute link path.
- The `install-deps.sh` script seems to call `sudo` to install packages on all distros. We need `su root -c "pkg install ..."` on FreeBSD as an alternative because running the script as root seems like a strange idea.
- Said script tries to download GSL using curl/fetch despite it being already installed as a system dependency

Another strange one:

`src/terminal/Line.h:212` gives a compiler error like:
```console
In file included from /usr/home/nico/src/contour/src/terminal/MatchModes.cpp:2:
In file included from /usr/home/nico/src/contour/src/terminal/Terminal.h:17:
In file included from /usr/home/nico/src/contour/src/terminal/RenderBuffer.h:18:
In file included from /usr/home/nico/src/contour/src/terminal/Grid.h:17:
/usr/home/nico/src/contour/src/terminal/Line.h:215:34: error: member reference base type 'gsl::span' is not a structure or union
        return gsl::span(buffer_).subspan(unbox<size_t>(_start), unbox<size_t>(_count));
               ~~~~~~~~~~~~~~~~~~^~~~~~~~
```

When I do not inline the call to the span constructor, it works for some magic reason. No idea what is going on there.

See entire diff at https://gist.githubusercontent.com/herrhotzenplotz/2aca719d5133dcce2e2378043bd3ecc8/raw/f907f13eb2224c3c8d5823f7fef1c5bdc7e397da/diff

After all built, I hit a segfault in the LRU cache:
https://gist.githubusercontent.com/herrhotzenplotz/2aca719d5133dcce2e2378043bd3ecc8/raw/f907f13eb2224c3c8d5823f7fef1c5bdc7e397da/foobar.txt

This happens when you decrease the font size (Ctrl+Scrollwheel) and then run the `notcurses-demo`.

herrhotzenplotz:
For fun and profit I built again with a different LLVM version to test the compilation error:

```
nico@triton:~/src/contour $ /usr/bin/c++  --version
[...]

PR #562, opened 2021-12-28
(#548) Try to fix linker issues again

herrhotzenplotz:
mostly using the wrong library (GNU Scientific Library instead of the correct Microsoft.GSL one)


herrhotzenplotz:
I hope this fixes the linker and dependency confusions once and forever :smile:

Disabling the dependency to termbench when not building the tests is hopefully fine. Maybe that could be solved differently, but not within the scope of this PR I guess.

scripts/install-deps.sh
```diff
@@ -201,8 +204,8 @@ install_deps_FreeBSD()
         qt5-qmake \
         qt5-widgets \
         range-v3 \
+        yaml-cpp
```
christianparpart, line 207: Oh you have gotten that to work? That would be awesome. I prefer system deps whenever possible.

scripts/install-deps.sh
```diff
@@ -216,6 +219,7 @@ install_deps_arch()
         fmt \
         fontconfig \
         git \
+        gsl \
```
christianparpart, line 222: Many thanks for also taking care of other platforms. - A safe exit would be to always embed a dependency, but that should be last resort.

src/crispy/CMakeLists.txt
```diff
@@ -31,7 +31,7 @@ set(crispy_SOURCES
 add_library(crispy-core ${crispy_SOURCES})
 add_library(crispy::core ALIAS crispy-core)
 
-set(CRISPY_CORE_LIBS range-v3::range-v3 fmt::fmt-header-only unicode::core GSL)
+set(CRISPY_CORE_LIBS range-v3::range-v3 fmt::fmt-header-only unicode::core Microsoft.GSL::GSL)
```
christianparpart, line 34: And this works for all platforms? I hope CI will give us a :+1: here :)

christianparpart (approved):
> I hope this fixes the linker and dependency confusions once and forever 😄

I am praying for that one now, too and hope that CI will now pick up FreeBSD issues as early as possible, too.

> Disabling the dependency to termbench when not building the tests is hopefully fine. Maybe that could be solved differently, but not within the scope of this PR I guess.

Oh that is absolutely fine. I personally try to keep PRs as atomic (topic-wise) as possible, but I know very good that this is not always feasible. Keeping it practical sometimes wins. ;)

christianparpart:
FYI: the failing external tests is due to AddressSanatizer picking up a memory leak on libfontconfig. I have a fix for that that at least works locally on Ubuntu 21.10, I seem to need to fix either more memory leaks based on the Ubuntu 20.04 provided fontconfig or simply `fetch` the latest stable and patch that (I want to submit a patch for that fix somewhat soonish, too, so I don't have to keep patching other libs just to get CI green :) )

src/crispy/CMakeLists.txt
herrhotzenplotz: As far as I can see, this is the way how you're supposed to do it: https://github.com/Microsoft/GSL#usage-in-cmake

It built on all the platforms that the CI tests for :man_shrugging: I guess it is fine then.

Issue #563, opened 2021-12-28
Ubuntu 20.04 missing dependency


trappedinspacetime:
First of all, I congratulate you on your project.
I am on Ubuntu 20.04 platform. I cloned this repo and tried to build it via 

        dpkg-buildpackage -rfakeroot -b -uc -us

but I got the following dependency error:        
        
        CMake Error at cmake/ThirdParties.cmake:82 (find_package):
          By not providing "Findtermbench-pro.cmake" in CMAKE_MODULE_PATH this
          project has asked CMake to find a package configuration file provided by
          "termbench-pro", but CMake did not find one.
        
          Could not find a package configuration file provided by "termbench-pro"
          with any of the following names:
        
            termbench-proConfig.cmake
            termbench-pro-config.cmake
        
          Add the installation prefix of "termbench-pro" to CMAKE_PREFIX_PATH or set
          "termbench-pro_DIR" to a directory containing one of the above files.  If
          "termbench-pro" provides a separate development package or SDK, be sure it
          has been installed.
        Call Stack (most recent call first):
          CMakeLists.txt:81 (include)
        
        
I attempted to build and install   https://github.com/contour-terminal/termbench-pro but I hit a `fmt` bug https://github.com/contour-terminal/contour/issues/189
I tried to work around that bug by replacing  `fmt::fmt-header-only` with `fmt::fmt`. It's built finally but `Makefile` won't install the library.

How can I build it from sorce on Ubunto 20.04?

herrhotzenplotz:
Did you install the embedded libraries? See `scripts/install-deps.sh`

christianparpart:
There are also two ways to create a .deb package file. The one you are using is what I use for creating PPA builds. This requires pre-fetching the embedded dependencies, so yes, best choice is to invoke `scripts/install-deps.sh`.

The second way to create a .deb package file is to use CMake's provided `cpack` tool for it. Both ways are used by CI.

So depending on your usecase, you might also want to simply grab the latest stable(?) release from the github release page that contains as many prebuilt packages as possible (so far).

We could however improve the documentation to reflect the today's state of the building & packaging procedures.

Happy Xmas-days ;)

trappedinspacetime:
@herrhotzenplotz After running `scripts/install-deps.sh` it finished building successfully. But when I launch it, it crashes with the following error:

[...]

PR #565, opened 2021-12-29
[contour] Adds config key `renderer` to choose a renderer: default, software, OpenGL.

christianparpart:
Not sure how this is possible with Qt, but it would be useful (for customers) to determine if the underlying hardware OpenGL implementation is prior 3.0 and then just default-select software rendering.

Note, the current "software" renderer is making use of the system provided OpenGL software renderer.

Why is `Apple Software Renderer` causing a SEGV? ;-(


christianparpart:
Should help avoiding problems like mentioned in #563.

PR #568, opened 2021-12-31
random improvements, making notcurses happy (as always) ^^

christianparpart:
* [x] implements `DA3`
* [x] implements `SM`/`RM` `? 8452`
* [x] fixes images not always being cleared when writing text into cell (mmmh 😠)
* [x] adds `latest`-symlink to latest state dump directory when dumping state.
* [x] fixes false warning wrt `on_mouse_select`
* [x] fixes trying to create an empty config directory.

### DIFF
--- a/src/terminal/Sequencer.cpp
+++ b/src/terminal/Sequencer.cpp
@@ -1368,7 +1368,7 @@ ApplyResult Sequencer<T>::apply(FunctionDefinition const& _function, Sequence co
         // terminal identification, 4 hex codes
         screen_.reply("\033P!|C0000000\033\\");
         break;
-    case DCH: screen_.deleteCharacters(_seq.param_or<ColumnCount>(0, ColumnCount { 0 })); break;
+    case DCH: screen_.deleteCharacters(_seq.param_or<ColumnCount>(0, ColumnCount { 1 })); break;
     case DECCRA: {
         // The coordinates of the rectangular area are affected by the setting of origin mode (DECOM).
         // DECCRA is not affected by the page margins.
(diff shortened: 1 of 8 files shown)


src/terminal/Sequencer.cpp
```diff
@@ -1363,7 +1363,10 @@ ApplyResult Sequencer<T>::apply(FunctionDefinition const& _function, Sequence co
     case CUU: screen_.moveCursorUp(_seq.param_or<LineCount>(0, LineCount { 1 })); break;
     case DA1: screen_.sendDeviceAttributes(); break;
     case DA2: screen_.sendTerminalId(); break;
-    case DA3: return ApplyResult::Unsupported;
+    case DA3:
+        // terminal identification, 4 hex codes
+        screen_.reply("\033PC0000000\033\\");
+        break;
```
j4james, line 1369: FYI the `DECRPTUI` response sequence is identified by the characters `!|` (see https://vt100.net/docs/vt510-rm/DA3.html). So what you need here is something like this:

`"\033P!|C0000000\033\\"`

That's assuming you want `C0000000` as your "unit id".

christianparpart: damn, I should not code while watching TV. Thanks @j4james, and a happy new year. ;-)

christianparpart: Personally I really dislike `DA3`, but it seems to be queried by notcurses, so I thought it might still be good to have one, even though notcurses is using the (IMHO better) `XTVERSION`, which replies with terminal string (and version) string.

j4james: Happy New Year to you too!

PR #569, opened 2022-01-01
fixing render cache bloat - prep-work

christianparpart:
### the problem

* [ ] texture atlas keeps growing - is this good?
* [ ] cache key storage items keeps growing (should be GC'd)
* [ ] glyph-to-texture mapping keeps growing (should be GC'd)
* [ ] after running `notcurses-demo u` the TE becomes slow, probably due to the above bloat

- the cleanups should happen as part of the LRU cache.
- can we maintain a free-list for the texture atlas? sounds like a hard problem wrt efficient use of re-using the space
- ...

This is a PREP-WORKS-PR, the actual renderer changes (addressing the above) I'll do in a follow-up PR.

### DIFF
--- a/src/contour/Config.cpp
+++ b/src/contour/Config.cpp
@@ -945,21 +945,21 @@ void softLoadFont(UsedKeys& _usedKeys,
             _store.familyName = _node["family"].as<string>();
         }
 
-        if (_node["slant"].IsScalar())
+        if (_node["slant"] && _node["slant"].IsScalar())
         {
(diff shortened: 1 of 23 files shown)


klamonte:
Funny enough I had a similar problem. I saw the notcurses demo crash my X server from unbounded cache, and then the same with 'mpv -vo tct movie.mp4'.

I ended up making a rule to not cache cells that had images, or had 24-bit RGB colors for foreground / background. 🤷‍♀️

PR #583, opened 2022-01-11
Use pacman instead of packman

sebastianrakel:
I don't know if it's a typo, but all packages are in the official
repo, so we should using pacman


uspasojevic96 (approved):
Yes it was a typo, thanks!

PR #587, opened 2022-01-12
Add formatter

sebastianrakel:
On arch the compiler fails on missing formatter, this PR should provide them

### DIFF
--- a/src/terminal/InputGenerator.h
+++ b/src/terminal/InputGenerator.h
@@ -496,4 +496,68 @@ struct formatter<terminal::MouseTransport>
         return format_to(_ctx.out(), "<{}>", unsigned(_value));
     }
 };
+
+template <>
+struct formatter<terminal::Key>
+{
+    template <typename ParseContext>
+    constexpr auto parse(ParseContext& ctx)
+    {
+        return ctx.begin();
+    }
+    template <typename FormatContext>
+    auto format(terminal::Key _value, FormatContext& _ctx)
+    {
+        static constexpr auto mappings = std::array { "F1",
+                                                      "F2",
+                                                      "F3",
+                                                      "F4",
+                                                      "F5",
+                                                      "F6",
+                                                      "F7",
+                                                      "F8",
+                                                      "F9",
+                                                      "F10",
+                                                      "F11",
+                                                      "F12",
+                                                      "F13",
+                                                      "F14",
+                                                      "F15",
+                                                      "F16",
+                                                      "F17",
+                                                      "F18",
+                                                      "F19",
+                                                      "F20",
+                                                      "DownArrow",
+                                                      "LeftArrow",
+                                                      "RightArrow",
+                                                      "UpArrow",
+                                                      "Insert",
+                                                      "Delete",
+                                                      "Home",
+                                                      "End",
+                                                      "PageUp",
+                                                      "PageDown",
+                                                      "Numpad_NumLock",
+                                                      "Numpad_Divide",
+                                                      "Numpad_Multiply",
+                                                      "Numpad_Subtract",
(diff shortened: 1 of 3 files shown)


src/terminal/InputGenerator.h
```diff
@@ -496,4 +496,67 @@ struct formatter<terminal::MouseTransport>
...
+{
+    template <typename ParseContext>
+    constexpr auto parse(ParseContext& ctx)
+    {
+        return ctx.begin();
+    }
+    template <typename FormatContext>
+    auto format(terminal::Key _value, FormatContext& _ctx)
+    {
+      std::vector<std::string> key_values = {"F1","F2",
```
christianparpart, line 511:
```suggestion
      static constexpr auto mappings = std::array{"F1","F2",
```

src/terminal/InputGenerator.h
```diff
@@ -496,4 +496,67 @@ struct formatter<terminal::MouseTransport>
...
+	"Numpad_1",
+	"Numpad_2",
+	"Numpad_3",
+	"Numpad_4",
+	"Numpad_5",
+	"Numpad_6",
+	"Numpad_7",
+	"Numpad_8",
+	"Numpad_9"};
+      return format_to(_ctx.out(), key_values[unsigned(_value)]);
```
christianparpart, line 559:
```suggestion
      return format_to(_ctx.out(), mappings.at(static_cast<size_t>(_value)));
```

src/terminal/InputGenerator.h
```diff
@@ -496,4 +496,67 @@ struct formatter<terminal::MouseTransport>
...
+    template <typename ParseContext>
+    constexpr auto parse(ParseContext& ctx)
+    {
+        return ctx.begin();
+    }
+    template <typename FormatContext>
+    auto format(terminal::Key _value, FormatContext& _ctx)
+    {
+      std::vector<std::string> key_values = {"F1","F2",
+	"F3",
```
christianparpart, line 512: You can try running clang-format over this file to fix the indentation here. 

The command on the CLI would be something like: `clang-format -i  ./src/terminal/InputGenerator.h`

christianparpart (changes requested):
Looks good. But maybe you want to try avoid unnecessary memory allocations by avoiding to use `std::vector` and use `std::array` instead.

Using member `at(index)` over `operator[](index)` also ensures you do not accidentally overflow as the `at(index)` member function will throw if out of bounds (the other will not).

christianparpart:
FYI: I just realized that these are due to fmtlib being too new on Arch Linux. I just ran into the same problem locally when upgrading the Ubuntu 21.10 to 22.04 (dev). Do you want to finish this? Then just go ahead an remove the "WIP" prefix when done // mark this PR as "ready for review". :)

sebastianrakel:
> FYI: I just realized that these are due to fmtlib being too new on Arch Linux. I just ran into the same problem locally when upgrading the Ubuntu 21.10 to 22.04 (dev). Do you want to finish this? Then just go ahead an remove the "WIP" prefix when done // mark this PR as "ready for review". :)

I can try to fix the others but maybe i need some help

christianparpart:
> > FYI: I just realized that these are due to fmtlib being too new on Arch Linux. I just ran into the same problem locally when upgrading the Ubuntu 21.10 to 22.04 (dev). Do you want to finish this? Then just go ahead an remove the "WIP" prefix when done // mark this PR as "ready for review". :)
> 
> I can try to fix the others but maybe i need some help

just ping us on Discord #contour-dev, or say up until which point you want to do it, and I can do the rest :)

christianparpart (approved):
LGTM. Many thanks @sebastianrakel

PR #598, opened 2022-02-04
Gamepad input bindings

christianparpart:
# input mapping for gamepads

Button identification names:

```
Gamepad_<DeviceId>_<ButtonName>
Gamepad_<PadName>_<ButtonName>
```
DeviceId is the identifying the gamepad. This works because this
value is stable until the user replugs it into a different USB slot.

ButtonName is something like:

     L1, L2, L3, R1, R2, R3, Up, Down, Left, Right, Center, Guide.

How to deal with axis events? Would be nice to map my thrust levers
(or rudder pedals or joystick for mouse/cursor movements).

Maybe something like this:
```
Gamepad_<DeviceId>_<ButtonName>_<Value>
Gamepad_<DeviceId>_<ButtonName>_<MinValue>_<MaxValue>
```

With `MinValue`/`MaxValue` not between 0.0 and 1.0 but between 0 and 100.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -91,7 +91,7 @@ IncludeCategories:
     Priority:  10
   - Regex:     '^<harfbuzz/'
     Priority:  11
-  - Regex:     '^<(QtCore|QtGui|QtWidgets)/'
+  - Regex:     '^<(QtCore|QtGamepad|QtGui|QtWidgets)/'
     Priority:  12
   - Regex:     '^<catch2/'
     Priority:  13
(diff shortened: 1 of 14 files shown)


christianparpart:
Okay, that's been REALLY FUN to play around with. But I think I'll have to give it a close-action.
QtGamepad is not ported to Qt6 yet (will it ever be??!?) and QtGamepad doesn't support all of my gamepad hardware (my HOTAS isn't detected, nor the Rudder).

PR #615, opened 2022-02-25
Adds an option to hide title bar

### LINKED ISSUE #379: Configuration to disable titlebar on macos
## Abstract
In current versions there is a titlebar in the macos version being able to disable it will be better

## Motivation
Other terminals like alacritty, kitty and iTerm have the option to disable titlebar(and minimize maximize and close options) being able to remove either titlebar and even the window decorators would be nice

Utkarsh-khambra:
Adds the config option `show_title_bar` in profile, which when set to false disables title bar. Fixes #379.


christianparpart:
Oh wow, nice, many thanks @Utkarsh-khambra.

I have an idea, if you are up to it. What about having the ability to also toggle that via an input action?
I am not sure if Qt even does support changing that at runtime, I'd need to figure that out myself, too. I will be having a closer look at your PR ASAP.

p.s.: I think that's worth a changelog entry. :)

Utkarsh-khambra:
Thanks for such nice words.
I like the idea to toggle title bar at runtime, I'll see we can do it with Qt.

christianparpart:
Hey @Utkarsh-khambra - sorry for the late response, as said in the other ticket again, i had a kind of unexpected week and could not do anything.

I think yes, you're right wrt y-position of the window. but maybe you could find that out by looking at the global screen coordinates, or simply remember in case of repeated toggling.

Nevertheless, I think that's a minor cosmetic thing and can be fix that can very well be done in its own PR, so I can merge this and you don't need to wait for too long on this one to be merged :)

Utkarsh-khambra:
> I think yes, you're right wrt y-position of the window. but maybe you could find that out by looking at the global screen coordinates, or simply remember in case of repeated toggling.

I will try to it

christianparpart:
> > I think yes, you're right wrt y-position of the window. but maybe you could find that out by looking at the global screen coordinates, or simply remember in case of repeated toggling.
> 
> I will try to it

I do not want to give you the feeling that I'm ignoring me. You'll tell me if I can "just merge" it or when you've finished trying to address the vertical-offset, okay? :)

Utkarsh-khambra:
I did fix vertical offset issue, at least it works on my machine(linux) now :). Though I don't have others to test right now.

christianparpart:
> I did fix vertical offset issue, at least it works on my machine(linux) now :). Though I don't have other OSs to test right now.

i can test. Many thanks ! :)

christianparpart (approved):
Many thanks :)

Utkarsh-khambra:
You're welcome :-)

PR #616, opened 2022-02-28
[contour] Fixes initial window size on Ubuntu 18.04 (/old Qt).

christianparpart:
@whisperity can you confirm that this is fixing your initial window-size problem?

### DIFF
--- a/src/contour/opengl/TerminalWidget.cpp
+++ b/src/contour/opengl/TerminalWidget.cpp
@@ -253,7 +253,18 @@ TerminalWidget::TerminalWidget(TerminalSession& session,
     adaptSize_ { std::move(adaptSize) },
     enableBlurBehind_ { std::move(enableBackgroundBlur) },
     lastFontDPI_ { fontDPI() },
-    renderer_ {},
+    renderer_ {
+        terminal().pageSize(),
+        sanitizeFontDescription(profile().fonts, fontDPI()),
+        terminal().screen().colorPalette(),
+        profile().backgroundOpacity,
+        session_.config().textureAtlasHashtableSlots,
+        session_.config().textureAtlasTileCount,
+        session_.config().textureAtlasDirectMapping,
+        profile().hyperlinkDecoration.normal,
+        profile().hyperlinkDecoration.hover
+        // TODO: , WindowMargin(windowMargin_.left, windowMargin_.bottom);
+    },
     filesystemWatcher_(this)
 {
     initializeResourcesForContourFrontendOpenGL();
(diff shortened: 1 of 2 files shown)


whisperity:
Roger.

(But I have to say I **really** do not like how the code looks in this patch...)

src/contour/opengl/TerminalWidget.cpp
```diff
@@ -491,17 +497,11 @@ void TerminalWidget::initializeGL()
...
-                         session_.config().textureAtlasDirectMapping,
-                         profile().hyperlinkDecoration.normal,
-                         profile().hyperlinkDecoration.hover
-                         // TODO: , WindowMargin(windowMargin_.left, windowMargin_.bottom);
-    );
+#if defined(CONTOUR_INIT_RENDERER_EARLY)
+    applyFontDPI();
+#else
+    initializeRenderer();
+#endif
```
whisperity, line 504: What is the problem with spawning the renderer early in every case? This macro magic is ***really*** weird. Also, why `#if defined` instead of `#ifdef`?

src/contour/opengl/TerminalWidget.cpp
```diff
@@ -105,6 +105,8 @@ using std::chrono::steady_clock;
 
 using actions::Action;
 
+#define CONTOUR_INIT_RENDERER_EARLY 1
```
whisperity, line 108: Hardcoded macro? Do you intend this to become a build-time (CMake) configuration parameter?

christianparpart: Hardcoded because I wasn't sure if that's the root cause. With your confirmation I'll be refactoring it such that it'll be always initialized early (I think).

src/contour/opengl/TerminalWidget.cpp
christianparpart: I prefer `#if defined` over `#ifdef` as it's more expressive. Apart from that, nothing else. :-)

src/contour/opengl/TerminalWidget.cpp
christianparpart: > Do you intend this to become a build-time (CMake) configuration parameter?
I still have to confirm with (especially) Mac not being broken by this change. In worst case it'll become an `#if __APPLE__` rather a cmake option. but I hope I don't need to.

christianparpart:
> (But I have to say I **really** do not like how the code looks in this patch...)

That's exactly the honesty I like. Given that, first step was simply to get your confirmation that it works. Now I can refactor it such that it looks good.

NB: The reason why initialization of the `Renderer` was postponed into the call of `initializeGL` is the fact that Qt's `devicePixelRatio()` reports the wrong value before the call to `initializeGL` (i.e. from within the ctor). And that means that potentially the DPI has to be applied twice. first with the wrong value (1) and later with the actual one (may be 1, but may very well be 2.5 or whatever).

PR #623, opened 2022-03-13
Thirdparties: fix finding HarfBuzz

herrhotzenplotz:
This caused builds of contour for the FreeBSD ports to fail. I am not too sure why the CI on github didn't catch that.

Anyways, now CMakes complaint about HarfBuzz should be gone.


herrhotzenplotz:
Pardon me while I try to look into this mess. There seems to be something awfully wrong here.

herrhotzenplotz:
Ok, the failure for Ubuntu with embedded HarfBuzz and Freetype seems to not be my fault: 

https://github.com/contour-terminal/contour/runs/5527782304?check_suite_focus=true

This is an old run and it is NOT picking up the embedded libraries, but goes about using the system packages (see config summary).

The new FindHarfBuzz.cmake is stolen from https://github.com/mosra/magnum-plugins/blob/master/modules/FindHarfBuzz.cmake . 
I don't know how you want this to be credited, so please give me some feedback on that.

christianparpart (approved):
LGTM. I however trust you on the BSD part a bit. I am having a VM for that in the far distance, but I'm stuck with my laptop for the time being.

@herrhotzenplotz is it okay to merge this PR or are there further changes coming?

christianparpart:
```
/home/runner/work/contour/contour/_deps/sources/freetype-cff026d41599945498044d2f4dcc0e610ffb6929/src/autofit/afshaper.h:28:10: fatal error: hb.h: No such file or directory
   28 | #include <hb.h>
      |          ^~~~~~
compilation terminated.
```

hmmm.... This is an excerpt from https://github.com/contour-terminal/contour/runs/5528949157?check_suite_focus=true#step:7:22

@herrhotzenplotz do you actually build with embedded HB/FT on your local BSD or do you use the system provided ones?

I'm inclined of just getting rid of the embedded HB/FT option otherwise (in order to improve build reliability and reproducability)

herrhotzenplotz:
> ```
> /home/runner/work/contour/contour/_deps/sources/freetype-cff026d41599945498044d2f4dcc0e610ffb6929/src/autofit/afshaper.h:28:10: fatal error: hb.h: No such file or directory
>    28 | #include <hb.h>
>       |          ^~~~~~
> compilation terminated.
> ```
> 
> hmmm.... This is an excerpt from https://github.com/contour-terminal/contour/runs/5528949157?check_suite_focus=true#step:7:22
> 
> @herrhotzenplotz do you actually build with embedded HB/FT on your local BSD or do you use the system provided ones?

I am testing this right now and trying it to actually get it to work here. In the previous builds (current master e.g.), it linked against the system libraries despite me telling it to use the embedded ones:

```console
nico@triton:~/src/contour $ file build/src/contour/contour
build/src/contour/contour: ELF 64-bit LSB executable, x86-64, version 1 (FreeBSD), dynamically linked, interpreter /libexec/ld-elf.so.1, for FreeBSD 13.0 (1300139), FreeBSD-style, with debug_info, not stripped
nico@triton:~/src/contour $ ldd build/src/contour/contour | grep harfbuzz
	libharfbuzz.so.0 => /usr/local/lib/libharfbuzz.so.0 (0x800ea9000)
nico@triton:~/src/contour $ nm build/src/contour/contour | grep hb_ft
                 U hb_ft_font_create
                 U hb_ft_font_create_referenced
nico@triton:~/src/contour $ 
```

As I am working on the solution rn, to the problem feels more and more like a hack tbh.

> I'm inclined of just getting rid of the embedded HB/FT option otherwise (in order to improve build reliability and reproducability)

That would be an obvious solution. If somebody wants to have it embedded, they can obviously build a static version of harfbuzz and freetype themselves and link against that.

herrhotzenplotz:
Before merging, lemme please revert the last commit. It was a WIP to test the build on CI.

herrhotzenplotz:
I have pushed some more changes here https://github.com/herrhotzenplotz/contour/commit/72b301c689e7e8ee1fe278cbbd5bff77cd1e87bf

This fixes the embedded build of both HarfBuzz and Freetype. However, in the process CMake picks up the shared objects provided by the system and puts them into the required dynamic dependencies of the built executable. Problem is very likely Freetype's CMakeLists.txt.

With all that in mind, I think the way going forward is indeed to just remove the build option for embedded HB and FT and instead offering to link them statically (if possible). That would remove the dynamic dependencies.

herrhotzenplotz:
@christianparpart  I have done some more (non-successful) testing. Unless my patch above is not worth committing here as well and you still want to remove the embedded builds, I'd be fine if this gets merged.

Also, I have sent you some screenshots the other day via e-mail - just wanted to remind you as I haven't received an answer :smiley:

PR #625, opened 2022-03-18
Drop build-time option to embed Harfbuzz+FreeType.

christianparpart:
@data-man I've finally decided to drop that compile-time option, as it's a valid argument to say that users (or package maintainers) wishing to statically link Harfbuzz and/or Freetype into the binary, can do so by simply providing them on the system.

/cc @herrhotzenplotz FYI :)


data-man:
I'm using Devuan now.
$ apt info libharfbuzz-dev

> Version: 2.7.4-1

But HarfBuzz 4.0.1 was released.

christianparpart:
> I'm using Devuan now. $ apt info libharfbuzz-dev
> 
> > Version: 2.7.4-1
> 
> But HarfBuzz 4.0.1 was released.

I didn't want to believe the version range difference and installed a docker image of Devuan. And then even fetched the exact source packages of harfbuzz. That is from year 2017 (6 years old). That is literally even older than Ubuntu 18.04 LTS, which is already giving me a hard time to support.

What could be done is to specifically address Devuan by providing .deb files made for Devuan. So in order to not pollute the global scope of the project (objectively speaking here), I'd suggest to have a .deb package build script for Devuan OS instead that creates the contour-for-devuan.deb package deals with this situation exclusively.

data-man:
No problem, you are boss! :)
I'll switch to CPM again (or to git submodules) in my local sources.

PR #651, opened 2022-04-23
[DRAFT] Resuming sessions

Utkarsh-khambra:
Hi,  writing back cells to terminal is still not correct.

### DIFF
--- a/src/contour/TerminalSession.cpp
+++ b/src/contour/TerminalSession.cpp
@@ -377,11 +377,11 @@ void TerminalSession::onClosed()
         auto activeProfileName = profileName_ + "\n";
         file.write(configPath.data(), configPath.size());
         file.write(activeProfileName.data(), activeProfileName.size());
-        auto& grid = terminal().primaryScreen().grid();
+        const auto& grid = terminal().primaryScreen().grid();
         auto result = std::stringstream {};
         auto writer = VTWriter(result);
-        auto lines = *terminal_.state().pageSize.lines;
-        for (int const line: ranges::views::iota(0, lines))
+        for (int const line:
+             ranges::views::iota(-unbox<int>(grid.historyLineCount()), unbox<int>(grid.pageSize().lines)))
         {
             writer.write(grid.lineAt(LineOffset(line)));
             writer.crlf();
--- a/src/contour/TerminalSession.cpp
+++ b/src/contour/TerminalSession.cpp
@@ -22,6 +22,7 @@
 #include <terminal/pty/Pty.h>
 
 #include <crispy/StackTrace.h>
+#include <crispy/stdfs.h>
 
 #include <range/v3/all.hpp>
 
@@ -43,6 +44,8 @@
 #include <algorithm>
 #include <fstream>
 
+#include "fmt/core.h"
+#include "terminal/Grid.h"
 #include <QtNetwork/QHostInfo>
 
 #if !defined(_WIN32)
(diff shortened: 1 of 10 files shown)


christianparpart:
Thanks. I will try ASAP! And then give feedback and hopefully hints :).

src/contour/TerminalWindow.cpp
```diff
@@ -17,6 +17,7 @@
 #include <contour/TerminalWindow.h>
 #include <contour/helper.h>
 
+#include "crispy/utils.h"
```
christianparpart, line 20:
```suggestion
#include <crispy/utils.h>
```

src/contour/TerminalSession.cpp
```diff
@@ -358,6 +363,28 @@ void TerminalSession::onClosed()
...
+        file.write(activeProfileName.data(), activeProfileName.size());
+        const auto& grid = terminal().primaryScreen().grid();
+        auto result = std::stringstream {};
+        auto writer = VTWriter(result);
+        for (int const line:
+             ranges::views::iota(-unbox<int>(grid.historyLineCount()), unbox<int>(grid.pageSize().lines)))
+        {
+            writer.write(grid.lineAt(LineOffset(line)));
+            writer.crlf();
+        }
```
christianparpart, line 384: This fits very good into a public API function that can then be also unit-testable (or simply used by unit tests if fitting).
The frontend logic would then simply call this function to serialize the grid contents.

src/crispy/utils.h
```diff
@@ -11,9 +11,11 @@
 #include <fstream>
 #include <functional>
 #include <optional>
+#include <pwd.h>
 #include <sstream>
 #include <string>
 #include <string_view>
+#include <unistd.h>
```
christianparpart, line 18: These two includes most likely do not exist on every platform (Windows).

src/contour/TerminalWindow.cpp
```diff
@@ -67,6 +70,26 @@ namespace contour
 
 using actions::Action;
 
+static auto loadSessionFile(FileSystem::path sessionFilePath)
```
christianparpart, line 73: I think it would make sense to even create a tiny struct here insteaad of returning a 3-tuple with each element type being the same. That could be easily mixed up accidentally. What about:

```cpp
struct SessionState
{
    std::string configPath;
    std::string profileName;
    std::string gridBuffer;
};
```

src/contour/TerminalWindow.cpp
```diff
@@ -107,6 +130,15 @@ TerminalWindow::TerminalWindow(std::chrono::seconds _earlyExitThreshold,
     if (config_.maxImageSize.height <= Height(0))
         config_.maxImageSize.height = defaultMaxImageSize.height;
     // }}}
+    if (config_.profile().sessionResume)
+    {
+        auto [configPath, profile, gridBuffer] = loadSessionFile(crispy::xdgStateHome() / "contour/session");
```
christianparpart, line 135: What if you had 3 terminal windows open, log out and back in again, then session restore kicks back in. Are all 3 windows being restored properly?

I think what is of interest is:

- https://doc.qt.io/qt-5/session.html
- https://doc.qt.io/qt-5/qguiapplication.html#sessionId

So if for example you're running a KDE desktop (not sure about other Linux desktops, never tried), and you had 3 contour processes open, they're all going to be respawned again by KDE desktop. Each having a session id passed via CLI (Contour actually properly passes this to underlying Qt's Application). 
My thinking is, that passed session Id could then be used to properly handle multiple stored sessions.

christianparpart:
How on earth did the bottom two commits end up in here? Can you `drop` them? :)

Utkarsh-khambra:
@christianparpart I think PR is ready. Need your review on it.

christianparpart:
merging to `edge` as intermediate branch, taking over, and finishing it. Many thanks @Utkarsh-khambra :)

Issue #689, opened 2022-05-10
FreeBSD build is broken


herrhotzenplotz:
```console
[ 56%] Building CXX object src/terminal/CMakeFiles/terminal.dir/pty/UnixPty.cpp.o
cd /usr/home/nico/src/contour/build/src/terminal && /usr/bin/c++ -DBOOST_STACKTRACE_USE_ADDR2LINE=1 -DCONTOUR_STACKTRACE_ADDR2LINE=1 -DFMT_HEADER_ONLY=1 -DHAVE_CXXABI_H -DHAVE_DLADDR -DHAVE_DLFCN_H -DHAVE_DLSYM -DHAVE_EXECINFO_H -DHAVE_SYS_
SELECT_H -DHAVE_UNWIND_H -DLIBTERMINAL_LOG_TRACE=1 -DLIBTERMINAL_NAME=\"contour\" -DLIBTERMINAL_PTY_BUFFER_OBJECTS=1 -DLIBTERMINAL_VERSION_MAJOR=0 -DLIBTERMINAL_VERSION_MINOR=3 -DLIBTERMINAL_VERSION_PATCH=2 -DLIBTERMINAL_VERSION_STRING=\"0.
3.2-unreleased-master-b1e39f25\" -I/usr/home/nico/src/contour/src -I/usr/home/nico/src/contour/_deps/sources/libunicode-bcd20f35d72bacd163fa45e985cf67fd2b6a3d64/src -isystem /usr/local/include -O2 -g -DNDEBUG -fdiagnostics-color=always -mae
s -std=c++17 -MD -MT src/terminal/CMakeFiles/terminal.dir/pty/UnixPty.cpp.o -MF CMakeFiles/terminal.dir/pty/UnixPty.cpp.o.d -o CMakeFiles/terminal.dir/pty/UnixPty.cpp.o -c /usr/home/nico/src/contour/src/terminal/pty/UnixPty.cpp
In file included from /usr/home/nico/src/contour/src/terminal/pty/UnixPty.cpp:15:
/usr/home/nico/src/contour/src/terminal/pty/UnixUtils.h:21:14: fatal error: 'pty.h' file not found
    #include <pty.h>
             ^~~~~~~
```

I remember fixing this one before. Also, all the CI builds for FreeBSD seem to be broken. :-(

PR #690, opened 2022-05-10
Fix broken FreeBSD builds

herrhotzenplotz:
See #689


christianparpart (approved):
Oh, nice. thanks! Now we only need to get FreeBSD CI green again (I think that's a CI issue however, but it doesn't help finding build errors if it's so often red :)

PR #695, opened 2022-05-17
Adds E3 capability for clearing scrollback and screen

### LINKED ISSUE #693: `clear` command should also clear scrollback history
## Abstract

Currently when `clear` is ran in contour, contour only clears to the top to screen and keeps the scrollback buffer intact.
## Motivation
The current behavior is already same as CTRL-L. So `clear` by default should clear scrollback to accommodate the use cases when user wants to clear the scrollback.

Utkarsh-khambra:
fix #693

### DIFF
--- a/src/terminal/Capabilities.cpp
+++ b/src/terminal/Capabilities.cpp
@@ -163,6 +163,7 @@ namespace
         String { "ec"_tcap, "ech"sv, "\033[%p1%dX"sv },
         // Clear to end of screen
         String { "cd"_tcap, "ed"sv, "\033[J"sv },
+        String { "E3"_tcap, "E3"sv, "\033[3J"sv },                    // Defines the E3 capability
         String { "ce"_tcap, "el"sv, "\033[K"sv },                     // Clear to end of line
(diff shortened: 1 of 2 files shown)


src/terminal/Capabilities.cpp
```diff
@@ -163,6 +163,7 @@ namespace
         String { "ec"_tcap, "ech"sv, "\033[%p1%dX"sv },
         // Clear to end of screen
         String { "cd"_tcap, "ed"sv, "\033[J"sv },
+        String { "E3"_tcap, "E3"sv, "\033[3J"sv },                    // Defines the E3 capability
```
christianparpart, line 166: It took me a while to find out where this `E3` capability is coming from, but looks like from this quote:
```
# E3 clears the terminal's scrollback buffer.  This was implemented in the
#    Linux 3.0 kernel as a security feature.  It matches a feature which was
#    added in xterm patch #107.
```
`E3` capability was first introduced by the linux kernel (version 3.0). I guess they're talking about the builtin tty layer there.

christianparpart (approved):
It took me

christianparpart (approved):

PR #698, opened 2022-05-24
fix #696

### LINKED ISSUE #696: Weird Light Color On The Line After Cursor
### Contour Terminal version

0.3.2-unreleased-master-88e6456f

### Installer source

Yay

### Operating System

Manjaro Linux

### Architecture

x86-64

### Other Software

None

### Steps to reproduce

1. Install Yay or any AUR Package Manager
2. Install [`contour-git`](https://aur.archlinux.org/packages/contour-git) Package (`yay contour-git`)
3. Run The Terminal

### Expected Behavior

This is how the terminal should look like:
image

### Actual Behavior

This is how it actually looks:
image

### Additional notes

I am using the Bleeding Edge Version of Contour, using the package available on [Arch Linux AUR](https://aur.archlinux.org/packages/contour-git).

I know i am using the Bleeding Edge Version and it may contain bugs, i am raising this issue just to let the developers know.

The problem might be of the prompt i am using: [Powerline Shell](https://github.com/b-ryan/powerline-shell)
But it doesn't occur on other terminals like the xfce4-terminal

### Important
Seems like the problem is with the prompt i am using which is [Powerline shell](https://github.com/b-ryan/powerline-shell) because the default prompt with colors works fine, but powerline-shell prompt works fine on xfce terminal and others too

Utkarsh-khambra:
fix #696

### DIFF
--- a/src/terminal/Line.cpp
+++ b/src/terminal/Line.cpp
@@ -11,6 +11,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+#include <terminal/GraphicsAttributes.h>
 #include <terminal/Line.h>
 
 #include <unicode/grapheme_segmenter.h>
@@ -188,7 +189,7 @@ InflatedLineBuffer<Cell> inflate(TriviallyStyledLineBuffer const& input)
     }
 
     while (columns.size() < unbox<size_t>(input.displayWidth))
-        columns.emplace_back(Cell { input.attributes });
+        columns.emplace_back(Cell { GraphicsAttributes {} });
 
     return columns;
 }


src/terminal/Line.cpp
```diff
@@ -188,7 +189,7 @@ InflatedLineBuffer<Cell> inflate(TriviallyStyledLineBuffer const& input)
     }
 
     while (columns.size() < unbox<size_t>(input.displayWidth))
-        columns.emplace_back(Cell { input.attributes });
+        columns.emplace_back(Cell { GraphicsAttributes {} });
```
christianparpart, line 192: That's been on my list since some time already but never had time recently. Thanks. :+1: :)

christianparpart (approved):

PR #713, opened 2022-06-04
Fix build on Windows

rhcher:
In my computer, the CMAKE_SYSTEM_PROCESSOR is "AMD64" so it will build failed with message "Target architecture not detected", add this fix it.


christianparpart (approved):
Thanks :)

Issue #714, opened 2022-06-04
Support XTPUSHCOLORS / XTPOPCOLORS


christianparpart:
It would be really nice to find out if there are actually any apps using these, however.

### From ctlseqs.txt
```txt
CSI # P
CSI Pm # P
          Push current dynamic- and ANSI-palette colors onto stack
          (XTPUSHCOLORS), xterm.  Parameters (integers in the range 1
          through 10, since the default 0 will push) may be used to
          store the palette into the stack without pushing.

CSI # Q
CSI Pm # Q
          Pop stack to set dynamic- and ANSI-palette colors
          (XTPOPCOLORS), xterm.  Parameters (integers in the range 1
          through 10, since the default 0 will pop) may be used to
          restore the palette from the stack without popping.

CSI # R   Report the current entry on the palette stack, and the number
          of palettes stored on the stack, using the same form as
          XTPOPCOLOR (default = 0) (XTREPORTCOLORS), xterm.

```

PR #726, opened 2022-06-09
[terminal] Adds VT sequence `XTPUSHCOLORS`, `XTPOPCOLORS` (#714).

christianparpart:


src/terminal/Terminal.cpp
```diff
@@ -1729,4 +1729,40 @@ void Terminal::setHighlightRange(HighlightRange _range)
...
+    if (slot == 0)
+        slot = state_.savedColorPalettes.size() - 1;
+    else
+        slot--;
+
+    if (!state_.savedColorPalettes.empty())
+    {
+        state_.colorPalette = move(state_.savedColorPalettes.back());
+        state_.savedColorPalettes.pop_back();
+    }
```
Utkarsh-khambra, line 1765: Slots 1 through 10 should only restore the color palette and not pop the stack.

> Parameters (integers in the range 1
 >         through 10, since the default 0 will pop) may be used to
  >        restore the palette from the stack without popping.

christianparpart: done

PR #734, opened 2022-06-12
Fix oversized non-emoji rgba glyphs

christianparpart:


src/terminal_renderer/TextRenderer.cpp
```diff
@@ -748,16 +748,23 @@ auto TextRenderer::createRasterizedGlyph(atlas::TileLocation tileLocation,
...
         {
+            RasterizerLog()("Scaling oversized glyph of {}+{} down to bounding box {}.",
+                            glyph.bitmapSize,
+                            glyph.position,
+                            emojiBoundingBox);
             auto [scaledGlyph, scaleFactor] = text::scale(glyph, emojiBoundingBox);
             glyph = move(scaledGlyph);
             // glyph.position.y = unbox<int>(glyph.bitmapSize.height) - _gridMetrics.underline.position;
+            // TODO(pr) Adapt glyph.position's x and y accordingly.
+            glyph.position.x = 0; // int(float(glyph.position.x) * scaleFactor);
```
christianparpart, line 766: If it was oversized, probably just center it after down sampling.

whisperity:
In general **before** landing this patch there should be a defence mechanism against corruption due to oversized glyphs in the texture atlas system first.

christianparpart:
> In general **before** landing this patch there should be a defence mechanism against corruption due to oversized glyphs in the texture atlas system first.

I've done that now as an "assert", because it's guarded earlier, and if something is leaking through, then it's a bug that should be reported and fixed rather than silently ignored. 

Oversized glyphs however should not be a problem now anyways (famous last words). :)

PR #735, opened 2022-06-13
adds vt sequence `DECPS`

### LINKED ISSUE #237: DECPS, play sound in terminal
https://github.com/microsoft/terminal/issues/8687 WT will likely have that soon too and i think that is a pretty fun VT sequence. So let's add it.

Utkarsh-khambra:
closes #237 

- [ ] Write tests
- [x] Support Qt6

### DIFF
--- a/.github/fedora/contour.spec
+++ b/.github/fedora/contour.spec
@@ -27,6 +27,7 @@ BuildRequires:  ninja-build
 BuildRequires:  pkgconf
 BuildRequires:  qt5-qtbase-devel
 BuildRequires:  qt5-qtbase-gui
+BuildRequires:  qt5-qtmultimedia-devel
 
 Requires:       fontconfig
 Requires:       freetype
@@ -35,6 +36,7 @@ Requires:       kf5-kwindowsystem
 Requires:       qt5-qtbase
 Requires:       qt5-qtbase-gui
 Requires:       yaml-cpp
+Requires:       qt5-qtmultimedia-devel
 
 %description
 contour is a modern terminal emulator, for everyday use.
--- a/Changelog.md
+++ b/Changelog.md
@@ -38,6 +38,7 @@
 - Adds configuration options `logging.enabled` as well as `logging.file`.
 - Adds VT sequences `XTPUSHCOLORS`, `XTPOPCOLORS`, `XTREPORTCOLORS` (#714).
 - Adds CLI syntax `contour info vt` to print list of supported VT sequences (#730).
+- Adds VT sequence `DECPS` (#237).
 
 ### 0.3.1 (2022-05-01)
 
--- a/src/contour/Audio.cpp
+++ b/src/contour/Audio.cpp
@@ -1,0 +1,131 @@
+#include <contour/Audio.h>
+
+#include <crispy/assert.h>
+#include <crispy/logstore.h>
+
+#include <qbuffer.h>
+#include <qthread.h>
+
+#if QT_VERSION >= 0x060000
+    #include <QtMultimedia/QMediaDevices>
+#endif
+
+using namespace contour;
+
+namespace
+{
+constexpr double SAMPLE_RATE = 44100;
+
+// TODO make this function constexpr when we switch to c++23
+double square_wave(double x) noexcept
+{
+    x = std::fmod(x, 2);
+    return std::isgreater(x, 1) ? -1 : 1;
+}
+
+auto createMusicalNote(double volume, double duration, double frequency) noexcept
+{
+    duration = duration / 32.0;
+    volume /= 7;
+    std::vector<std::int16_t> buffer;
+    buffer.reserve(static_cast<size_t>(std::ceil(duration * SAMPLE_RATE)));
+    for (double i = 0; i < duration * SAMPLE_RATE; ++i)
+        buffer.push_back(
+            static_cast<int16_t>(0x7fff * volume * square_wave(frequency / SAMPLE_RATE * i * 2)));
+    return buffer;
+}
+
+} // namespace
+
+Audio::Audio()
+{
+    QAudioFormat f;
+    f.setSampleRate(44100);
+    f.setChannelCount(1);
+
+#if QT_VERSION >= 0x060000
+    f.setSampleFormat(QAudioFormat::Int16);
+    QAudioDevice info(QMediaDevices::defaultAudioOutput());
+#else
+    f.setSampleSize(16);
+    f.setCodec("audio/pcm");
+    f.setByteOrder(QAudioFormat::LittleEndian);
+    f.setSampleType(QAudioFormat::SignedInt);
+    QAudioDeviceInfo info(QAudioDeviceInfo::defaultOutputDevice());
+#endif
+
+    if (!info.isFormatSupported(f))
+    {
+        errorlog()("Default output device doesn't support 16 Bit signed integer PCM");
+        return;
+    }
+
+#if QT_VERSION < 0x060000
+    using QAudioSink = QAudioOutput;
+#endif
+
+    audio = std::make_unique<QAudioSink>(f);
+
+    audio->moveToThread(&soundThread_);
+
+    connect(audio.get(), &QAudioSink::stateChanged, this, &Audio::handleStateChanged);
+    qRegisterMetaType<std::vector<int>>();
+    connect(this, &Audio::play, this, &Audio::handlePlayback);
+    soundThread_.start();
+}
+
+Audio::~Audio()
+{
+    soundThread_.quit();
+    soundThread_.wait();
+}
+
+void Audio::fillBuffer(int volume, int duration, crispy::span<int const> notes)
+{
+    for (auto const i: notes)
+    {
+        auto b = createMusicalNote(volume, duration, i);
+        byteArray_.append(reinterpret_cast<char const*>(b.data()), static_cast<int>(2 * b.size()));
+    }
+}
+
+void Audio::handlePlayback(int volume, int duration, std::vector<int> const& notes)
+{
+    Require(audio);
+    if (audio->state() == QAudio::State::ActiveState)
+    {
+        fillBuffer(volume, duration, crispy::span(notes.data(), notes.size()));
+        return;
+    }
+    fillBuffer(volume, duration, crispy::span(notes.data(), notes.size()));
+    audioBuffer_.setBuffer(&byteArray_);
+    audioBuffer_.open(QIODevice::ReadWrite);
+    audio->start(&audioBuffer_);
+}
+
+void Audio::handleStateChanged(QAudio::State state)
+{
+    switch (state)
+    {
+        case QAudio::IdleState:
+            audio->stop();
+            audioBuffer_.close();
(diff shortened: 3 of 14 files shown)


src/contour/TerminalSession.cpp
```diff
@@ -616,6 +616,13 @@ void TerminalSession::onHighlightUpdate()
 {
     terminal_.resetHighlight();
 }
+
+void TerminalSession::playSound(terminal::Sequence::Parameters const params_)
+{
+    auto range = params_.range();
+    emit audio.play(params_.at(0), params_.at(1), { range.begin() + 2, range.end() });
```
christianparpart, line 623: You are constructing a new `std::vector<int>` for the notes here. Is it possible to pass a `span<int>` instead?
If ownership is a problem, we could probably use some reusable pre-allocated vector?
I am just thinking out loud here.

src/crispy/Note.cpp
```diff
@@ -0,0 +1,68 @@
+#include <crispy/Note.h>
+#include <crispy/assert.h>
+
+#include <gsl/gsl_assert>
```
christianparpart, line 4: Can you use `<crispy/assert.h>` instead? It provides alternative implementations with a little bit more flexibility. :)

src/crispy/Note.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+} // namespace
+std::vector<std::int16_t> crispy::createNote(double volume, int duration, Note note_) noexcept
+{
+    Ensures(static_cast<int>(note_) >= 0 && static_cast<int>(note_) < 26);
+    switch (note_)
+    {
+        case Note::A5: return ::createNote(volume, duration, 880.0); break;
+        case Note::A6: return ::createNote(volume, duration, 1760.0); break;
+        case Note::A_SHARP_5: return ::createNote(volume, duration, 932.3275); break;
```
christianparpart, line 42: This switch can be greatly simplified as the formula is is something like `2 ^ something` (I forgot), but it's a power of two with the base node concert A (A4) being 442 Hz, nowadays. I'm just dropping of my brain here, it's okay to have it like that for now and do that kind of algorithmic simplification later. :)

src/terminal/Functions.h
```diff
@@ -352,6 +352,7 @@ constexpr inline auto CAPTURE     = detail::CSI('>', 0, 2, std::nullopt, 't', VT
 
 constexpr inline auto DECSSDT     = detail::CSI(std::nullopt, 0, 1, '$', '~', VTType::VT320, "DECSSDT", "Select Status Display (Line) Type");
 constexpr inline auto DECSASD     = detail::CSI(std::nullopt, 0, 1, '$', '}', VTType::VT420, "DECSASD", "Select Active Status Display");
+constexpr inline auto DECPS 			= detail::CSI(std::nullopt, 3, 18, ',', '~', VTType::VT520, "DECPS", "Controls the sound frequency or notes");
```
christianparpart, line 355: is `18` a documented limit? Otherwise we could use `ArgsMax`.

src/crispy/Note.h
```diff
@@ -0,0 +1,39 @@
+#pragma once
+#include <array>
+#include <cstdint>
+#include <vector>
+namespace crispy
+{
+enum class Note
```
christianparpart, line 7: The class and the filename could probably be better named `MusicalNote`. I personally find `Note` a little bit too synonym-rich. :)

src/contour/TerminalSession.cpp
Utkarsh-khambra: `span<int>` has a problem that `params_` gets deallocated before slot gets called, so slot gets garbled values.
A preallocated vector is a nice idea I'll change it to use that.

src/crispy/Note.h
Utkarsh-khambra: Done

Changelog.md
```diff
@@ -35,6 +35,7 @@
 - Adds new action `ToggleInputProtection` to protect terminal application against accidental input (#697).
 - Adds configuration options `logging.enabled` as well as `logging.file`.
 - Adds VT sequences `XTPUSHCOLORS`, `XTPOPCOLORS`, `XTREPORTCOLORS` (#714).
+- Adds VT sequence `DECPS` (#237).
```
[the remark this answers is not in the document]
j4james: Note that this sequence isn't really related to MIDI. I initially used MIDI in the Windows Terminal implementation, but it's quite likely we might have to replace that at some point.

src/terminal/Functions.h
j4james: FYI, jerch did some testing of DECPS on a real VT525, and it wasn't actually able to handle more than one note at a time. It's possible the multi-note capabilities were introduced in later firmware revisions though.

In any event, we decided it best to keep the multi-note functionality anyway, since it was at least documented in the manual. But I haven't limited it to 16 in Windows Terminal - just the existing parameter limit, which for us is 30 notes (32 - 2). I think MinTTY might be the same. Not sure about XTerm.js.

Nothing wrong with limiting it to 16 though.

christianparpart: I think then sticking with the value `18` (`16 + 2`) in there is okay, as long as the limit isn't less than what the spec is saying.

christianparpart:
```suggestion
constexpr inline auto DECPS       = detail::CSI(std::nullopt, 3, 18, ',', '~', VTType::VT520, "DECPS", "Controls the sound frequency or notes");
```
(use spaces instead of tabs, btw, you should be able to configure your editor to evaluate the project root's .editorconfig file, that contains this information :) )

christianparpart:
@Utkarsh-khambra with regards to the build failures when building using Qt6 (the one CI test), the [`QAudioOutput`](https://doc.qt.io/qt-6/qaudiooutput.html) API has changed slightly.

I hope it's possible to `#ifdef` that out to distinguish between `< QT_VERSION(6, 0, 0)` and `>=`.

Do we actually have *any* possibility to run Qt6 on Linux already (with what distro)? Hmm...

Changelog.md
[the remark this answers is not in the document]
Utkarsh-khambra: Since Qt doesn't support MIDI out of the box, but supports playing `PCM`. I am just generating `PCM` of a square wave for the given note.

src/contour/Audio.cpp
```diff
@@ -0,0 +1,115 @@
...
+            }
+            break;
+        default: break;
+    }
+}
+
+std::vector<std::int16_t> Audio::createMusicalNote(double volume, int duration, int note_) noexcept
+{
+    Require(static_cast<int>(note_) >= 0 && static_cast<int>(note_) < 26);
+    double frequency = (note_ != 0) * 440.0 * std::pow(2, (note_ + 2) / 12.0);
```
christianparpart, line 113: Very nice. Looks much much shorter than the initial version.

EDIT: deleted all my text. 

I did a little test in [compiler explorer](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:1,endLineNumber:4,positionColumn:1,positionLineNumber:4,selectionStartColumn:1,selectionStartLineNumber:4,startColumn:1,startLineNumber:4),source:'%23include+%3Ccmath%3E%0A%23include+%3Cclimits%3E%0A%0Aint+branching_cpp(double+note)+%7B%0A++++if+(note+%3C+std::numeric_limits%3Cdouble%3E::epsilon())%0A++++++++return+0%3B%0A++++else%0A++++++++return+440.0+*+std::pow(2,+(note+%2B+2)+/+12.0)%3B%0A%7D%0A%0Aint+branchless_cpp(double+note)+%7B%0A++++return+(note+!!%3D+0)+*+440.0+*+std::pow(2,+(note+%2B+2)+/+12.0)%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:g121,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'-O3+-msse2',selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1,tree:'1'),l:'5',n:'0',o:'x86-64+gcc+12.1+(C%2B%2B,+Editor+%231,+Compiler+%231)',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4).

Turns out that using an explicit branch is (from simply looking at it) shorter and uses only 2 basic blocks whereas the branchless version uses three. Not sure what's faster. I'm fine with both versions. :-)

src/contour/Audio.cpp
```diff
@@ -0,0 +1,115 @@
...
+{
+    duration = duration / 32.0;
+    volume /= 7;
+    std::vector<std::int16_t> buffer;
+    buffer.reserve(static_cast<size_t>(std::ceil(duration * SAMPLE_RATE)));
+    for (double i = 0; i < duration * SAMPLE_RATE; ++i)
+    {
+        auto sample = static_cast<int16_t>(0x7fff * volume * square_wave(frequency / SAMPLE_RATE * i * 2));
+        buffer.push_back(sample);
+    }
```
christianparpart, line 32:
```suggestion
        buffer.push_back(static_cast<int16_t>(0x7fff * volume * square_wave(frequency / SAMPLE_RATE * i * 2)););
```

just a suggestion. :)

christianparpart:
> Write tests

I think here the test should be sufficient if it's abstractr, not testing the *actual* audio but only if it's forwarded  correctly to the event listener.

.github/fedora/contour.spec
```diff
@@ -27,6 +27,7 @@ BuildRequires:  ninja-build
 BuildRequires:  pkgconf
 BuildRequires:  qt5-qtbase-devel
 BuildRequires:  qt5-qtbase-gui
+BuildRequires:  qt5-qtmultimedia-devel
```
christianparpart, line 30:
```suggestion
BuildRequires:  qt5-qtmultimedia-devel

Requires: qt5-qtmultimedia
```

Don't we also need to add this to the runtime dependencies? (probably alphabetically sorted in the Requires:-list below).

src/contour/TerminalSession.cpp
```diff
@@ -616,6 +616,15 @@ void TerminalSession::onHighlightUpdate()
 {
     terminal_.resetHighlight();
 }
+
+void TerminalSession::playSound(terminal::Sequence::Parameters const params_)
```
christianparpart, line 620:
```suggestion
void TerminalSession::playSound(terminal::Sequence::Parameters const& params_)
```
Looks like it's okay to pass the parameter by cref.

Issue #738, opened 2022-06-16
Broken SGR due to buggy trivial grid line inflate() call?


christianparpart:
When fixed again, this time definitely with a test case that proofs it's not happening again. :-)

### Current buggy behavior:

This is nano with the menu bar at the bottom inverted (so it seems).
![unknown](https://user-images.githubusercontent.com/56763/174082931-43f83d25-ea60-4d60-a67d-4508aa308611.png)

### Expected behavior:

![image](https://user-images.githubusercontent.com/56763/174083095-0663ce1c-1121-4d29-9937-b3b3d3845b2a.png)

PR #744, opened 2022-06-20
Trivial line related refactors and an alternative less intrusive fix proposal.

christianparpart:
Based on the work at #742.
Should be closing #738.

@Utkarsh-khambra - please look at the diff per commit individually.

The first commit does some refactorings, reducing some code duplications by introducing `RenderAttributes`. This looks like a little bit of overkill at first, but makes sense in the third commit.

The second commit simply renames the trival-line buffer struct to `TrivialLineBuffer` to finally fix my brain damage and hopefully have a nice and yet still self-explaining type name. :)

The third commit **actually fixes** the line inflation bug by differenciating between SGR at the time of line creation/erase/reset versus current SGR at the time of actually writing the text.

@Utkarsh-khambra I by no means intended to steal that from you, but I hoped to give you a less intrusive solution with that.
What do you think?

### DIFF
--- a/src/terminal/RenderBuffer.h
+++ b/src/terminal/RenderBuffer.h
@@ -25,8 +25,18 @@
 #include <optional>
 #include <vector>
 
+#include <terminal_renderer/RenderTarget.h>
+
 namespace terminal
 {
+
+struct RenderAttributes
+{
+    RGBColor foregroundColor {};
+    RGBColor backgroundColor {};
+    RGBColor decorationColor {};
+    CellFlags flags {};
+};
 
 /**
  * Renderable representation of a grid cell with color-altering pre-applied and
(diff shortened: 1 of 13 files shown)


src/terminal/RenderBuffer.h
```diff
@@ -25,9 +25,19 @@
 #include <optional>
 #include <vector>
 
+#include "terminal_renderer/RenderTarget.h"
```
Utkarsh-khambra, line 28:
```suggestion
#include <terminal_renderer/RenderTarget.h>
```

Utkarsh-khambra (approved):

Utkarsh-khambra:
I like your fix, mine was too intrusive.

Utkarsh-khambra:
Also can you remove `fillRemainingCells` function, I added it previously, but with this fix we  do not need it now :). 
It should be in three places, Line.h, Line.cpp and Terminal.cpp

PR #751, opened 2022-06-22
Selection snaps to end of line, when it reaches last filled cell

### LINKED ISSUE #665: selecting past the last filled character in a line should snap to the end of that line
Selecting past the last filled character in a line should snap to the end of that line

Emerging out of #641 to explicitly address this one item. /cc @herrhotzenplotz 

I initially planned to address this one for 0.3.1, but I'd prefer to release 0.3.1 as early as possible, so it might be 0.3.2 instead, FYI.

Utkarsh-khambra:
Closes #665

### DIFF
--- a/src/terminal/Terminal.cpp
+++ b/src/terminal/Terminal.cpp
@@ -582,7 +582,7 @@ bool Terminal::sendMouseMoveEvent(Modifier _modifier,
 
     currentMousePosition_ = newPosition;
 
-    auto const relativePos = viewport_.translateScreenToGridCoordinate(currentMousePosition_);
+    auto relativePos = viewport_.translateScreenToGridCoordinate(currentMousePosition_);
 
     changed = changed || updateCursorHoveringState();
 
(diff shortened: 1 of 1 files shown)


src/terminal/Line.h
```diff
@@ -235,6 +235,19 @@ class Line
         return inflatedBuffer().at(unbox<size_t>(column)).width();
     }
 
+    [[nodiscard]] bool compareCellTextAt(ColumnOffset column, char codepoint) const noexcept
```
christianparpart, line 238: Please already assume that `TrivialLineBuffer` already contains UTF-8.
Currently it's only containing US-ASCII, but with a future PR, that will be backwards-compatibly UTF-8.

Assuming that this function is **only** being invoked when snapping interactive selections, I would say it is okay to inflate the underlying buffer in order to easily access the grapheme cluster at a given column. That makes it more expensive for the first call, but faster for further calls, as users usually don't stop the mouse-move as soon as it snaps. - The general case of improved performance for trivial lines is also still retained :)

With that, I think this input parameter should at least be `char32_t` (or `u32string_view`).

src/terminal/Screen.h
```diff
@@ -485,11 +485,7 @@ class Screen final: public ScreenBase, public capabilities::StaticDatabase
 
     [[nodiscard]] bool compareCellTextAt(CellLocation position, char codepoint) const noexcept override
     {
-        return grid()
-            .lineAt(position.line)
-            .inflatedBuffer()
-            .at(position.column.as<size_t>())
-            .compareText(codepoint);
+        return grid().lineAt(position.line).compareCellTextAt(position.column, codepoint);
```
christianparpart, line 488: I think - with the above comment - the old code part here could still be used instead (no need for a special function unless an optimized implementation could be easily found). What do you think?

Utkarsh-khambra: I changed it to be like previous. Maybe we can think about it later :)

PR #754, opened 2022-06-27
Normalizes `\r\n` and `\r` from clipboard to `\n` for non windows platforms

Utkarsh-khambra:
GTK3 apps seems to add CRLF line endings on copying text into clipboard. I just replaced those CRLF to LF before sending it pty.

### DIFF
--- a/src/contour/TerminalSession.cpp
+++ b/src/contour/TerminalSession.cpp
@@ -85,6 +85,15 @@ namespace
         pthread_setname_np(pthread_self(), name);
 #endif
     }
+
+    string normalize_crlf(QString&& text)
+    {
+#if !defined(_WIN32)
+        return text.replace("\r\n", "\n").replace("\r", "\n").toUtf8().toStdString();
+#else
+        return text.toUtf8().toStdString();
+#endif
+    }
 } // namespace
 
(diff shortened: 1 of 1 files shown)


src/contour/TerminalSession.cpp
```diff
@@ -375,7 +374,15 @@ void TerminalSession::pasteFromClipboard(unsigned count)
         SessionLog()("pasteFromClipboard: mime data contains {} formats.", md->formats().size());
         for (int i = 0; i < md->formats().size(); ++i)
             SessionLog()("pasteFromClipboard[{}]: {}\n", i, md->formats().at(i).toStdString());
+#if !defined(_WIN32)
+        string const text = clipboard->text(QClipboard::Clipboard)
+                                .replace("\r\n", "\n")
+                                .replace("\r", "\n")
+                                .toUtf8()
+                                .toStdString();
+#else
```
christianparpart, line 383: This logic is generally the same as below. Maybe we can put it into a private (anonymous namespace) helper function on the top part of this file, to avoid code duplication?

christianparpart (commented):
sorry for being late (reasons as mentioned on Discord :) )

christianparpart (approved):
Given the conversation, LGTM :)

PR #759, opened 2022-07-12
Keep working directory same when new terminal is spawned

Utkarsh-khambra:
- [ ]  ~cwd in Windows~

### DIFF
--- a/src/contour/ContourGuiApp.cpp
+++ b/src/contour/ContourGuiApp.cpp
@@ -366,12 +366,12 @@ int ContourGuiApp::terminalGuiAction()
     return rv;
 }
 
-TerminalWindow* ContourGuiApp::newWindow(contour::config::Config const& _config)
+TerminalWindow* ContourGuiApp::newWindow(contour::config::Config const& config)
 {
     auto const* profile = _config.profile(profileName());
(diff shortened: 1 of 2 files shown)


Utkarsh-khambra:
[Here](https://stackoverflow.com/a/2967713) is a way to get cwd in windows. But don't have a windows pc to build it currently. So I'll leave it for future.

src/contour/ContourGuiApp.cpp
```diff
@@ -371,7 +371,7 @@ TerminalWindow* ContourGuiApp::newWindow(contour::config::Config const& _config)
     auto const* profile = _config.profile(profileName());
     if (!profile)
         return nullptr;
-
+    ContourGuiApp::_config = _config;
```
christianparpart, line 374: hm, okay, that's my bad for sure. We have multiple coding styles, sadly, and it's all my fault. I am sorry for that. It's probably best to simply rename the function parameter's `_config` to `config` and then have this line like `_config = config`.

christianparpart (changes requested):
I am so sorry for being late. The hotness must have taken my consciousness.

christianparpart (approved):

Issue #766, opened 2022-07-23
Add appdata.xml file & improve contour.desktop file


christianparpart:
While being in process of adding flatpak support I realized I am missing a appdata.xml file and the file name for the .desktop file isn't the best, probably.

- [ ] Rename `contour.desktop` to `org.contour.Terminal.desktop` (assuming I'll keep the app ID as proposed: `org.contour.Terminal`).
- [ ] add appdata.xml file with the same basename as above and have it installed on Linux installs into /usr/share/metainfo.

PR #767, opened 2022-07-24
Applying some tweaks with regards to Flatpak and AppStream (XML) recommendations.

christianparpart:
Contains an improved .desktop file, better file naming for that as well as for the application icon and provides an AppStream XML file.

This is the result coming out of https://github.com/flathub/flathub/pull/3349. Wth that merged and a potential next release on flatpak, the respective patch file from the flatpak repo must be removed to not break future releases.

Refs #766

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,8 @@
 ### 0.3.3 (unreleased)
 
 - Adds vim-like `scrolloff` feature to normal mode cursor movements to ensure a line padding when scrolling up/down.
+- [Linux] Changes the .desktop file name and icon file name to conform to the flatpak recommendations.
+- [Linux] Provide an AppStream XML file.
 - Internal: Y-axis inverted to match GUI coordinate systems where (0, 0) is top left rather than bottom left.
 
 ### 0.3.2 (2022-07-07)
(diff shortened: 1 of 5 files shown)


AsciiWolf:
There is something wrong with the second paragraph of AppStream metainfo description:

screenshot

christianparpart:
> There is something wrong with the second paragraph of AppStream metainfo description:
> 
> screenshot

I realized that today too. It very much looks like a website bug that we however can probably work around. Gonna try ASAP.

AsciiWolf:
It is not a website bug - I see the same problem when opening Contour in GNOME Software.

christianparpart:
@AsciiWolf oh. thanks for that information. I looked at it again, however, it is indeed valid XML (and HTML), however, and this is why i said it's likely a website render bug, it's simply a `<p>` tag that contains a `<li>` tag with the enumeration of features. I've looked at other projects (Discord is a great example here) and noticed that they don't embed the `li` inside the `p` but simply use `p` for each sentence, and `li` (w/o any `p`) in between. 

Trying that one out now: https://github.com/flathub/org.contourterminal.Contour/pull/1

christianparpart:
> There is something wrong with the second paragraph of AppStream metainfo description:

fixed. I think there's still room to make the text more interesting in the future though. I'm all open for suggestions at any time :)

PR #770, opened 2022-08-04
Fix autogen.sh

Marenz:


autogen.sh
```diff
@@ -18,9 +18,7 @@ else
     ROOTDIR="$(dirname $0)"
 fi
 
-if [[ "${CXX}" == "" ]]; then
-    CXX="g++"
-fi
+CXX=${CXX="g++"}
```
christianparpart, line 21: `[[` does not work in `sh` because that's actually bash syntax. I don't mind using this oneliner, because this `autogen.sh` script is meant for dev-only anyways - not for users, not for package maintainers (everybody can use it though).

FreeBSD's /bin/sh is much more strict, so I wonder why this wasn't reported yet. So probably go with this line instead?

```suggestion
if [ "x$CXX" = "x" ]; then
    CXX="g++"
fi
```

christianparpart:
just giving it a try. thx :)

Issue #52, opened 2019-09-17
aeroterm: Blurred background support


christianparpart:
Support blurred background on all platforms.

* [x] Windows (Acrylic backdrop)
* [ ] Linux (X11/Wayland)
* [ ] OS/X

References:
* https://gist.github.com/ethanhs/0e157e4003812e99bf5bc7cb6f73459f
* https://github.com/riverar/sample-win32-acrylicblur/blob/master/MainWindow.xaml.cs
* https://stackoverflow.com/questions/44000217/mimicking-acrylic-in-a-win32-app

Summertime:
* [ ] Linux: Gnome's Blur my Shell!

https://github.com/aunetx/blur-my-shell#application-blurring

christianparpart:
> * [ ] Linux: Gnome's Blur my Shell!
> 
> https://github.com/aunetx/blur-my-shell#application-blurring

Nice. Thanks!

christianparpart:
@Summertime I've merged it to master, but we need to wait for https://github.com/aunetx/blur-my-shell/issues/300 to be addressed. That may not be instant, and you may need to just add contour's WM_CLASS to explicitly enable it until the referenced ticket is fixed, but it should hopefully only be a matter of time. And thx for the info on blur-my-shell ;-)

PR #789, opened 2022-08-11
Fixing blur-behind for XCB platforms.

christianparpart:
Refs #52.

The PR #771 dropped kwin dependency but didn't do it right. `QWindow::setProperty(...)` is indeed NOT setting X11 like properties but that function is related to Qt properties.

This most likely might be fixing the issue at https://github.com/aunetx/blur-my-shell/issues/300. Sorry for the noise. :)


data-man:
Thanks! :fire:

PR #803, opened 2022-09-05
Add suse to install-deps

Marenz:


scripts/install-deps.sh
```diff
@@ -128,6 +128,33 @@ prepare_fetch_and_unpack()
...
+        libqt5-qtx11extras-devel
+        libqt5gui5
+        libqt5opengl5-dev
+        libyaml-cpp-dev
+        make
+        ncurses-bin
+        pkg-config
+        qtbase5-dev
+        qtmultimedia5-dev
+    "
```
christianparpart, line 156: isn't there something missing? like a `yum install $packages` or so? :-)

PR #825, opened 2022-09-20
Now moves cursor vertically for sixel newlines even if there is no sixel

### LINKED ISSUE #822: no vertical cursor progression on GLFs (sixel)
### Contour Terminal version

0.3.4.223

### Installer source

GitHub: release page

### Operating System

Ubuntu 18 LTS

### Architecture

x86-64

### Other Software

_No response_

### Steps to reproduce

Plz see https://github.com/jerch/xterm-addon-image/issues/37 for details on the issue. There is also this [testscript](https://github.com/hackerb9/vt340test/blob/main/jerch/gcrglf.sh) done by @hackerb9 to check TE behavior.

### Expected Behavior

GLFs from sixel should advance the text cursor to the bottom, if sixel scrolling is on.

### Actual Behavior

Currently there is no text cursor progression from multiple GLFs.

### Additional notes

(@Utkarsh-khambra  This is the follow-up from #820.)

Utkarsh-khambra:
fixes #822

To test this use [this script](https://github.com/hackerb9/vt340test/blob/main/jerch/gcrglf.sh). Since the script requires cell height to be 20px, you can change font size to be 11 in default config with monospace font family.
Screenshot from 2022-09-20 22-53-58

### DIFF
--- a/src/terminal/Screen.cpp
+++ b/src/terminal/Screen.cpp
@@ -1621,9 +1621,8 @@ void Screen<Cell>::renderImage(shared_ptr<Image const> _image,
                                ImageResize _resizePolicy,
                                bool _autoScroll)
 {
-    // TODO: make use of _imageOffset and _imageSize
+    // TODO: make use of _imageOffset
     (void) _imageOffset;
-    (void) _imageSize;
 
     auto const linesAvailable = _state.pageSize.lines - _topLeft.line.as<LineCount>();
     auto const linesToBeRendered = min(_gridSize.lines, linesAvailable);
@@ -1634,7 +1633,21 @@ void Screen<Cell>::renderImage(shared_ptr<Image const> _image,
     // TODO: make use of _imageOffset and _imageSize
     auto const rasterizedImage = _state.imagePool.rasterize(
         move(_image), _alignmentPolicy, _resizePolicy, gapColor, _gridSize, _state.cellPixelSize);
-
+    const auto lastSixelBand = _imageSize.height.value % 6;
+    const LineOffset offset = [&]() {
+        auto offset =
+            LineOffset::cast_from(std::ceil(static_cast<float>(_imageSize.height.value - lastSixelBand)
+                                            / float(*_state.cellPixelSize.height)))
+            - 1 * (lastSixelBand == 0);
+        auto const h = _imageSize.height.value - 1;
+        // VT340 has this behavior where for some heights it text cursor is placed not
+        // at the final sixel line but a line above it.
+        // See
+        // https://github.com/hackerb9/vt340test/blob/main/glitches.md#text-cursor-is-left-one-row-too-high-for-certain-sixel-heights
+        if (h % 6 > h % _state.cellPixelSize.height.value)
+            return offset - 1;
+        return offset;
+    }();
     if (*linesToBeRendered)
     {
         for (GridSize::Offset const offset: GridSize { linesToBeRendered, columnsToBeRendered })
@@ -1643,7 +1656,7 @@ void Screen<Cell>::renderImage(shared_ptr<Image const> _image,
             cell.setImageFragment(rasterizedImage, CellLocation { offset.line, offset.column });
             cell.setHyperlink(_state.cursor.hyperlink);
         };
-        moveCursorTo(_topLeft.line + unbox<int>(linesToBeRendered) - 1, _topLeft.column);
+        moveCursorTo(_topLeft.line + offset, _topLeft.column);
     }
 
     // If there're lines to be rendered missing (because it didn't fit onto the screen just yet)
@@ -1667,7 +1680,7 @@ void Screen<Cell>::renderImage(shared_ptr<Image const> _image,
         }
     }
     // move ansi text cursor to position of the sixel cursor
-    moveCursorToColumn(_topLeft.column + _gridSize.columns.as<ColumnOffset>());
+    moveCursorToColumn(_topLeft.column);
 }
 
 template <typename Cell>
--- a/src/terminal/SixelParser.cpp
+++ b/src/terminal/SixelParser.cpp
@@ -412,13 +412,19 @@ void SixelImageBuilder::write(CellLocation const& _coord, RGBColor const& _value
                 size_.width = Width::cast_from(_coord.column + 1);
         }
 
-        auto const base =
-            unbox<unsigned>(_coord.line) * unbox<unsigned>((explicitSize_ ? size_.width : maxSize_.width)) * 4
-            + unbox<unsigned>(_coord.column) * 4;
-        buffer_[base + 0] = _value.red;
-        buffer_[base + 1] = _value.green;
-        buffer_[base + 2] = _value.blue;
-        buffer_[base + 3] = 0xFF;
+        for (auto i = 0; i < aspectRatio_.nominator; ++i)
+        {
+            auto const base = unbox<unsigned>(_coord.line + i)
+                                  * unbox<unsigned>((explicitSize_ ? size_.width : maxSize_.width)) * 4
+                              + unbox<unsigned>(_coord.column) * 4;
+            for (auto j = 0; j < aspectRatio_.denominator; ++j)
+            {
+                buffer_[base + 0 + 4 * static_cast<unsigned int>(j)] = _value.red;
+                buffer_[base + 1 + 4 * static_cast<unsigned int>(j)] = _value.green;
+                buffer_[base + 2 + 4 * static_cast<unsigned int>(j)] = _value.blue;
+                buffer_[base + 3 + 4 * static_cast<unsigned int>(j)] = 0xFF;
+            }
+        }
     }
 }
 
@@ -451,7 +457,7 @@ void SixelImageBuilder::setRaster(int _pan, int _pad, ImageSize _imageSize)
     size_.width = clamp(_imageSize.width, Width(0), maxSize_.width);
     size_.height = clamp(_imageSize.height, Height(0), maxSize_.height);
 
-    buffer_.resize(size_.area() * 4);
+    buffer_.resize(size_.area() * static_cast<size_t>(_pad * _pan) * 4);
     explicitSize_ = true;
 }
 
--- a/src/terminal/SixelParser.cpp
+++ b/src/terminal/SixelParser.cpp
@@ -440,8 +440,7 @@ void SixelImageBuilder::rewind()
 void SixelImageBuilder::newline()
 {
     sixelCursor_.column = {};
-
-    if (unbox<int>(sixelCursor_.line) + 6 < unbox<int>(maxSize_.height))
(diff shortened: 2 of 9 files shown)


christianparpart:
Many thanks!

@Utkarsh-khambra with that, I don't think we do not need the configuration option `images.sixel_cursor_conformance` (bool) anymore as I was clearly misunderstanding/misreading/missing some of the Sixel documentation back in the days. I think it's save to completely get rid of that config option. ;)

src/terminal/SixelParser.cpp
```diff
@@ -477,6 +476,12 @@ void SixelImageBuilder::render(int8_t _sixel)
 
 void SixelImageBuilder::finalize()
 {
+    if (unbox<int>(size_.height) == 1)
+    {
+        size_.height = Height::cast_from(sixelCursor_.line.value);
```
christianparpart, line 481:
```suggestion
        size_.height = Height::cast_from(sixelCursor_.line);
```

src/terminal/SixelParser_test.cpp
```diff
@@ -338,3 +338,17 @@ TEST_CASE("SixelParser.newline", "[sixel]")
         }
     }
 }
+
+TEST_CASE("SixelParser.vertical_cursor_advance", "[sixel]")
+{
+    auto constexpr defaultColor = RGBAColor { 0, 0, 0, 255 };
+    SixelImageBuilder ib(
+        { Width(5), Height { 30 } }, 1, 1, defaultColor, std::make_shared<SixelColorPalette>(16, 256));
```
christianparpart, line 346: Maybe a copy'n'paste thing, but there's a style breakage. Width is using `()` whereas Height is using `{}`. both solution work, but maybe only one should be used for consistency.

src/terminal/Screen.cpp
```diff
@@ -1643,7 +1643,7 @@ void Screen<Cell>::renderImage(shared_ptr<Image const> _image,
             cell.setImageFragment(rasterizedImage, CellLocation { offset.line, offset.column });
             cell.setHyperlink(_state.cursor.hyperlink);
         };
-        moveCursorTo(_topLeft.line + unbox<int>(linesToBeRendered) - 1, _topLeft.column);
+        moveCursorTo(_topLeft.line + unbox<int>(linesToBeRendered), _topLeft.column);
```
Utkarsh-khambra, line 1646: from [this test](https://github.com/hackerb9/vt340test/blob/main/j4james/cursor_position.sh#L3-L8) 
> Test of the text cursor position after a Sixel image is output.
 On a genuine VT340, after an image is output, the text cursor position is
 set to the same row as the top of the final sixel line (regardless of
 whether anything is output on that line). The column position remains
 unchanged, i.e. it should align with the left of the image.

@christianparpart  this change should place the ansi cursor to last sixel row, right?

christianparpart: that is how i read it. any objections, @jerch?

Utkarsh-khambra: If that's the case I think this is ready to merge.

jerch: I think "top of final sixel line" here means to clip the text cursor into the text row, wherever the first of the 6 pixel line of the last sixel band ended up above. So its not the last line (sixth pixel line), but the first one determining it. Note that this has a weird consequence, if the last sixel band is not perfectly mapped on text line borders, possibly cutting off lower pixel rows, or not trigger scrolling at the end of the screen.
My impl still usees the last pixel line for this, but I prolly will change that.

Utkarsh-khambra: So that means you may overwrite pixels in the last sixel line, if you print text just after drawing a sixel image.

jerch: Yes thats how I understand it so far. But lets see if it gets confirmed...

hackerb9: The VT340 definitely does overwrite pixels in the last sixel line, but the cases which trigger it happen very rarely as long as a text newline is sent after the sixel image. Here is an image from some of the [testing](https://github.com/hackerb9/vt340test/blob/main/sixeltests/textcursor.sh) I had done to figure out where the text cursor would be positioned based on line endings (only NL, only GNL, or neither). 
Test of Graphics NL and Text NL after sixel images of different height

First, I should state that it seems like the intention was for Sixel images to _not_ include a GLF at the end, but for there to be a text LF after the image if text or another image is going to be written. As @j4james pointed out to me, this allows one to show a full screen sixel image without causing scrolling. 

Second, when only LF is used at the end of the sixel data, 90% of the time there will be no overwrite. In actual practice, the glitch is much rarer than 10%. That is because people usually have images heights that are divisible by ten or a power of two or the character cell height, none of which can trigger what I call the [VT340 text cursor glitch](https://github.com/hackerb9/vt340test/blob/main/glitches.md#text-cursor-is-left-one-row-too-high-for-certain-sixel-heights). 

Personally, I think it is a programming mistake in the VT340 firmware since I cannot see anything in the documentation that would tip off an application developer to expect that behaviour. And, there does not appear to be any point to it. 

I am not sure how to emulate it, but you can predict it on the VT340. For a sixel image of height $h$, let $a=(h-1)\bmod6$ and $b=(h-1)\bmod20$, then, the text will overlap the image when $a>b$.

j4james: @hackerb9 If you want to see some serious overlap, try this:

    echo -e '\eP1q#"60;1#1!360~\e\\'

Utkarsh-khambra:
Thanks everybody for help in this. I got it to output similar image for [this test as well](https://github.com/hackerb9/vt340test/blob/main/sixeltests/textcursor.sh), only discrepancy I see is with horizontal alignment but I am pretty sure this is because contour has 9px wide cells for my testing config.
Screenshot from 2022-09-22 12-51-54

christianparpart (approved):
Thanks @Utkarsh-khambra for your work, and thanks @jerch, @hackerb9 and @j4james for your very valuable feedback and the joined discussion over here. :-)

hackerb9:
> Thanks everybody for helping me in this. I got it to output similar image for [this test as well](https://github.com/hackerb9/vt340test/blob/main/sixeltests/textcursor.sh), only discrepancy I see is with horizontal alignment but I am pretty sure this is because contour has 9px wide cells for my testing config. Screenshot from 2022-09-22 12-51-54

Wow! I did not expect anybody would replicate that behaviour and I am impressed. I do think it is a firmware bug that a text newline is not always sufficient to separate sixels from text and I suggest new terminals, like Contour, may want to make precise bug-for-bug VT340 compatibility available but disabled by default.

By the way, now that you've duplicated the overlap glitch, how is it affected when the size of the font goes up? If I'm correct, anything taller than 10×20 will reduce the percentage of "bad" image sizes that cause overlap.

Utkarsh-khambra:
> may want to make precise bug-for-bug VT340 compatibility available but disabled by default.

That's a good suggestion.
Here are some tests that I did

<table>
<tr>
<th> Cell Height</th>
<th> Overlap </th>
 </tr>
<tr>
<td>15</td> <td>No</td>
</tr>
<tr>
<td>17</td> <td>No</td>
</tr>
<tr>
<td>18</td> <td>No</td>
</tr>
<tr>
<td>23</td> <td>No</td>
</tr>
<tr>
<td>25</td> <td>No</td>
</tr>
<tr>
<td>27</td> <td>Yes</td>
</tr>
<tr>
<td>28</td> <td>Yes</td>
</tr>
<tr>
<td>30</td> <td>No</td>
</tr>
<tr>
<td>31</td> <td>No</td>
</tr>
<tr>
<td>33</td> <td>No</td>
</tr>
<tr>
<td>35</td> <td>No</td>
</tr>
<tr>
<td>36</td> <td>No</td>
</tr>
<tr>
<td>38</td> <td>No</td>
</tr>
<tr>
<td>40</td> <td>Yes</td>
</tr>
<tr>
<td>41</td> <td>Yes</td>
</tr>
<tr>
<td>43</td> <td>No</td>
</tr>
<tr>
<td>44</td> <td>No</td>
</tr>
<tr>
<td>46</td> <td>No</td>
</tr>
<tr>
<td>48</td> <td>No</td>
</tr>
<tr>
<td>49</td> <td>No</td>
</tr>
<tr>
<td>51</td> <td>No</td>
</tr>
<tr>
<td>53</td> <td>No</td>
</tr>
<tr>
<td>54</td> <td>No</td>
</tr>
<tr>
<td>56</td> <td>No</td>
</tr>
<tr>
<td>57</td> <td>No</td>
</tr>
<tr>
<td>58</td> <td>No</td>
</tr>
<tr>
<td>61</td> <td>No</td>
</tr>
<tr>
<td>62</td> <td>No</td>
</tr>
<tr>
<td>64</td> <td>No</td>
</tr>
</table>

hackerb9:
> > may want to make precise bug-for-bug VT340 compatibility available but disabled by default.
> 
> That's a good suggestion. Here are some tests that I did [...]

That looks close, but not what I would have predicted. You're only testing image heights 84 and 96, right?

If we can extrapolate from the VT340 glitch, I would expect the overlap bug to occur when

$$
(h_i-1)\bmod6 > (h_i-1)\bmod h_c
$$

where $h_i$ is the height of the image and $h_c$ is the height of the character cell. On a VT340, $h_c$ is fixed at $20$. 

With that formula, a character cell of height 23 would have overlap and 28 would not. It's possible my formula is wrong.

But, I wouldn't spend too much time trying to extend this bug to modern terminals. It's not like there is any compatibility to worry about and it'll only cause application developers grief as they have to implement workarounds to not overlap images.

j4james:
The VT340's behavior may not be to your liking, but it's not a bug or a glitch. It actually serves a very useful purpose when you're working with images that you want to align on text boundaries.

For example, if you want to output an image that is exactly one text cell in height, you need to output 4 sixel bands (24px height) with the last 4 pixels being transparent. If you end that with a GNL, the text cursor is exactly below the image.

Similarly if you want an image two text cells in height, it's 7 sixel bands (42px height) with the last 2 pixels transparent. And for an image three text cells in height, it's exactly 6 sixel bands (60px height).

In every case the text cursor ends up exactly below the visible portion of the image, overwriting only the transparent pixels. And if you leave off the final GNL, you can get the text cursor on the bottom row of the image instead.

Example test cases:

```
echo -en '\eP9;1q#1!800~-!800~-!800~-!800B-\e\\'
echo -en '\eP9;1q#1!800~-!800~-!800~-!800~-!800~-!800~-!800N-\e\\'
echo -en '\eP9;1q#1!800~-!800~-!800~-!800~-!800~-!800~-!800~-!800~-!800~-!800~-\e\\'
```

Please do not try to "fix" sixel. If you don't like it, and feel the need to change the standard, rather put that effort into a new graphics protocol instead.

hackerb9:
> Please do not try to "fix" sixel. If you don't like it, and feel the need to change the standard, rather put that effort into a new graphics protocol instead.

I respectfully disagree with you, but I also see that one of us is misunderstanding the problem. It's probably me and I want to understand, so let's discus that on a new bug so as to not clutter up this one: https://github.com/hackerb9/vt340test/issues/25.

PR #845, opened 2022-10-06
Fixes vi cursor color

Utkarsh-khambra:
# Creating a pull request

Thank you for taking the time to create a pull request in this repository!
Please fill out this form before submitting your PR.

Please provide a general summary of your changes in the Title above.

```markdown
Bugfix in vi cursor color.
```

## Description

```markdown
This pr fixes a bug in vi mode cursor. 
```

## Motivation and Context

If you print some text like `cat someFile` and then switch to VI mode and use vi keybinds to move around, sometimes
cursor is just completely invisible.

## How Has This Been Tested?

## Checklist:

- [x] I have read the [**`CONTRIBUTING`**](/CONTRIBUTING.md) document in my spoken language, and understand the terms

### DIFF
--- a/src/terminal/RenderBufferBuilder.cpp
+++ b/src/terminal/RenderBufferBuilder.cpp
@@ -449,7 +449,8 @@ ColumnCount RenderBufferBuilder<Cell>::renderUtf8Text(CellLocation screenPositio
     auto graphemeClusterSegmenter = unicode::utf8_grapheme_segmenter(text);
     for (u32string const& graphemeCluster: graphemeClusterSegmenter)
     {
-        auto const gridPosition = terminal.viewport().translateScreenToGridCoordinate(screenPosition);
+        auto const gridPosition = terminal.viewport().translateScreenToGridCoordinate(
+            screenPosition + ColumnOffset::cast_from(columnCountRendered));
         auto const [fg, bg] = makeColorsForCell(gridPosition,
                                                 textAttributes.flags,
                                                 textAttributes.foregroundColor,
--- a/metainfo.xml
+++ b/metainfo.xml
@@ -112,6 +112,7 @@
           <li>Adds VT sequence DECSCA, DECSEL, DECSED and DECSERA to support protected grid areas during erase operations (#29, #30, #31).</li>
           <li>Improve Input Method (IME) handling, visualizing preedit-text.</li>
           <li>Update Unicode data to version 15.0.0 (release). See Announcing The Unicode® Standard, Version 15.0.0.</li>
+          <li>Fixes cursor highlight in VI mode</li>
         </ul>
       </description>
     </release>


christianparpart:
NB: @Utkarsh-khambra i've removed some spaces in the top posting. :)

src/terminal/RenderBufferBuilder.cpp
```diff
@@ -461,16 +462,15 @@ ColumnCount RenderBufferBuilder<Cell>::renderUtf8Text(CellLocation screenPositio
...
-                                   screenPosition.column + ColumnOffset::cast_from(columnCountRendered)));
+        output.cells.emplace_back(makeRenderCellExplicit(terminal.colorPalette(),
+                                                         graphemeCluster,
+                                                         width,
+                                                         textAttributes.flags,
+                                                         fg,
+                                                         bg,
+                                                         textAttributes.underlineColor,
+                                                         baseLine + screenPosition.line,
+                                                         screenPosition.column));
```
christianparpart, line 473: i think this is just a reformat, isn't it? I was trying to figure what the difference here is, but seems like it's only changing the formatting, right?

christianparpart: EDIT: Ah, no, the last line actually changed. wasn't that obvious from the start.

The column offsetting was removed and instead put to the top.

src/terminal/RenderBufferBuilder.cpp
```diff
@@ -449,6 +449,7 @@ ColumnCount RenderBufferBuilder<Cell>::renderUtf8Text(CellLocation screenPositio
     auto graphemeClusterSegmenter = unicode::utf8_grapheme_segmenter(text);
     for (u32string const& graphemeCluster: graphemeClusterSegmenter)
     {
+        screenPosition.column = ColumnOffset::cast_from(columnCountRendered);
```
christianparpart, line 452: I think this won't work if `screenPosition.colum` - as passed by the caller - is a non-zero value.
This code assumes that `screenPosition.column` would always be zero and then offsetting into it by columnCountRendered works. It wont however when `screenPosition.column` is greater than `0`.

This won't be generally the case for standard text, but (I'm not sure), it may very well be that for non-terminal text, when this function is used, the screen position might not always be forced to start at the left-most column (`0`).

NB: This function is reused for the host-programmable and indicator statusline.

src/terminal/RenderBufferBuilder.cpp
```diff
@@ -449,6 +449,7 @@ ColumnCount RenderBufferBuilder<Cell>::renderUtf8Text(CellLocation screenPositio
...
         auto const gridPosition = terminal.viewport().translateScreenToGridCoordinate(screenPosition);
```
christianparpart, line 453:
```suggestion
        auto const gridPosition = terminal.viewport().translateScreenToGridCoordinate(screenPosition + ColumnOffsset::cast_from(columnCountRendered));
```

Instead of modifying `screenPosition`, I think this should do it, too (then the lines below should probably be undone).

christianparpart:
p.s.: I think that's worth mentioning in the changelog (see metainfo.xml)

christianparpart (approved):
Thanks.

PR #847, opened 2022-10-12
Add support for infinite grid 

Yaraslaut:
## Description

```markdown
added support of infinite grid 
```

## Motivation and Context

```markdown
At present in configuration file there is a note on possibility of using infinite grid by setting limit of history to -1. But contour doesn’t  support this option right now and sets history to 0
```

## How Has This Been Tested?

- [x] visually 
- [x] unit tests 

## Checklist:

- [x] I have not updated (or added) the documentation accordingly.


src/terminal/Grid.h
```diff
@@ -563,10 +563,24 @@ class Grid
...
+    // {{{
+    void defineMaxHistoryLineCount_(LineCount _maxHistoryLineCount)
+    {
+        if(unbox<int>(_maxHistoryLineCount)==-1)
+        {
+            maxHistoryLineCount_ = LineCount::cast_from(0);
+            isInfinite_ = true;
+        }else {
+            maxHistoryLineCount_ = _maxHistoryLineCount;
+            isInfinite_ = false;
```
christianparpart, line 575: While this works, this makes both variables dependant on each other.

Isn't it that either you have a maximum history line count or you have no limit ag all (infinite)? So with this in mind I think having both exclusive makes more sense to me, that is (something similar to):

```cpp
struct Infinite {};
using MaxHistoryLineCount = std::variant<LineCount, Infinite>;
```
Then make use of this strong type below and use `MaxHistoryLineCount` rather than `LineCount` for storing this value. That would also revert the required change in the constructor again (not needed).

The implication would be that you'd need to touch more code, all the lines that read from `maxHistoryLineCount_`, but wouldn't that be correct?

NB: Since sometimes `maxHistoryLineCount_` is being read from, even though it's meant to be infinite (optionally), I think you could achieve the same by reading the value `lines_.size()`, since this is pre-allocated anyways (with history line count + main page line couynt).

What do you think?

christianparpart: p.s.: It would mean more coding (sorry for that), but I think that would make the intend much more clear and less error prone in future code changes due to less side data dependent variables one has to take care of.

src/terminal/Grid.cpp
```diff
@@ -123,11 +123,12 @@ template <typename Cell>
 Grid<Cell>::Grid(PageSize _pageSize, bool _reflowOnResize, LineCount _maxHistoryLineCount):
     pageSize_ { _pageSize },
     reflowOnResize_ { _reflowOnResize },
-    maxHistoryLineCount_ { _maxHistoryLineCount },
```
christianparpart, line 126: See below for why I personally think that's not an ideal solution. :-)

src/terminal/Grid.h
Yaraslaut: > While this works, this makes both variables dependant on each other.
> 
> Isn't it that either you have a maximum history line count or you have no limit ag all (infinite)? So with this in mind I think having both exclusive makes more sense to me, that is (something similar to):
> 
> ```c++
> struct Infinite {};
> using MaxHistoryLineCount = std::variant<LineCount, Infinite>;
> ```
> 
> Then make use of this strong type below and use `MaxHistoryLineCount` rather than `LineCount` for storing this value. That would also revert the required change in the constructor again (not needed).
> 
> The implication would be that you'd need to touch more code, all the lines that read from `maxHistoryLineCount_`, but wouldn't that be correct?
> 
> NB: Since sometimes `maxHistoryLineCount_` is being read from, even though it's meant to be infinite (optionally), I think you could achieve the same by reading the value `lines_.size()`, since this is pre-allocated anyways (with history line count + main page line couynt).
> 
> What do you think?

I've updated my implementation, now it is based on variant

src/terminal/Grid.cpp
```diff
@@ -123,9 +125,22 @@ template <typename Cell>
 Grid<Cell>::Grid(PageSize _pageSize, bool _reflowOnResize, LineCount _maxHistoryLineCount):
```
christianparpart, line 125:
```suggestion
Grid<Cell>::Grid(PageSize _pageSize, bool _reflowOnResize, MaxHistoryLineCount _maxHistoryLineCount):
```
I think it would be best to also adapt the calling parameters to strongly type the max history line count. This way you enable the caller to explicitly pass `Infinite {}` (which btw should not be a child struct of `Grid<T>` in order to easy usage of that type. Just move this empty struct to `<terminal/primitives.h>`, somewhere around `LineCount`, same for `MaxHistoryLineCount`, such that you can even make use of `MaxHistoryLineCount` in the `Config` struct to properly reflect infinite scroll history state in a typed way (rather than knowing that -1 means infinite).

src/terminal/Grid_test.cpp
```diff
@@ -846,4 +846,24 @@ TEST_CASE("Grid.reflow.tripple", "[grid]")
...
+TEST_CASE("Grid infinite", "[grid]")
+{
+    auto grid_finite = Grid<Cell>(PageSize { LineCount(2), ColumnCount(8) }, true, LineCount(0));
+    grid_finite.setLineText(LineOffset { 0 }, "ABCDEFGH"sv);
+    grid_finite.setLineText(LineOffset { 1 }, "abcdefgh"sv);
+    grid_finite.scrollUp(LineCount { 1 });
+    REQUIRE(grid_finite.lineText(LineOffset(0)) == "abcdefgh");
+    REQUIRE(grid_finite.lineText(LineOffset(-1)) == std::string(8, ' '));
+
+    auto grid_infinite = Grid<Cell>(PageSize { LineCount(2), ColumnCount(8) }, true, LineCount(-1));
```
christianparpart, line 859:
```suggestion
    auto grid_infinite = Grid<Cell>(PageSize { LineCount(2), ColumnCount(8) }, true, Infinite {});
```
okay, I know this code change won't work because I think `Grid`'s constructor does not accept a `MaxHistoryLineCount` yet (see comment above), but I think that would be the best way to also visually and typed express the intent here.

src/terminal/Grid.cpp
```diff
@@ -22,6 +22,8 @@
 #include <algorithm>
 #include <iostream>
 
+#include "terminal/primitives.h"
```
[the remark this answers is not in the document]
Yaraslaut: > Then auto-format (e.g. clang-format -i src/terminal/Grid.cpp` should place this line also a few lines above, to the other sibling includes.

Done

src/terminal/Grid.cpp
```diff
@@ -123,9 +125,22 @@ template <typename Cell>
 Grid<Cell>::Grid(PageSize _pageSize, bool _reflowOnResize, LineCount _maxHistoryLineCount):
     pageSize_ { _pageSize },
     reflowOnResize_ { _reflowOnResize },
-    maxHistoryLineCount_ { _maxHistoryLineCount },
+    historyLimit_ { [&]() -> MaxHistoryLineCount {
```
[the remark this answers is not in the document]
Yaraslaut: > Let's capture only the minimal amount, explicitly. :)
lambda is gone after refactor

src/terminal/Grid.h
```diff
@@ -383,21 +383,40 @@ class Grid
...
 
     Grid(): Grid(PageSize { LineCount(25), ColumnCount(80) }, false, LineCount(0)) {}
 
     void reset();
 
     // {{{ grid global properties
-    [[nodiscard]] LineCount maxHistoryLineCount() const noexcept { return maxHistoryLineCount_; }
+    [[nodiscard]] LineCount maxHistoryLineCount() const noexcept
+    {
+        if (const LineCount* maxLineCount = std::get_if<LineCount>(&historyLimit_))
```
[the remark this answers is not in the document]
Yaraslaut: > (it's not so much about the `auto`, but mind the position of the `const` :)

Done

src/terminal/Grid.h
```diff
@@ -563,11 +582,13 @@ class Grid
     void rotateBuffersRight(LineCount count) noexcept { lines_.rotate_right(unbox<size_t>(count)); }
     // }}}
 
+    // {{{
+    // }}}
```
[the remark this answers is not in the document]
Yaraslaut: > ?

Deleted

src/terminal/Grid.cpp
Yaraslaut: > I think it would be best to also adapt the calling parameters to strongly type the max history line count. This way you enable the caller to explicitly pass `Infinite {}` (which btw should not be a child struct of `Grid<T>` in order to easy usage of that type. Just move this empty struct to `<terminal/primitives.h>`, somewhere around `LineCount`, same for `MaxHistoryLineCount`, such that you can even make use of `MaxHistoryLineCount` in the `Config` struct to properly reflect infinite scroll history state in a typed way (rather than knowing that -1 means infinite).

Changed it, see 'Refactor' commit

src/terminal/Grid_test.cpp
Yaraslaut: > okay, I know this code change won't work because I think `Grid`'s constructor does not accept a `MaxHistoryLineCount` yet (see comment above), but I think that would be the best way to also visually and typed express the intent here.

Changed, see 'Refactor' commit

christianparpart (approved):
Many thanks.

PR #851, opened 2022-10-19
Preserve TrivialLineBuffer on grid resize

### LINKED ISSUE #840: grid resize should make more use of `TrivialLineBuffer`
Especially doing process startup, it sometimes seems to be the case that there is

1. initial window size (say: 80x25)
2. then getting resized to (80x2)
3. later on resized to, say (130x34)

While it is questionable why this is happening (GUI resize events), this all happens while the grid buffer is still completely empty. Initially the Grid buffer has all lines set to **empty** `TrivialLineBuffer`. But with a resize, this changes to `InflatedBuffer`, even though this could ideally remain a trivial line buffer.

What should be touched?

- [ ] `Line<Cell>::resize()` should be refactored to not inflate the buffer when resizing a trivial line buffer, if possible.
- [ ] `Grid<Cell>::resize()` should probably need to be touched as well due to the line wrapping feature.

Yaraslaut:
Closes #840

### DIFF
--- a/src/terminal/Line.cpp
+++ b/src/terminal/Line.cpp
@@ -34,7 +34,7 @@ typename Line<Cell>::InflatedBuffer Line<Cell>::reflow(ColumnCount _newColumnCou
     {
         switch (crispy::strongCompare(_newColumnCount, ColumnCount::cast_from(trivialBuffer().text.size())))
         {
-            case Comparison::Greater: trivialBuffer().displayWidth = _newColumnCount; return{};
+            case Comparison::Greater: trivialBuffer().displayWidth = _newColumnCount; return {};
             case Comparison::Equal: return {};
             case Comparison::Less:;
         }
--- a/src/terminal/Line.cpp
+++ b/src/terminal/Line.cpp
@@ -13,6 +13,7 @@
  */
 #include <terminal/GraphicsAttributes.h>
 #include <terminal/Line.h>
+#include <terminal/primitives.h>
 
 #include <unicode/grapheme_segmenter.h>
 #include <unicode/utf8.h>
@@ -29,6 +30,15 @@ template <typename Cell>
 typename Line<Cell>::InflatedBuffer Line<Cell>::reflow(ColumnCount _newColumnCount)
 {
     using crispy::Comparison;
+    if (isTrivialBuffer())
+    {
+        switch (crispy::strongCompare(_newColumnCount, ColumnCount::cast_from(trivialBuffer().text.size())))
+        {
+            case Comparison::Greater: trivialBuffer().displayWidth = _newColumnCount;
+            case Comparison::Equal: return {};
+            case Comparison::Less:;
+        }
+    }
     auto& buffer = inflatedBuffer();
     // TODO: Efficiently handle TrivialBuffer-case.
     switch (crispy::strongCompare(_newColumnCount, size()))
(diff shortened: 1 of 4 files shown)


src/terminal/Line.cpp
```diff
@@ -29,8 +29,13 @@ template <typename Cell>
 typename Line<Cell>::InflatedBuffer Line<Cell>::reflow(ColumnCount _newColumnCount)
 {
     using crispy::Comparison;
+    if (isTrivialBuffer())
+    {
+        auto buffer = TrivialBuffer();
+        buffer.displayWidth = _newColumnCount;
+        return {};
```
christianparpart, line 36: Does this make sense?

christianparpart: What confuses me is that two lines above you create a new variable of type `TrivialBuffer`, assign some value to `displayWidth` right after, but then don't use that variable. Did you probably mean this instead (two lines above)?

```cpp
auto& buffer = trivialBuffer();
```

(p.s.: Use `&` to denote that you are actually also want to use a reference, that might have maybe caught the typo.)

Yaraslaut: Yes, I've made two typos there :)

src/terminal/Line.cpp
```diff
@@ -29,8 +29,13 @@ template <typename Cell>
 typename Line<Cell>::InflatedBuffer Line<Cell>::reflow(ColumnCount _newColumnCount)
 {
     using crispy::Comparison;
+    if (isTrivialBuffer())
+    {
+        auto& buffer = trivialBuffer();
+        buffer.displayWidth = _newColumnCount;
```
christianparpart, line 35: I think this optimization can only be done if the text on that line is smaller or equal to the `_newColumnCount`. Otherwise the slow path could be taken.

What would happen now is that if you shrink only the columns, they'll not be wrapped around but simply not shown whatever would overflow (or wrap around).

src/terminal/Line_test.cpp
```diff
@@ -36,6 +36,57 @@ TEST_CASE("Line.BufferFragment", "[Line]")
...
+{
+    auto constexpr DisplayWidth = ColumnCount(4);
+    auto text = "abcd"sv;
+    auto pool = BufferObjectPool(32);
+    auto bufferObject = pool.allocateBufferObject();
+    bufferObject->writeAtEnd(text);
+
+    auto const bufferFragment = bufferObject->ref(0, 4);
+
+    auto sgr = GraphicsAttributes {};
```
christianparpart, line 49:
```suggestion
    auto const sgr = GraphicsAttributes {};
```

src/terminal/Line_test.cpp
```diff
@@ -36,6 +36,57 @@ TEST_CASE("Line.BufferFragment", "[Line]")
...
```
christianparpart, line 76:
```suggestion
    auto const sgr = GraphicsAttributes {};
```
Now I am questioning myself. You use `const` everywhere else by default, which is good, except for `sgr`. What am I missing? 🤔

christianparpart (commented):
Sorry for the late review. It would be best if you do a `git fetch && git rebase origin/master` and fixing the conflict shouldn't be hard. Tell me if you need assistance there. :)

src/terminal/Line_test.cpp
Yaraslaut: You are not missing anything, that was me who missed it, I just copy pasted it from another unit test where some GraphicsAttributes were used

Yaraslaut:
@christianparpart  I updated PR, please take a look
Also, please squash on merge :)

christianparpart (approved):
Thanks @Yaraslaut 😀

PR #856, opened 2022-10-26
Adding new Cell storage implementation and restructure Cell storage related files

christianparpart:
- grouping Cell storage related classes into common sub directory: `src/terminal/cell/`
- rename `Cell` to `CompactCell`
- adding new `SimpleCell` but only making use of it for indicator status line (for now)
- adding Cell configuration file `<terminal/cell/CellConfig.h>` as single central shared place to determine what type of cell to use for which buffer (good for easy reconfiguration).


data-man:
I'm not sure `Thin` is a good name.

christianparpart:
> I'm not sure `Thin` is a good name.

Naming is one of the hardest things in Tech. I am open for proposals. The actual struct is indeed thin though, because it's laying out all the lesser used properties to `CellExtra`.

christianparpart:
> `SimpleCell`?
> `SimplifiedCell`?
> `ReducedCell`?
> `BasicCell`?
> `EasyCell`?

Simle, Simplified, Basic, are all not correct (IMHO), because that's not what they are. Reduced could remotely make sense as it's reducing the storage size of the main cell instance by layering off the lesser used ones, but that makes it no way near simple or basic, nor easy.

data-man:
> Reduced could remotely make sense as it's reducing the storage size of the main cell instance by layering off the lesser used ones

`CompactCell`?
`CompressedCell`?

christianparpart:
I'll keep the code changes due to being a good refactor for cleaner code, but the initial attempt to get the alt-screen faster with that failed. I'll research into that direction later as it's now really trivial to exchange one cell type of another for each type of screen individually.

PR #863, opened 2022-10-30
[CI] Upgrade GitHub Actions scripts, dependencies

whisperity:
- Fixed deprecation warnings around `save-state` and `set-output` [1]
 - Fixed deprecation issue with older versions of reused NodeJS-based workflows [2]

[1]: http://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands
[2]: http://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12


.github/workflows/labeler.yml
```diff
@@ -7,6 +7,6 @@ jobs:
   triage:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/labeler@v3
+    - uses: actions/labeler@v4
```
whisperity, line 10: `v4` is the newest and yet GitHub did not upgrade the target NodeJS version of their code... :shrug:

.github/workflows/build.yml
```diff
@@ -313,7 +313,7 @@ jobs:
           ./scripts/install-deps.ps1 --skip-vcpkg
           type ./_deps/sources/CMakeLists.txt
       - name: Install Qt
-        uses: jurplel/install-qt-action@v2
+        uses: jurplel/install-qt-action@v3
       - name: "vcpkg: Install dependencies"
         uses: lukka/run-vcpkg@v5
```
whisperity, line 318: Unfortunately, I could not upgrade this one to v**10** (even though that's available and would fix the deprecation of `set-output` used *internally* in the action) because then [the Windows build would fail](https://github.com/contour-terminal/contour/actions/runs/3355085839/jobs/5559094033):

```
CMake Error at D:/a/contour/vcpkg/scripts/buildsystems/vcpkg.cmake:829 (_find_package):
Error:   By not providing "Findfmt.cmake" in CMAKE_MODULE_PATH this project has
-- Configuring incomplete, errors occurred!
  asked CMake to find a package configuration file provided by "fmt", but
See also "D:/a/contour/contour/build/CMakeFiles/CMakeOutput.log".
See also "D:/a/contour/contour/build/CMakeFiles/CMakeError.log".
  CMake did not find one.

  Could not find a package configuration file provided by "fmt" with any of
  the following names:

    fmtConfig.cmake
    fmt-config.cmake

  Add the installation prefix of "fmt" to CMAKE_PREFIX_PATH or set "fmt_DIR"
  to a directory containing one of the above files.  If "fmt" provides a
  separate development package or SDK, be sure it has been installed.
Call Stack (most recent call first):
  cmake/ContourThirdParties.cmake:58 (find_package)
  CMakeLists.txt:87 (include)


Error: Process completed with exit code 1.
```

.github/workflows/release.yml
```diff
@@ -64,12 +64,12 @@ jobs:
           vcpkgTriplet: x64-windows
```
whisperity, line 64: Ditto, in this file.

.github/workflows/build.yml
christianparpart: The fmt thing we can address separately.

christianparpart (approved):
Oh thanks. :)

Issue #870, opened 2022-11-05
Assertion failiure in contour on searching for text that got split because of line wrapping


Utkarsh-khambra:
### Contour Terminal version

Contour Terminal Emulator 0.3.8-master-4e509121

### Installer source

Github: source code cloned

### Operating System

Fedora 36

### Architecture

x86-64

### Steps to reproduce

1. Compile a debug build.
2. Write a long text without any newllne so that line gets wrapped and make sure that because of wrapping a word must get split and it must be a unique word in the whole line.
2. Search for the word that gets split.

### Expected Behavior

Word that gets searched must get highlighted

### Actual Behavior

Contour segfaluts if the word is unique in the wrapped line.

https://user-images.githubusercontent.com/29328319/200104643-f1f65177-9020-4c8e-9cbd-aa2d3286f1dd.mp4

Issue #869, opened 2022-11-05
Contour doesn't hightlight the word that gets broken due to line wrapping


Utkarsh-khambra:
### Contour Terminal version

Contour Terminal Emulator 0.3.8-master-4e509121

### Installer source

Github: source code cloned

### Operating System

Fedora 36

### Architecture

x86-64

### Steps to reproduce

1. Write text in terminal that is so long that the line gets wrapped, also a word must also get split in two parts because of wrapping and multiple instances of that word must apear.
2. Go into search mode and search for the word that got split into two parts.
3. Jump to next or previous matches

### Expected Behavior

The jumping between matches should highlight matches that the cursor is on.

### Actual Behavior

The cursor skips the match that got split because of line wrapping.

https://user-images.githubusercontent.com/29328319/200102154-ae188df8-28ed-4e31-acb5-a4e85d9907a0.mp4

PR #873, opened 2022-11-13
Vi mode search takes into account line wrapping

Utkarsh-khambra:
## Description
Makes `Line<Cell>::search` and `Line<Cell>::searchReverse` return appropriate result even if a partial match is found on 

## Motivation and Context

see #869 and #870 

```markdown
These changes now make Vi  search mode take into the line wrapping. So now if a word gets cut off due to line wrapping it still gets matched.  
```

## How Has This Been Tested?

Visually.

## Checklist:


src/terminal/Grid.h
```diff
@@ -147,6 +147,59 @@ struct LogicalLine
...
+            ++i;
+        }
+        return std::nullopt;
+    }
+
+    // Searches from right to left, taking into account line wrapping
+    [[nodiscard]] std::optional<terminal::CellLocation> searchReverse(std::u32string_view searchText,
+                                                                      ColumnOffset startPosition) const
+    {
+        // TODO Handle searchText larger than the line length
```
christianparpart, line 182: FYI: We've a tiny CI check such that if you have comments like "// TODO(pr): ..." the CI check will fail. I personally use that to mark things as TODO and prevent accidentally merging because I forgot about it.

But i think that one here is for future reference, right? We could create'n'ref a ticket for that right away.

src/terminal/Grid.h
```diff
@@ -199,7 +252,7 @@ struct LogicalLines
                 return *this;
             }
 
-            Require(!lines.get()[unbox<int>(next)].wrapped());
+            // Require(!lines.get()[unbox<int>(next)].wrapped());
```
christianparpart, line 255: This is a "logical line iterator" so it should give access to logical lines in an atomic way, e.g. spanning multiple lines. So I think `wrapped` flag must be always unset. Why did you comment this out?

src/terminal/Line.h
```diff
@@ -373,82 +377,100 @@ class Line
...
     }
 
-    // Searches for the given text in this line starting at a specific column.
-    // Returns either the location of the beginning of found search term or the same input start column if
-    // not.
-    [[nodiscard]] std::optional<ColumnOffset> search(std::u32string_view text,
-                                                     ColumnOffset startColumn) const noexcept
+    // Searches for the given text in the line at a specific column.
+    // Returns SearchResult with column set to location of beginning of found location or std::nullopt and
+    // remainingText set to 0
```
christianparpart, line 382:
```suggestion
    // remainingText set to 0.
```

src/terminal/primitives.h
```diff
@@ -893,6 +894,12 @@ constexpr unsigned setDynamicColorCommand(DynamicColorName name)
     }
 }
 
+struct SearchResult
```
christianparpart, line 897: I think the field members & struct deserve some documentation (even if it might sound trivial to you) :-)

Not sure I get it right, but column is set to the beginning of the match or nullopt if not found. what's remainingText? My point is, maybe `column` should not be optional but both values mandatory and the whole `SearchResult` should be returned by the search functions as `optional<SearchResult>` instead, denoting a "not found" when a `nullopt` was returned. Does sound more intuitive to me.

I'll need to have a second look at this PR later with more time though.

src/terminal/Grid.h
Utkarsh-khambra: Nah I'm thinking of also finishing it with this pr now. So I'll remove the comment later.

src/terminal/Grid.h
Utkarsh-khambra: I also added a function `logicalLinesFrom(LineOffset offset)` which is supposed to give you all the logical lines from offset to ` pageSize_.lines - 1` which mean you can specify offset to be a line number in the middle of a group of wrapped lines, if that happens then when you increment the iterator this precondition will fail.

src/terminal/Grid.h
```diff
@@ -147,6 +147,201 @@ struct LogicalLine
...
+                                                       - unbox<size_t>(startPosition),
+                                                   unbox<size_t>(startPosition));
+                auto const result = searchPartialMatchReverse(textOnThisLine, line->get());
+                if (result != 0)
+                {
+                    std::u32string_view remainingText(searchText.data(), searchText.size() - result);
+                    // Check if the searchText can even fit in the available lines
+                    auto const willFit = [&] {
+                        auto const count =
+                            static_cast<size_t>(std::max(0l, std::distance(line + 1, lines.rend())));
```
christianparpart, line 219: In this line, MSVC/Windows CI build is failing. It cannot find a proper overload for `std::max(A, B)` most likely because the type of A and B is not equal, which it must be.

```suggestion
                            static_cast<size_t>(std::max(long { 0 }, static_cast<long>(std::distance(line + 1, lines.rend()))));
```

This could do it. I'm sorry, I didn't verify, as I'm just typing in my web browser, but you simply need to make sure the type of each parameter is equal to each other. :)

Issue #897, opened 2022-12-09
Wrong placement of cursor after exit from `screen`


Yaraslaut:
### Contour Terminal version

Contour Terminal Emulator 0.3.5-master-6e047fd9

### Installer source

Github: source code cloned

### Operating System

Manjaro Linux

### Architecture

x86-64

### Other Software

Screen version 4.09.00 (GNU) 30-Jan-22

### Steps to reproduce

https://user-images.githubusercontent.com/44506630/206787815-f2a35ed9-9eed-4379-a409-be7fd4f14c24.mp4

### Expected Behavior

Similar to tmux

### Actual Behavior

Cursor is placed on second line from the top of the window

### Additional notes
vt trace
```
[2022-12-10 13:35:42.196238] [vt.trace.sequence] control U+0D (Carriage Return)
[2022-12-10 13:35:42.196272] [vt.trace.sequence] control U+0A (Line Feed)
[2022-12-10 13:35:42.200540] [vt.trace.sequence] Handle VT sequence: CSI ? 1 l
[2022-12-10 13:35:42.200579] [vt.trace.sequence] Handle VT sequence: ESC >
[2022-12-10 13:35:42.200598] [vt.trace.sequence] Handle VT sequence: CSI 28;1 H
[2022-12-10 13:35:42.200613] [vt.trace.sequence] control U+0D (Carriage Return)
[2022-12-10 13:35:42.200623] [vt.trace.sequence] control U+0A (Line Feed)
[2022-12-10 13:35:42.200639] [vt.trace.sequence] Handle VT sequence: CSI ? 1049 l
[2022-12-10 13:35:42.200873] [vt.trace.sequence] Handle VT sequence: CSI 23;0;0 t
```

whisperity:
Is this Contour-specific? A quick guess from my end is that the problem is not the cursor placement *per se*, but the fact that the existing buffer is not wiped. I guess screen is dumb enough to always reposition the exit to the top left corner first.

I tried the same ordeal in PuTTY, and right now I get the existing buffer properly cleaned, and the cursor properly placed on the line after the `[screen is terminating]` printout.

https://user-images.githubusercontent.com/1969470/206794000-f7213cf8-13b0-427e-839e-24690edd46a4.mp4

 * What happens with tmux?
 * Are you using a multiline prompt?

Yaraslaut:
This is what is happening for  `tmux`. So, everything is alright 

[...]

PR #902, opened 2022-12-11
Don't move cursor when setting top,bottom,left and right margins.

Utkarsh-khambra:
## Description

```markdown
Your issue description (body) goes here
```

## Motivation and Context
see #897 

```markdown

```

## How Has This Been Tested?

Visually and added a test
No effect
## Checklist:

### DIFF
--- a/src/vtbackend/Screen.cpp
+++ b/src/vtbackend/Screen.cpp
@@ -3461,6 +3461,7 @@ ApplyResult Screen<Cell>::apply(FunctionDefinition const& function, Sequence con
             auto l = decr(seq.param_opt<ColumnOffset>(0));
             auto r = decr(seq.param_opt<ColumnOffset>(1));
             _terminal.setLeftRightMargin(l, r);
+            moveCursorTo({}, {});
         }
         break;
         case DECSM: {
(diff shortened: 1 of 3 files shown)


src/vtbackend/Screen.cpp
```diff
@@ -3475,7 +3475,11 @@ ApplyResult Screen<Cell>::apply(FunctionDefinition const& function, Sequence con
             _terminal.setTopBottomMargin(decr(seq.param_opt<LineOffset>(0)),
                                          decr(seq.param_opt<LineOffset>(1)));
             break;
-        case DECSTR: _terminal.softReset(); break;
+        case DECSTR:
+            // For VTType VT100 and VT52 ignore this sequence
+            if (_terminal.state().terminalId != VTType::VT100)
+                _terminal.softReset();
```
christianparpart, line 3481:
```suggestion
            if (_terminal.state().terminalId == VTType::VT100)
                return ApplyResult::Invalid;
            _terminal.softReset();
```
if should probably be reported (good habbit) if a VT sequence was to be evaluated that shouldn't be supported by that VT type.

src/vtbackend/Terminal.cpp
```diff
@@ -1487,7 +1486,6 @@ void Terminal::setLeftRightMargin(optional<ColumnOffset> _left, optional<ColumnO
         {
             state_.margin.horizontal.from = left;
             state_.margin.horizontal.to = right;
```
christianparpart, line 1488: With this change and the one above you should apply the following to not break DECSTBM / DECSLRM:

```diff
diff --git a/src/vtbackend/Screen.cpp b/src/vtbackend/Screen.cpp
index 063ba91a..0bfa038e 100644
--- a/src/vtbackend/Screen.cpp
+++ b/src/vtbackend/Screen.cpp
@@ -3461,6 +3461,7 @@ ApplyResult Screen<Cell>::apply(FunctionDefinition const& function, Sequence con
             auto l = decr(seq.param_opt<ColumnOffset>(0));
             auto r = decr(seq.param_opt<ColumnOffset>(1));
             _terminal.setLeftRightMargin(l, r);
+            moveCursorTo({}, {});
         }
         break;
         case DECSM: {
@@ -3474,6 +3475,7 @@ ApplyResult Screen<Cell>::apply(FunctionDefinition const& function, Sequence con
         case DECSTBM:
             _terminal.setTopBottomMargin(decr(seq.param_opt<LineOffset>(0)),
                                          decr(seq.param_opt<LineOffset>(1)));
+            moveCursorTo({}, {});
             break;
         case DECSTR:
             // For VTType VT100 and VT52 ignore this sequence
```

We don't have an explicit test case for both VT sequences, because I **thought** we're using the implementation function a lot already. but now it turns out that we probably should have that (not part of this PR, necessarily :) )

christianparpart (changes requested):

christianparpart (approved):

PR #932, opened 2022-12-20
Fixes packaging for Qt6-based builds (e.g. affects Ubuntu 22.04).

### LINKED ISSUE #929: [Debian/Ubuntu] CI-built package missing dependencies that are actually dependent upon
### Contour Terminal version

0.3.9.4459 (but has been happening since 0.3.6 at least!)

### Installer source

Github: source code cloned

### Operating System

Ubuntu 22.04

### Architecture

x86-64

### Other Software

_No response_

### Steps to reproduce

Download Contour as built in the CI and install the dpkg, ask it to install all unresolved dependencies (that it knows about from the deb). A few packages will be installed, however, Contour still doesn't start:

```
contour
contour: error while loading shared libraries: libQt6OpenGLWidgets.so.6: cannot open shared object file: No such file or directory
```

Then, `apt-get install libqt6openglwidgets6`, and try again:

```
qt.qpa.plugin: Could not find the Qt platform plugin "xcb" in ""
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Aborted (core dumped)
```

Then, install `qt6-qpa-plugins` and try again, and it does start.

christianparpart:
Fixes #929

### DIFF
--- a/scripts/install-deps.sh
+++ b/scripts/install-deps.sh
@@ -210,10 +210,12 @@ install_deps_ubuntu()
             libgl1-mesa-dev
             libglvnd-dev
             libqt6opengl6-dev
+            libqt6openglwidgets6
             libqt6widgets6
             qt6-base-dev
             qt6-base-dev-tools
             qt6-multimedia-dev
+            qt6-qpa-plugins
         "
     else
         packages="$packages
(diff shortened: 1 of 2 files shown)


whisperity (approved):
In general it would be better if these would be versioned in a single source of truth but yeah I see the format is massively different sadly.

Either way, tried on real 22.04 and in Docker (22.04 guest on 20.04 host) and it seems to start the program now nicely:
Screenshot_20221221_105242

Issue #85, opened 2019-10-07
formatted copy to clipboard 


christianparpart:
Should support:

| Action Name | Description |
|---------------|--------------|
| CopyToClipboardHTML | Copies selection to clipboard in HTML format. |
| CopyToClipboardVT      | Copies selection to clipboard in VT escape sequences. |
| CopyToClipboard          | Copies selection to clipboard, but only the text (already present) |
| CopyToClipboardPNG  | Copies selection to clipboard in PNG image file format. |

theoniko:
Could i give it a try? Are there any similar already implemented actions to check as walkthrough?

christianparpart:
Oh absolutely, and many thanks for your interest.
Would you mind joining Discord for that for a more interactive guided walkthrough?

christianparpart:
@theoniko to wich of the 3 yet missing ones are you interested in the most? I think the easiest one that might have an impact on users that want it would be the HTML version. for the image (PNG) version I think some prep-work would be needed to have that trivially implemented. I could do the prep-work first such that you'd have it easier to implement `CopyToClipboardPNG` after that. So starting with `CopyToClipboardHTML` might probably be most rewarding as a starter.

Our Discord (if you haven't found it already) is here: https://discord.gg/xQCRUCtc

theoniko:
Ok, i will try to implement the CopyToClipboardHTML then. Let's talk there for the details.

theoniko:
I created a draft pr https://github.com/contour-terminal/contour/pull/942. Basically, i added an extra variable on yml and based on the string value selected (internally converted to enum class values) and in case of HTML i use https://doc.qt.io/qt-5/qstring.html#toHtmlEscaped.

PR #942, opened 2022-12-23
Implement action CopyToClipboardHTML

theoniko:
## Description

```markdown
Upon pressing ctrl h in the keyboard, action  CopyToClipboardHTML  is triggered which will copy to clipboard the selected text in html format.
```

## Motivation and Context

```markdown

```

## How Has This Been Tested?

## Checklist:

### DIFF
--- a/src/vtbackend/Terminal.cpp
+++ b/src/vtbackend/Terminal.cpp
@@ -1174,6 +1174,28 @@ string Terminal::extractSelectionText() const
     }
 }
 
+string Terminal::extractSelectionTextAsHtml() const
+{
+    auto const _lock = scoped_lock { *this };
+
+    if (!_selection)
+        return "";
+
+    if (isPrimaryScreen())
+    {
+        auto se = SelectionRenderer<PrimaryScreenCell> { *this, pageSize().columns.as<ColumnOffset>() - 1 };
+        terminal::renderSelection(*_selection, [&](CellLocation _pos) { se(_pos, _primaryScreen.at(_pos)); });
+        return se.finish();
+    }
+    else
+    {
+        auto se = SelectionRenderer<AlternateScreenCell> { *this, pageSize().columns.as<ColumnOffset>() - 1 };
+        terminal::renderSelection(*_selection,
+                                  [&](CellLocation _pos) { se(_pos, _alternateScreen.at(_pos)); });
+        return se.finish();
+    }
+}
+
 string Terminal::extractLastMarkRange() const
 {
     auto const _ = std::lock_guard { *this };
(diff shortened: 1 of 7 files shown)


christianparpart (commented):
Looking back at #85, I think we need to realize this a little bit differently.

We need to add a new action, called `CopyToClipboardHTML`, and that action handler must then be implemented in `TerminalSession` (right next to the one for `CopyToClipboard`) to place an the HTML version of the text to the clipboard rather than simply the text.

Your idea with `on_mouse_select_output_type` is an interesting one, but I think users should make an explicit choice (case by case) in what format they want to copy the selected area.

christianparpart:
An inspiration beginning would probably be something like this:

```diff
diff --git a/src/contour/Actions.h b/src/contour/Actions.h
index 18a3c815..6f86ab5a 100644
--- a/src/contour/Actions.h
+++ b/src/contour/Actions.h
@@ -22,12 +22,28 @@
 namespace contour::actions
 {
 
+// Defines the format to use when extracting a selection range from the terminal.
+enum class CopyFormat
+{
+    // Copies purely the text (with their whitespaces, and newlines, but no formatting).
+    Text,
+
+    // Copies the selection in HTML format.
+    HTML,
+
+    // Copies the selection in escaped VT sequence format.
+    VT,
+
+    // Copies the selection as PNG image.
+    PNG,
+};
+
 // clang-format off
 struct CancelSelection{};
 struct ChangeProfile{ std::string name; };
 struct ClearHistoryAndReset{};
 struct CopyPreviousMarkRange{};
-struct CopySelection{};
+struct CopySelection{ CopyFormat format = CopyFormat::Text; };
 struct CreateDebugDump{};
 struct DecreaseFontSize{};
 struct DecreaseOpacity{};
```

Then have the config loader (Config.cpp) also fill this value when parsing the YAML and making use of this enum in the implementation when text has to be extracted in the action handler.

christianparpart:
@theoniko Would you mind rebasing your PR on top of latest master?

```sh
git fetch && git rebase origin/master
```

This will bring in also the latest prep-work #943.

theoniko:
> @theoniko Would you mind rebasing your PR on top of latest master?
> 
> ```shell
> git fetch && git rebase origin/master
> ```
> 
> This will bring in also the latest prep-work #943.

Done

src/vtbackend/Terminal.cpp
```diff
@@ -1139,6 +1139,44 @@ string Terminal::extractSelectionText() const
...
+    switch(cellFlags)
+    {
+        case CellFlags::Bold: result.append("font-weight: bold;\n"); break;
+        case CellFlags::Italic: result.append("font-style: italic;\n"); break;
+        case CellFlags::Underline: result.append("text-decoration: underline;\n"); break;
+        case CellFlags::DottedUnderline: result.append("text-decoration: underline dotted;\n"); break;
+        default: 
+    }
+
+    return result+ "\n}\n";
```
christianparpart, line 1177: This function should not be part of the `Terminal` class. Please have a look at `VTWriter`, which basically translates the grid into the VT format as well.

You will need something similar to `VTWriter` but for HTML+CSS. You could give it the name `HtmlWriter` or `HtmlBuilder`.

PR #963, opened 2022-12-29
Adds CI check for clang-tidy.

christianparpart:
add CI check for clang-tidy.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -44,6 +44,34 @@ jobs:
         check-path: 'src'
         exclude-regex: 'crispy/sse2neon.h'
 
+  check_clang_tidy:
+    name: "Check C++ clang-tidy"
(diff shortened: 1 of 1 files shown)


whisperity (changes requested):
This is already performed by the CodeChecker job. It calls Tidy. And newer versions will also be able to call CppCheck!

whisperity:
Also we can set the CI job to run Clang-Tidy in a way that it respects the `.clang_tidy` files. I don't remember the flag for this at the tip of my tongue but can look it up!

christianparpart:
> This is already performed by the CodeChecker job. It calls Tidy. And newer versions will also be able to call CppCheck!

yeah, I'm **mostly** interested in enforcing a proper coding style. I'm not sure why clang-format is not doing that for the identifier naming conventions. IMHO that would be fitting there, too, but yeah...

PR #965, opened 2023-01-02
Add option to hide mouse while typing

### LINKED ISSUE #819: Make configurable: hide mouse on keyboard input
This is currently already the case but it cannot be disabled.

So, make it configurable.

Yaraslaut:
Closes #819

### DIFF
--- a/src/contour/contour.yml
+++ b/src/contour/contour.yml
@@ -220,6 +220,11 @@ profiles:
             # whether or not to hide the scrollbar when in alt-screen.
             hide_in_alt_screen: true
 
+        # mouse setting
+        mouse:
+            # whether or not to hide mouse when typing
+            hide_while_typing: true
+
         # Some VT sequences should need access permissions.
         #
         # These can be to:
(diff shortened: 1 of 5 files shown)


src/contour/contour.yml
```diff
@@ -220,6 +220,11 @@ profiles:
             # whether or not to hide the scrollbar when in alt-screen.
             hide_in_alt_screen: true
 
+        # mouse setting
+        mouse:
+            # whether or not to hide mouse when typing
```
christianparpart, line 225:
```suggestion
            # whether or not to hide mouse when typing
            #
            # Default value: true
```
I think it's worth mentioning the default (or simply just to stay consistent with most of the options) :)

christianparpart (commented):
I think such an option is worth adding to the changelog, as well. :)

Yaraslaut:
> I think such an option is worth adding to the changelog, as well. :)

Done

christianparpart (approved):
Many thanks, @Yaraslaut :)

PR #999, opened 2023-01-22
WIP

Utkarsh-khambra:
handles Tabstops.
Previously  tried to track tab character by storing starting tab positions in `Line` but soon abandoned that since in a lot of places we directly handle cells then those places needed to changed as well. So right now tracking tabs by setting a flag(bool) for each cell.


christianparpart:
I think it's actually harder than it looks. How's VTE lib doing that? I mean, writing `A\tB` will write `A` to (say) column 1 and `B` to column 9 as the `\t` will move the cursor forward to the next tab stop (default modulo 8). So far so good.
Which cell does not remember the tabstop? Cell at column 2, or all cells in between? 🤔 I think that's hard to say, but maybe all spacing cells in between should get that flag (or only that at which you've been writing the TAB, column 2).
Now however, what if the cursor is placed in between A and B. say, column 4, to write a C?
The only think I can make sense here is to destroy the remembered TAB information and make it look like simple spaces (or empty cells), dropping the TAB flag, then writing the C at column 4. So now we'd have `A  C     B` which used to be `A\tB` if reconstructed from the line.

Another note, a bool takes 8 bits, probably with padding. maybe we can make it part of `CellFlags` that is stored in `CellExtra` for the `CompactCell` case.

Also, the `CellConcept` must reflect the API change in the Cell (if we stick with this). You could probably simplify here if moving that information to `CellFlags`.

Utkarsh-khambra:
> Which cell does not remember the tabstop? Cell at column 2, or all cells in between? thinking I think that's hard to say, but maybe all spacing cells in between should get that flag (or only that at which you've been writing the TAB, column 2).

Currently I am doing all cells in between get the flag, we can also do that only first one gets the flag but then to check which others are also covered by tab we need to do calculations every time.

> Now however, what if the cursor is placed in between A and B. say, column 4, to write a C?
The only think I can make sense here is to destroy the remembered TAB information and make it look like simple spaces (or empty cells), dropping the TAB flag, then writing the C at column 4. So now we'd have A  C     B which used to be A\tB if reconstructed from the line.

In gnome terminal if I do `printf 'A\tB\e[4DC\n'` it converts it into `A    B\tC`, maybe we can also do the same, I don't know what the actual hardware used to do for this case. 

> Another note, a bool takes 8 bits, probably with padding. maybe we can make it part of CellFlags that is stored in CellExtra for the CompactCell case.

Yep I'll do that.

Utkarsh-khambra:
Todo
[ ] handle tabs when in vim mode.
[ ] handle when a character is inserted between a tab

PR #1001, opened 2023-01-23
Adds shell-integration for fish.

Utkarsh-khambra:
Adds shell integration for fish. I moved shell integration files to their own directory because in future I also want to add completions for fish, so keeping a different folder will help organization.


metainfo.xml
```diff
@@ -138,6 +138,7 @@
           <li>Improvements to vi-like normal mode: Add `M` motion to jump to middle screen line (same column).</li>
           <li>Improvements to vi-like normal mode: Add `P` to paste the clipboard with newlines stripped.</li>
           <li>Improvements to vi-like normal mode: Add `SP` and `BS` to move left/right.</li>
+          <li>Adds shell integration for fish shell.</li>
```
christianparpart, line 141: I think this can go up into the next release section now :)

src/contour/shell-integration/shell-integration.fish
```diff
@@ -0,0 +1,46 @@
...
+#function update_profile()
+#    switch "$PWD"
+#        case "$HOME"/work* 
+#            contour set profile to work
+#        case "$HOME"/projects* 
+#            contour set profile to main
+#        case '*' 
+#            contour set profile to mobile
+#    end
+#end
```
christianparpart, line 27: I think this is POSIX syntax here, right? then we should probably get rid of it or update to be fish syntax.

src/contour/shell-integration/shell-integration.fish
```diff
@@ -0,0 +1,46 @@
...
+
+
+function precmd_hook_contour -d "Run before printing prompt" -e fish_prompt
```
christianparpart, line 30:
```suggestion
function precmd_hook_contour -d "Shell Integration hook to be invoked before each prompt" -e fish_prompt 
```

I think that's a better one - to describe what this function is there for.

christianparpart (changes requested):
only minors. many thanks :-)

p.s.: `git fetch && git rebase origin/master`

christianparpart (approved):

PR #1030, opened 2023-02-14
Remove apt update in ci

Yaraslaut:
Github updates its linux images from time to time and there is no need to update all of then manually


christianparpart:
Thanks :-)

whisperity:
This did not appropriately fix the CodeChecker CI job, nor the release job.

PR #1051, opened 2023-02-25
fix: take into account the status line height when rendering sixels.

### LINKED ISSUE #1050: Sixel graphics not rendered while intersects with status line
### Contour Terminal version

Contour Terminal Emulator 0.3.11-qml-take-2-150427be

### Installer source

Github: source code cloned

### Operating System

Arch

### Architecture

x86-64

### Other Software

_No response_

### Steps to reproduce

https://user-images.githubusercontent.com/44506630/221343173-df896c72-aef0-42fb-a8b1-c7c0ed5daaff.mp4

### Expected Behavior

_No response_

### Actual Behavior

part of the image is not displayed while intersect with status_line

### Additional notes

[lines](https://user-images.githubusercontent.com/44506630/221343195-ebd0f902-9237-4951-a132-acc0c58517d8.png)

Utkarsh-khambra:
When sixel image is being rendered, if status line is visible then total renderable area should not cover the statusline area.
Fixes #1050


metainfo.xml
```diff
@@ -122,6 +122,7 @@
           <li>Adds shell integration for tcsh shell.</li>
           <li>Improve performance of parsing UTF-8 text on ARM64 platfroms through the use SIMD instructions.</li>
           <li>Contour can now run on platforms not supporting hardware crypto extension for ARM64 nor AES-NI for x86-64. Hardware acceleration support can be configured to be included at compile time.</li>
+          <li>Fixes rendering sixel image when status line is active (#1050).</li>
```
christianparpart, line 125: Can you move this up to the bottom of the other "Fixes"-lines, to keep them grouped? :)

src/vtbackend/Screen.cpp
```diff
@@ -1852,7 +1852,8 @@ void Screen<Cell>::renderImage(shared_ptr<Image const> image,
     // TODO: make use of imageOffset
     (void) imageOffset;
 
-    auto const linesAvailable = _settings.pageSize.lines - topLeft.line.as<LineCount>();
+    auto const linesAvailable = _settings.pageSize.lines - topLeft.line.as<LineCount>()
+                                - (_terminal.state().statusDisplayType == StatusDisplayType::None ? 0 : 1);
```
christianparpart, line 1856: I think this should be a little bit more future-proof, because the line height of the statusline does not necessarily have to be 1 all the time.

You'd need to fetch the value from `_indicatorStatusScreen` and `_hostWritableStatusLineScreen`'s `pageSize().lines`. They do not necessarily need to match either.

Currently that's `1` yes, but it may and most likely will change in the future. Imagine using the statusline to also display multiline error messages (not done yet, but likely in the future), that would be a value `> 1` then :).

src/vtbackend/Screen.cpp
```diff
@@ -1901,7 +1902,9 @@ void Screen<Cell>::renderImage(shared_ptr<Image const> image,
                 auto const offset =
                     CellLocation { boxed_cast<LineOffset>(linesToBeRendered) + lineOffset, columnOffset };
                 Cell& cell =
-                    at(boxed_cast<LineOffset>(_settings.pageSize.lines) - 1, topLeft.column + columnOffset);
+                    at(boxed_cast<LineOffset>(_settings.pageSize.lines) - 1
+                           - (_terminal.state().statusDisplayType == StatusDisplayType::None ? 0 : 1),
```
christianparpart, line 1906: same here. Maybe we could use the same code path here to lower the risk of bugs due to code duplications.

christianparpart (changes requested):

src/vtbackend/Terminal.h
```diff
@@ -226,6 +226,18 @@ class Terminal
 
     [[nodiscard]] PageSize totalPageSize() const noexcept { return _settings.pageSize; }
 
+    // Returns status line height
```
christianparpart, line 229:
```suggestion
    // Returns number of lines for the currently displayed status line,
    // or 0 if status line is currently not displayed.
```

src/vtbackend/Terminal.h
```diff
@@ -226,6 +226,18 @@ class Terminal
...
+    [[nodiscard]] LineCount statusLineHeight() const noexcept
+    {
+        switch (_state.statusDisplayType)
+        {
+            case StatusDisplayType::None: return LineCount(0);
+            case StatusDisplayType::Indicator: return _indicatorStatusScreen.pageSize().lines;
+            case StatusDisplayType::HostWritable: return _hostWritableStatusLineScreen.pageSize().lines;
+        }
+        crispy::unreachable();
```
christianparpart, line 238: very nice! 😄

christianparpart (approved):

PR #1055, opened 2023-02-27
Update install-deps.sh script

Yaraslaut:
On newly installed ubuntu running `./scripts/install-deps.sh ` cause error 
` ./scripts/install-deps.sh : 47 curl not found` 
Also, while we are installing `ninja` on all platform `ubuntu` and `pop os` were missing


christianparpart (dismissed):
nice. thanks. :)

PR #1061, opened 2023-03-07
Adds `profiles.*.permissions.display_host_writable_statusline` to allow the user to intervene in `DECSSDT`

christianparpart:
Refs #1060.

### DIFF
--- a/metainfo.xml
+++ b/metainfo.xml
@@ -107,6 +107,7 @@
       <description>
         <ul>
           <li>Fixes terminfo `tsl` to how todays apps are using it (e.g. zsh) to simply set the window title.</li>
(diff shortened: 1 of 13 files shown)


whisperity:
Sounds good, doesn't work, unfortunately. I've just upgraded to [0.3.12.4890](https://github.com/contour-terminal/contour/actions/runs/4358651794) and if I have `TERM=contour` or `TERM=contour-latest` then spawning ZSH will get this ugly status bar appear, even though I have set the permission flag to `deny`.

Conversely, having it set to `ask` (the default) while using **the default configuration** and spawning in to `bash` then going `TERM=contour-latest zsh` will get it to appear **without** any confirmation.

Having, in all of my profiles, disabled some sort of (?) statusline also does not make this thing go away...

```yaml
   status_line:
      display: none
```

Issue #715, opened 2022-06-04
[error] Shader log: Fragment info on start


GoldsteinE:
### Contour Terminal version

Contour Terminal Emulator 0.3.2-unreleased

### Installer source

Github: source code cloned

### Operating System

NixOS

### Architecture

x86-64

### Steps to reproduce

I’m not sure if this is reproducible, so I’ll dump some additional information here.

I use nvidia with proprietary drivers (version 515.48.07) and contour built from commit 6d706d066534514e9af050d9bc31d88850bfdd24. My WM is i3 without a compositor. Settings that seem render-relevant are set to default values.

The issue happens on terminal start.

### Expected Behavior

contour starts with no errors.

### Actual Behavior

contour shows the following error on start:

```
[error] Shader log: Fragment info
        -------------
        0(30) : warning C7012: not unrolling loop that executes 16384 times since maximum loop unroll count is 4096
        0(30) : warning C7012: not unrolling loop that executes 16384 times since maximum loop unroll count is 4096
```

<details><summary>Full log up to error message (with debug gui.display,gui.input,gui.session,error )</summary>

```
[2022-06-05 00:34:53.204408] [gui.session] Configuring terminal.
[2022-06-05 00:34:53.204448] [gui.session] Setting terminal ID to VT525.
[2022-06-05 00:34:53.204454] [gui.session] maxImageSize=3840x2160, sixelScrolling=true
[2022-06-05 00:34:53.204490] [gui.display] contentScale: 1
[2022-06-05 00:34:53.204499] [gui.display] contentScale: 1
[2022-06-05 00:34:53.232039] [gui.display] contentScale: 1
[2022-06-05 00:34:53.232059] [gui.session] Assigning display.
[2022-06-05 00:34:53.232093] [gui.display] contentScale: 1
[2022-06-05 00:34:53.232099] [gui.display] sizeHint: 1920x1500 (1920x1500)
[2022-06-05 00:34:53.232246] [gui.session] Starting main loop with thread id 140655279994432
[2022-06-05 00:34:53.243269] [gui.display] contentScale: 1
[2022-06-05 00:34:53.243283] [gui.display] sizeHint: 1920x1500 (1920x1500)
[2022-06-05 00:34:53.243344] [gui.display] TerminalWindow.resizeEvent: size 1962x1500 (1962x1500)
[2022-06-05 00:34:53.243353] [gui.display] called with 1962x1500 in Primary
[...]

PR #1071, opened 2023-04-04
Create separate category for shader log

Yaraslaut:
Closes https://github.com/contour-terminal/contour/issues/715

### DIFF
--- a/src/contour/display/ShaderConfig.cpp
+++ b/src/contour/display/ShaderConfig.cpp
@@ -31,7 +31,7 @@ using namespace std::string_literals;
 namespace contour::display
 {
 
-auto const ShaderLog = logstore::Category("shader", "Logs shader configuration");
+auto const ShaderLog = logstore::Category("gui.shader", "Logs shader configuration");
 
 namespace
 {
--- a/src/contour/display/ShaderConfig.cpp
+++ b/src/contour/display/ShaderConfig.cpp
@@ -137,7 +137,7 @@ std::unique_ptr<QOpenGLShaderProgram> createShader(ShaderConfig const& _shaderCo
     }
 
     if (auto const logString = shader->log().toStdString(); !logString.empty())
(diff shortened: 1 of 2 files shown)


src/contour/display/ShaderConfig.cpp
```diff
@@ -135,7 +137,7 @@ std::unique_ptr<QOpenGLShaderProgram> createShader(ShaderConfig const& _shaderCo
     }
 
     if (auto const logString = shader->log().toStdString(); !logString.empty())
-        errorlog()("Shader log: {}", logString);
+        ShaderLog()("Shader log: {}", logString);
```
christianparpart, line 140:
```suggestion
        ShaderLog()("{}", logString);
```

Because the logger will already prefix it with the shader log's tag. :)

christianparpart (changes requested):

src/contour/display/ShaderConfig.cpp
christianparpart: This may be worth mentioning in the `metainfo.xml`'s changelog section. What do you think?

Yaraslaut: > Because the logger will already prefix it with the shader log's tag. :)

Oh, right, didn't notice

src/contour/display/ShaderConfig.cpp
```diff
@@ -31,6 +31,8 @@ using namespace std::string_literals;
 namespace contour::display
 {
 
+auto const ShaderLog = logstore::Category("shader", "Logs shader configuration");
```
christianparpart, line 34:
```suggestion
auto const ShaderLog = logstore::Category("gui.shader", "Logs shader configuration");
```

All the GUI code has (IIRC) all logging tags prefixed with `gui.`

christianparpart (approved):

PR #1083, opened 2023-04-14
Qt6 build on windows and CONTOUR_QT_VERSION cmake variable 

### LINKED ISSUE #1082: Building on Windows 11 with Qt6
### Contour Terminal version

latest master

### Installer source

Github: source code cloned

### Operating System

Windows 11 Enterprise 22H2 22621.1413

### Architecture

x86-64

### Other Software

_No response_

### Steps to reproduce

Try building from source (took the liberty to sort of document the process as well):

- Clone vcpkg repository somewhere (preferably high up the tree, i.e. to `C:\vcpkg` and add C:\vcpkg to PATH.
- Install Visual Studio Build Tools (in components manually select CLI tool options for C++).
- Open the  _developer_ version of Powershell.
- Clone contour and run `.\scripts\install-deps.ps1`, prepare to wait ages.
- Install Qt6 (i.e. to C:\Qt)
- run ` cmake -S . -B build -DCMAKE_TOOLCHAIN_FILE=C:\vcpkg\scripts\buildsystems\vcpkg.cmake -DCMAKE_PREFIX_PATH=C:\Qt\6.5.0\msvc2019_64\lib\cmake`

Observe error message:
```
  Could not find a package configuration file provided by "Qt5Core" with any
  of the following names:

    Qt5CoreConfig.cmake
    qt5core-config.cmake
```

### Expected Behavior

Contour builds with Qt6.

### Actual Behavior

Contour complains about missing Qt5.

### Additional notes

Adding `-D CONTOUR_BUILD_WITH_QT6=ON` does not change behaviour.,

Yaraslaut:
Fix Qt6 build on windows,
at the moment `Release` build is working, while `Debug` is not
Closes #1082


src/contour/CMakeLists.txt
```diff
@@ -202,8 +202,11 @@ if(WIN32)
     set(CPACK_PACKAGE_ICON "${CMAKE_CURRENT_SOURCE_DIR}/res/images/contour-logo.ico")
     set(CPACK_WIX_UPGRADE_GUID "0E736497-2B72-4117-95E9-54EC6DC2432A")
 
-    include(DeployQt)
-    windeployqt(contour)
+    if(CONTOUR_BUILD_WITH_QT6)
```
christianparpart, line 205: I think we still need that for Qt6 builds, but inside `DeployQt.cmake`, the qt version is still hardcoded to Qt 5. That should somehow make use of `CONTOUR_BUILD_WITH_QT6` to mandate what to use. 

Maybe we should even move away from `CONTOUR_BUILD_WITH_QT6` and introduce something like `USE_QTVERSION_MAJOR` cmake-var that can be either `5` or `6` for easier use of it. 🤔

Yaraslaut: Indeed, win deploy qt for qt6 does not exist, what we need is to manually execute windeployqt.exe
I will check what we need to  do here

christianparpart: image
I think what I did on windows, is simply (but sadly manually) changing this value from 5 to 6. That should not be done automatically of course. Am I being mislead? 🤔

Yaraslaut: > image
> I think what I did on windows, is simply (but sadly manually) changing this value from 5 to 6. That should not be done automatically of course. Am I being mislead? thinking

oh, okay, i think we can easilty adjust cmake file to use inside contour, and then `USE_QTVERSION_MAJOR` can be introduced

.github/fedora/contour.spec
```diff
@@ -50,7 +50,7 @@ for everyday use. It is aiming for power users with a modern feature mindset.
 cmake . \
     -DCMAKE_BUILD_TYPE=RelWithDebInfo \
     -DCMAKE_INSTALL_PREFIX=/usr \
-    -DCONTOUR_BUILD_WITH_QT6=ON \
+    -DCONTOUR_QT_VERSION=6 \
```
christianparpart, line 53: Yes, nice, I remember we talked about this. many thanks!

christianparpart (approved):
Many thanks

PR #1086, opened 2023-04-15
Use docker for Ubuntu 18.04 packaging

Yaraslaut:


christianparpart:
For as long as @whisperity needs Ubuntu 18.04 at work, we need to figure out a way to build binaries on Ubuntu 18.04

What could work, is, adding a new action that builds Ubuntu images through docker. That way we can support building binaries that are not directly supported by GitHub CI runners.

Yaraslaut:
@christianparpart  Please squash if no modifications is required to merge

.github/ubuntu18/Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+    software-properties-common \
+    libqt5gui5 \
+    libqt5opengl5-dev \
+    libqt5x11extras5-dev \
+    qtbase5-dev \
+    qtmultimedia5-dev
+
+RUN add-apt-repository ppa:ubuntu-toolchain-r/test
+RUN apt-get update
+RUN apt-get install -y g++-10
```
christianparpart, line 34: I think we can drop this line, meaning, a default compiler should be pulled in as dependency via `install-scripts.sh`.

.github/workflows/build.yml
```diff
@@ -78,6 +78,69 @@ jobs:
...
+          docker buildx build --platform ${{ steps.tags.outputs.PLATFORM }} \
+          --tag contour:${{ steps.tags.outputs.ARCH }} \
+          -f .github/ubuntu18/Dockerfile \
+          --load \
+          .
+          docker create --name contour-${{ steps.tags.outputs.ARCH }} contour:${{ steps.tags.outputs.ARCH }}
+          docker cp contour-${{ steps.tags.outputs.ARCH }}:/home/docker/contour/package.deb \
+          "contour-${{ steps.set_vars.outputs.VERSION_STRING }}-ubuntu18.04-amd64.deb"
+          docker cp contour-${{ steps.tags.outputs.ARCH }}:/home/docker/contour/package.ddeb \
+          "contour-dbgsym-${{ steps.set_vars.outputs.VERSION_STRING }}-ubuntu18.04-amd64.ddeb"
```
christianparpart, line 127: Veeery nice. I was afraid that this would be more complicated to get them out of there. :)

christianparpart (commented):
Sad thing is, the `release.yaml` needs to be adapted to.

But overall it looks very good. Many many thanks for going through this. :)

.github/ubuntu18/Dockerfile
Yaraslaut: Default compiler is g++-8 and to install g++-10 we need to add-apt-repository, i think it's better to have this here than update install-deps.sh script to handle docker image of ubuntu 18.04

christianparpart: Ok. Understood. :)

christianparpart (approved):
Many thanks! :)

Issue #964, opened 2023-01-01
Contour crashes when Lohit Devanagari font is used.


Utkarsh-khambra:
### Contour Terminal version

Contour Terminal Emulator 0.3.10-master-f2a0c3d5

### Installer source

Github: source code cloned

### Operating System

Fedora 37

### Architecture

x86-64

### Steps to reproduce

Set fonts to `Lohit Devanagari`

### Expected Behavior

Contour starts up fine.

### Actual Behavior

Contour crashes. and this is the backtrace.
```
/usr/include/c++/12/bits/stl_vector.h:1123: constexpr std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](size_type) [with _Tp = unsigned char; _Alloc = std::allocator<unsigned char>; reference = unsigned char&; size_type = long unsigned int]: Assertion '__n < this->size()' failed.

Thread 1 "contour" received signal SIGABRT, Aborted.
__pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at pthread_kill.c:44
44      return INTERNAL_SYSCALL_ERROR_P (ret) ? INTERNAL_SYSCALL_ERRNO (ret) : 0;
(gdb) bt
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at pthread_kill.c:44
#1  0x00007ffff5aafee3 in __pthread_kill_internal (signo=6, threadid=<optimized out>) at pthread_kill.c:78
#2  0x00007ffff5a5faa6 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007ffff5a497fc in __GI_abort () at abort.c:79
#4  0x00007ffff5cd7e80 in std::__glibcxx_assert_fail(char const*, int, char const*, char const*)
    (file=file@entry=0x6b01e8 "/usr/include/c++/12/bits/stl_vector.h", line=line@entry=1123, function=function@entry=0x6e94b8 "constexpr std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](size_type) [with _Tp = unsigned char; _Alloc = std::allocator<unsigned char>; reference = unsigned char&; size_type ="..., condition=condition@entry=0x6b0543 "__n < this->size()") at ../../../../../libstdc++-v3/src/c++11/debug.cc:60
[...]

Issue #1065, opened 2023-03-16
Memory corruption crashes when using larger font sizes with Shalassian


whisperity:
### Contour Terminal version

0.3.10.4591

### Installer source

Github: CI actions

### Operating System

Ubuntu 20.04

### Architecture

x86-64

### Steps to reproduce

1. Download [Shalassian](https://feyawen.carrd.co/) and set it as the regular font family, size 12.
2. Start Contour and do <kbd>Ctrl</kbd>-<kbd>MWheelUp</kbd> to increase the font size **4 times**.
3. Contour crashes.

### Expected Behavior

No crash.

### Actual Behavior

Crash, with no explanation as to what is happening. Sometimes `munmap_chunk(): Invalid pointer.`, sometimes `free(): Invalid next size (fast)`, sometimes `double free or corruption (out)`.

```
[2023-03-16 10:17:45.272049] [vt.renderer] Updating grid metrics: (pageSize=67x21, cellSize=13x16, baseline=2, underline=1@1, margin=(left=0, bottom=0))
[2023-03-16 10:17:45.272136] [font.locator] Calculating font metrics for (family=Shalassian Font weight=Bold slant=Italic spacing=Monospace, strict_spacing=no): (17, 14, 15, 0, -2, 1)
[2023-03-16 10:17:45.272140] [vt.renderer] Loading grid metrics (pageSize=67x21, cellSize=14x17, baseline=2, underline=0@1, margin=(left=0, bottom=0))
[2023-03-16 10:17:45.272310] [gui.display] configureAtlas: 2048x2048 RGBA
[2023-03-16 10:17:45.272367] [vt.renderer] Configuring texture atlas.
[2023-03-16 10:17:45.272372] [vt.renderer] - Atlas properties     : tile size 14x17, format RGBA, direct-mapped 116
[2023-03-16 10:17:45.272375] [vt.renderer] - Atlas texture size   : 2048x2048 pixels
[2023-03-16 10:17:45.272378] [vt.renderer] - Atlas hashtable      : 4096 slots
[2023-03-16 10:17:45.272381] [vt.renderer] - Atlas tile count     : 17520 = 146x * 120y
[2023-03-16 10:17:45.272384] [vt.renderer] - Atlas direct mapping : enabled (for text rendering)
[...]

PR #1087, opened 2023-04-18
Fix `int` to `unsigned` overflow conversion

Yaraslaut:
Fixes https://github.com/contour-terminal/contour/issues/964 and https://github.com/contour-terminal/contour/issues/1065

### DIFF
--- a/src/vtrasterizer/DecorationRenderer.cpp
+++ b/src/vtrasterizer/DecorationRenderer.cpp
@@ -153,7 +153,7 @@ auto DecorationRenderer::createTileData(Decorator decoration, atlas::TileLocatio
         case Decorator::Underline: {
             auto const thickness_half = max(1u, unsigned(ceil(underlineThickness() / 2.0)));
             auto const thickness = thickness_half * 2;
-            auto const y0 = max(0u, unsigned(underlinePosition()) - thickness_half);
+            auto const y0 = max(0, (underlinePosition() - static_cast<int>(thickness_half)));
             auto const height = Height(y0 + thickness);
             auto const imageSize = ImageSize { width, height };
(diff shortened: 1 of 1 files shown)


src/vtrasterizer/DecorationRenderer.cpp
```diff
@@ -153,7 +153,7 @@ auto DecorationRenderer::createTileData(Decorator decoration, atlas::TileLocatio
         case Decorator::Underline: {
             auto const thickness_half = max(1u, unsigned(ceil(underlineThickness() / 2.0)));
             auto const thickness = thickness_half * 2;
-            auto const y0 = max(0u, unsigned(underlinePosition()) - thickness_half);
+            auto const y0 = max(0, (underlinePosition() - static_cast<int>(thickness_half)));
```
whisperity, line 156: I believe this is the part where we can safely decide to stop overusing `auto` for everything and explicitly mentioning the expected type here for `y0` in the declaration itself...

whisperity:
#1065 is indeed fixed by this patch. Comment about the code style still stands.

Utkarsh-khambra:
Fixes #964 too.

src/vtrasterizer/DecorationRenderer.cpp
Yaraslaut: > I believe this is the part where we can safely decide to stop overusing `auto` for everything and explicitly mentioning the expected type here for `y0` in the declaration itself...

We want to have non-negative value in y0, so unsigned is the minimal type that we can use for y0, while for ADL for max function we want to have int as argument, so something like `max<int>(a,b)`  can be used to get proper value

christianparpart (approved):
Many thanks (especially for the patience!)

PR #1089, opened 2023-04-22
If there is no URI with the given ID in cache then add one with the g…

### LINKED ISSUE #1088: Contour doesn't recognize an `OSC 8` link if it has an id in it.
### Contour Terminal version

0.3.12-master-bedec7ae

### Installer source

Github: source code cloned

### Operating System

Ubuntu 22.04

### Architecture

x86-64

### Other Software

_No response_

### Steps to reproduce

printf '\e]8;id=colorize_1;http://example.com\e\\This is a link\e]8;;\e\\\n'

### Expected Behavior

See a clickable link.

### Actual Behavior

Link is not clickable.

### Additional notes

_No response_

Utkarsh-khambra:
Fixes #1088


christianparpart (approved):
good catch. many thanks.

Issue #1084, opened 2023-04-14
Hovering mouse on an `OSC 8` hyperlink when scolled up forces scroll down


whisperity:
### Contour Terminal version

0.3.12.4932

### Installer source

Github: CI actions

### Operating System

Ubuntu 22.04

### Architecture

x86-64

### Other Software

bash 5.1.16(1)-release (x86_64-pc-linux-gnu)
zsh 5.8.1 (x86_64-ubuntu-linux-gnu)

### Steps to reproduce

1. Execute: `clear; printf '\e]8;;http://example.com\e\\This is a link!\e]8;;\e\\\n'; seq 1 100`
2. Press <kbd>Shift</kbd>-<kbd>Home</kbd> to scroll to top (or scroll manually)
3. Hover over `This is a link!`

### Expected Behavior

The hovering is decorated and then I can execute <kbd>Ctrl</kbd>-<kbd>Mouse1</kbd> to open it?

### Actual Behavior

The viewport jumps down to the bottom, making the link inaccessible!

### Additional notes

Happens with both my customised but also with a default-created configuration.

christianparpart:
that fix can come easily with a unit test to cover it :)

whisperity:
Ubuntu 20.04 & Contour 0.3.10.4591 (installed from GHA as usual) does **not** exhibit bogus behaviour. This should narrow the bisection interval a little bit.

christianparpart:
I think we can closes this now, with #1090 being merged. :)

PR #1090, opened 2023-04-22
When updating hovering state, only renderBuffer should be redrawn.

Utkarsh-khambra:
Fixes (#1084)
@christianparpart I am not sure, how to write unit test to cover this case.


src/vtbackend/Terminal.cpp
```diff
@@ -1040,7 +1040,7 @@ void Terminal::updateHoveringHyperlinkState()
     auto const oldState = _hoveringHyperlinkId.exchange(newState);
 
     if (newState != oldState)
-        screenUpdated();
+        renderBufferUpdated();
```
Utkarsh-khambra, line 1043: I am just updating render buffer here, is there any need to flush input on hovering state update event?

christianparpart: No, there isn't. I am however not sure if there are any other implications to this. I do not think so right out of my head, however. 🤔

christianparpart (approved):
Looks good to me. Thanks.

christianparpart:
@christianparpart I am not sure, how to write unit test to cover this case.

Me neither, for now. I mean, sure, but that would probably be way to complex. We may need a better testing infrastructure to trivially test visual expectations (including graphical attributes). I had the intend to do that multiple/many times already, but never came up with a good solution for it just yet.

christianparpart (approved):

PR #1092, opened 2023-05-01
Use default shell if can not spawn user defined

### LINKED ISSUE #519: [Windows] Contour silently closes if the configured shell cannot be started
## Description

If the `shell` configuration variable in the profile refer to a shell that cannot be started, Contour just closes immediately without opening a window.

## Environment

- Contour Version or Git commit hash: https://github.com/contour-terminal/contour/releases/tag/v0.2.2.175
- Operating System (name / distribution / version): Windows 10 20.04

## Steps to Reproduce

 1. Use your Linux-y Contour profile on Windows which says `/bin/zsh` is your default shell.
 2. Double-click `contour.exe`
 3. The cursor changes to the "semi-busy" icon, but then goes back to normal immediately. The spawned process exits. No error message is generated.

Yaraslaut:
Closes  #519

### DIFF
--- a/src/vtpty/Process_unix.cpp
+++ b/src/vtpty/Process_unix.cpp
@@ -239,9 +239,8 @@ void Process::start()
             // Fallback: Try login shell.
             auto theLoginShell = loginShell(_d->escapeSandbox);
             fprintf(stdout,
-                    "\r\n\033[31;1mFailed to spawn \"%s\". %s\033[m\r\nTrying login shell: %s\n",
+                    "\r\n\033[31;1mFailed to spawn \%s\033[m\r\nTrying login shell: %s\n",
                     argv[0],
-                    strerror(errno),
                     crispy::joinHumanReadableQuoted(theLoginShell, ' ').c_str());
             fflush(stdout);
             if (!theLoginShell.empty())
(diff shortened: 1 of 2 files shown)


src/vtpty/Process_unix.cpp
```diff
@@ -239,9 +239,8 @@ void Process::start()
             // Fallback: Try login shell.
             auto theLoginShell = loginShell(_d->escapeSandbox);
             fprintf(stdout,
-                    "\r\n\033[31;1mFailed to spawn \"%s\". %s\033[m\r\nTrying login shell: %s\n",
+                    "\r\n\033[31;1mFailed to spawn \%s\033[m\r\nTrying login shell: %s\n",
```
christianparpart, line 242: good catch!

christianparpart (approved):
Looks good to me. I assume you also tested this on Windows. :)

Many thanks!

src/vtpty/Process_unix.cpp
christianparpart, line 242:
```suggestion
                    "\r\n\033[31;1mFailed to spawn %s\033[m\r\nTrying login shell: %s\n",
```

christianparpart: This is why the linux builds are failing.

christianparpart (changes requested):
FYI

PR #1103, opened 2023-05-25
Properly handle dependency installation on Fedora 38+

christianparpart:


Yaraslaut:
I think at some point we need to move to catch v3 anyway, here some info https://github.com/catchorg/Catch2/blob/devel/docs/migrate-v2-to-v3.md

christianparpart:
> I think at some point we need to move to catch v3 anyway, here some info https://github.com/catchorg/Catch2/blob/devel/docs/migrate-v2-to-v3.md

We can give that a try in its own PR and see if that works on all supported platforms (Windows, OS/X, Fedora, Arch, Ubuntu... down to 18.04) Then I guess we simply invert it, meaning, that for cases like Ubuntu 18.04 we'd simply go for embedding Catch2 version 3 then. I'm okay with that. :)

PR #1117, opened 2023-06-07
Invert direction for downwards glyphs

### LINKED ISSUE #1115: Glyphs overlap and crash
### Contour Terminal version

Contour Terminal Emulator 0.3.13-master-12703311

### Installer source

Github: source code cloned

### Operating System

Arch

### Architecture

x86-64

### Other Software

_No response_

### Steps to reproduce

code that creates this issue 
```
use std::{thread, time::Duration};


fn print_u32(cha:u32)
{
    print!(
        "{}",
        match std::char::from_u32(cha) {
            Some(c) => c,
            None => '?',
        }
    );
}

fn print_u32_with_index(cha:u32)
{
    println!(
        "{}: {}",
        cha,
        match std::char::from_u32(cha) {
            Some(c) => c,
            None => '?',
        }
    )
}

fn wait(mils:u64)
{
    thread::sleep(Duration::from_millis(mils));
}

fn output_cha(cha_init:u32, cha_size:u32, print: fn(u32))
{
    let mut cha : u32 = cha_init;
    let mut counter = 0;
    while  cha < cha_init + cha_size
    {
        cha +=1;
        counter +=1;
        print(cha);

        if counter == 100
        {
            counter = 0;
            println!("");
            println!("{}", cha as f32 / (cha_size + cha_init) as f32);
            wait(100);
        }
    }
}


fn main() {
    println!("Output of Plane 1 unicode");
    let pr = print_u32; // print_u32_with_index;
    output_cha(0,55634,pr);
    wait(1000);

    println!("Output of Plane 2 unicode");
    output_cha(55634,23276,pr);
    wait(1000);

    println!("");
    println!("Output of Plane 3 unicode");
    output_cha(55634 + 23276, 60873,pr);
    wait(1000);
}

```

### Expected Behavior

_No response_

### Actual Behavior

image

some values  and assert 
```
_gridMetrics.baseline: 4
glyph.position.y: -11
yMax: -7
yMin: -16
bitmapsize.height: 9
[/home/yaraslau/repo/contour/src/vtrasterizer/TextRenderer.cpp:863] Precondition failed. rowCount <= unbox<unsigned>(glyph.bitmapSize.height)

```

### Additional notes
Example of alacritty 
image

Yaraslaut:
Closes #1115  
See https://freetype.org/freetype2/docs/reference/ft2-base_interface.html#ft_glyphslotrec
`bitmap_top` can  be negative, meaning that glyph direction is downward, this PR fixes the crash when contour can not display such glyphs, this is itself a hack since this is not how properly handle this case i suppose.
For example such figures still can be generated in contour 
image

### DIFF
--- a/src/vtrasterizer/TextRenderer.cpp
+++ b/src/vtrasterizer/TextRenderer.cpp
@@ -837,7 +837,13 @@ auto TextRenderer::createRasterizedGlyph(atlas::TileLocation tileLocation,
     }
 
     // y-position relative to cell-bottom of glyphs top.
-    auto const yMax = _gridMetrics.baseline + glyph.position.y;
+    auto yMax = _gridMetrics.baseline + glyph.position.y;
+
+    if(yMax < 0)
+    {
+        RasterizerLog()("Encountered glyph with inverted direction, swaping to normal");
+        yMax *= -1;
+    }
 
     // y-position relative to cell-bottom of the glyphs bottom.
     auto const yMin = yMax - glyph.bitmapSize.height.as<int>();
--- a/src/vtrasterizer/TextRenderer.cpp
+++ b/src/vtrasterizer/TextRenderer.cpp
@@ -839,10 +839,10 @@ auto TextRenderer::createRasterizedGlyph(atlas::TileLocation tileLocation,
     // y-position relative to cell-bottom of glyphs top.
     auto yMax = _gridMetrics.baseline + glyph.position.y;
 
-    if(yMax < 0)
+    if (yMax < 0)
     {
         RasterizerLog()("Encountered glyph with inverted direction, swaping to normal");
-        yMax *= -1;
+        yMax = std::abs(yMax);
     }
 
     // y-position relative to cell-bottom of the glyphs bottom.
--- a/metainfo.xml
+++ b/metainfo.xml
@@ -108,6 +108,7 @@
         <ul>
           <li>Do not clear search term when entering search editor again.</li>
           <li>Fixes search term rendering highlighting for search terms containing whitespaces (#966).</li>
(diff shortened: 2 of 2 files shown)


src/vtrasterizer/TextRenderer.cpp
```diff
@@ -837,7 +837,13 @@ auto TextRenderer::createRasterizedGlyph(atlas::TileLocation tileLocation,
...
 
     // y-position relative to cell-bottom of glyphs top.
-    auto const yMax = _gridMetrics.baseline + glyph.position.y;
+    auto yMax = _gridMetrics.baseline + glyph.position.y;
+
+    if(yMax < 0)
+    {
+        RasterizerLog()("Encountered glyph with inverted direction, swaping to normal");
+        yMax *= -1;
+    }
```
christianparpart, line 846:
```suggestion
    if (yMax < 0)
    {
        RasterizerLog()("Encountered glyph with inverted direction, swaping to normal");
        yMax = std::abs(yMax);
    }
```
While I am perfectly fine with inverting yMax by multiplying with `-1` I feel that using `|x|` instead is more readable, `abs(x)` I mean :)

christianparpart: Also, as you also said this is a hack, I am not 100% certain that this is how it is supposed to be fixed. I mean, if it works, i'm happy, actually. But I didn't have the time yet to actually understand this change more deeply.

christianparpart: NB: that is probably worth mentioning in the changelog section in `metainfo.xml`. Some user-facing text describing this fix.

christianparpart (changes requested):
Looks good to me, but I also didn't have too much time deeply understanding the root of this issue.

christianparpart (approved):
thanks :)

PR #1120, opened 2023-06-14
Deploy docs only from master branch commits

Yaraslaut:
We need to update  web page only when we make commits in master, otherwise if someone creates a pull request into master branch this person can update our web page


christianparpart (approved):
thanks

Issue #1063, opened 2023-03-11
Implement inheritance of config trees (without YAML anchors...)


whisperity:
Because YAML anchors do not allow specialising a child node of an anchored-in tree, we need to instead roll our own joint of merging user-specific changes via inheritance.

## Use case

 * I have **three** profiles (one is `default` which is what contour supplies, then I have `master` and `light` which are two themed customised things). When Contour comes with changes to the default schema, it's a PITA to upgrade the other two profiles because a simple diff tool just can't help as my custom profiles are always additions versus the default.
 * Automatic update of config files will get messy if entire sub-trees need update which may or may not have some keys.

## Proposal

**1.** Hardcode for each top-level listing map (e.g. `profiles` and `colour_schemes`) the `__default__` key. This is the key where Contour's "official" configuration should be put (e.g., in the `contour generate config ...` output).

This will make updates to the official configuration easier, as we need to only update **one** well-defined and reserved tree.

**2.** (Optionally) Introduce the `__inherits__` or `__base__` (I'm purposefully using some reserved underscore mangling yadda here) which takes as an argument a single (or multiple?) other named entity from the same tree: e.g., a profile might inherit another profile.

This could make custom configurations more composable (a'la CMakePresets), but might make the implementation unnecessarily messy!

**3.** (Irrespective of whether **2.** is implemented...) when resolving a specific `<top-level-tree-structure>.<user-defined-identifier>.whatever.blah` key, use the following logic (in this example, rolling with the profiles, but the `colour_schemes` is the same yadda!):

   1. If `profiles.foo.whatever.blah` exist, resolve to that value.
   2. If `profiles.foo.whatever.blah` does *not* exist, but `profiles.<T>.whatever.blah` exists, resolve to that.
       * If entities might inherit from a **single** but user-specified base "class": `<T>` is that
       * If entities might inherit from a **list of multiple** user-specified base classes: `<T>` is the last in that list
       * If entities might not user-customisable inherit, `<T>` is `__default__`
   3. If no such value exists, fall back to Contour's baked-in default **and emit a warning**

**4.** When new keys are added to the schema (a.k.a. when the warning mentioned above is emitted), *optionally* offer merging such changes into the `__default__` base "class". This way, it will also automatically be inherited into every user-customised profile.

## Example

With this, I could achieve the following way of customisation which would make my profile YAML easier to read and edit:

```yaml

profiles:
  __default__:
     # All the hardcoded b.s.
     ...

  master:
[...]

PR #1126, opened 2023-06-25
Inheritance of configuration profile based on default

Yaraslaut:
Closes https://github.com/contour-terminal/contour/issues/1063

This PR partially implements inheritance of configuration file from proposal #1063 
Configuration profile is loaded based on `default_profile: ...... ` field and this profile is the only profile that must be properly configured. 
Following logic is implemented for loading configuration of profile:
 * user defined `default_profile` is loaded based on provided configuration file,  if entry does not exist error will be emmited and default value providede by contour will be used
 * each profile in the configuration file inherits all values from `default_profile` and if some of the field is provided inside configuration provided value for this profile will be used
 
example:  
```
profiles:
    main:
.......................
    inher:
        shell: "/usr/bin/bash"
        history:
            limit: 0
```
Inheritance also implemented for color chemes with the only difference that `color_schemes.default` is always used as default color chemes and must be properly configured

### DIFF
--- a/metainfo.xml
+++ b/metainfo.xml
@@ -109,6 +109,7 @@
           <li>Do not clear search term when entering search editor again.</li>
           <li>Fixes search term rendering highlighting for search terms containing whitespaces (#966).</li>
           <li>Fixes rendering in cases of glyphs with inverted orientation (#1115).</li>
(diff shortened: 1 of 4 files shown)


christianparpart (approved):
Many thanks!

Issue #1127, opened 2023-06-28
Building error of contour-terminal with fmt soversion bump to 10 on Fedora rawhide


topazus:
### Contour Terminal version

0.3.12.262

### Installer source

GitHub: release page

### Operating System

Fedora rawhide

### Architecture

x86-64

### Steps to reproduce

The fmt library will update to 10.0.0 version, the following build error of contour-terminal appeared:

```bash
[...]

PR #1128, opened 2023-06-29
Add rawhide fedora into github actions

Yaraslaut:
Add fedora rawhide in github action to be aware of issues with dependencies as early as possible

Closes https://github.com/contour-terminal/contour/issues/1127

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -152,7 +152,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os_version: [36, 37, 38, rawhide]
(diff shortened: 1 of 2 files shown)


christianparpart (approved):
i've added some commits to make it pass CI.
Seems like we cannot trivially use the `rawhide` string. It's way easier to give it its CURRENT numerical representation, which currently is `39`. This changes every 6 months to the latest release plus one, obviously.

christianparpart:
Forgot to mention, with this PR we also generally bumped the auto-installed fmtlib dependency to version 10. Note, that version 9.1.0 should still work, as no relevant code changes have been made.

PR #1130, opened 2023-07-02
Fix font inheritance for profiles

Yaraslaut:
At the moment font settings for bold and italic does not inherit font family from regular font.
This cause errors during rendering, this fix usage of same font family for regular bold and italic text

### DIFF
--- a/src/contour/Config.cpp
+++ b/src/contour/Config.cpp
@@ -1602,6 +1602,7 @@ namespace
                      "regular",
                      profile.fonts.regular);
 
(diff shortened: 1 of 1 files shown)


christianparpart (approved):
Great! Many thanks. This was the concern i was raising either in the last PR or in Discord lsat time :-)

Yaraslaut:
> 

Oh, somehow i missed this :( 
I was testing with bash and did not noticed it in the beginning

PR #1135, opened 2023-07-13
clear search when switch to insert mode

elmarsan:
## Description

There is a weird behavior when entering insert vi mode if you were previously doing searches. 
When you search for a term an then switch to insert vi mode, the search term is not deleted.  So the search term continues being evaluated if you still using the terminal.

## Demo of current behavior

[vi-search-mode.webm](https://github.com/contour-terminal/contour/assets/43180519/4b7ef934-91d2-44a5-aec1-39b4801fe9fa)

```markdown
Clear search term when switch to insert vi mode
```

## Motivation and Context

```markdown
Do not execute search term while using insert vi mode
```

## How Has This Been Tested?

- Manually
- Unit tests

## Checklist:

- [X] I have read the [**`CONTRIBUTING`**](https://github.com/contour-terminal/contour/blob/master/CONTRIBUTING.md) document in my spoken language, and understand the terms
- [X] I have added tests to cover my changes.
- [X] I have gone through all the steps, and have thoroughly read the instructions

### DIFF
--- a/metainfo.xml
+++ b/metainfo.xml
@@ -110,6 +110,7 @@
           <li>Fixes search term rendering highlighting for search terms containing whitespaces (#966).</li>
           <li>Fixes rendering in cases of glyphs with inverted orientation (#1115).</li>
           <li>Adds inheritance of profiles in configuration file based on default profile (#1063).</li>
+          <li>Clear search term when switch to insert vi mode (#1135)</li>
         </ul>
       </description>
     </release>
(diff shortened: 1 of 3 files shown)


christianparpart (commented):
I actually liked that behavior but I totally understand why it may be confusing.

Can you add a changeling entry for this into the metainfo.xml?

Yaraslaut:
> I actually liked that behavior but I totally understand why it may be confusing.
> 
> Can you add a changeling entry for this into the metainfo.xml?

Maybe we want to make it configurable then? Then you both can have desired behavior, right ?

christianparpart:
But if it is just me. I do not depend on it

elmarsan:
I believe the best option is make it configurable. For new users like me is weird, but I can imagine long term users might miss the original behavior. 

Shall I make it configurable?

christianparpart (commented):
Many thanks, @elmarsan.

PR #1142, opened 2023-07-23
Fix names to pass tidy and pedantic

Yaraslaut:
While trying to compile with `-DPEDANTIC_COMPILER=ON` and `-DENABLE_TIDY=ON` i am getting errors about names. 
This PR fixes this errors

### DIFF
--- a/src/vtpty/LinuxPty.h
+++ b/src/vtpty/LinuxPty.h
@@ -31,9 +31,10 @@ class LinuxPty final: public Pty
   private:
     class Slave final: public PtySlave
     {
+        int _slaveFd;
+
       public:
-        int slaveFd;
-        explicit Slave(PtySlaveHandle fd): slaveFd { unbox<int>(fd) } {}
+        explicit Slave(PtySlaveHandle fd): _slaveFd { unbox<int>(fd) } {}
         ~Slave() override;
         [[nodiscard]] PtySlaveHandle handle() const noexcept;
         void close() override;
--- a/src/vtpty/LinuxPty.h
+++ b/src/vtpty/LinuxPty.h
@@ -32,8 +32,8 @@ class LinuxPty final: public Pty
     class Slave final: public PtySlave
     {
       public:
-        int _slaveFd;
-        explicit Slave(PtySlaveHandle fd): _slaveFd { unbox<int>(fd) } {}
+        int slaveFd;
+        explicit Slave(PtySlaveHandle fd): slaveFd { unbox<int>(fd) } {}
         ~Slave() override;
         [[nodiscard]] PtySlaveHandle handle() const noexcept;
         void close() override;
--- a/src/vtpty/UnixPty.h
+++ b/src/vtpty/UnixPty.h
@@ -56,9 +56,10 @@ class UnixPty final: public Pty
   private:
     class Slave final: public PtySlave
     {
+        int _slaveFd;
+
       public:
-        int slaveFd;
-        explicit Slave(PtySlaveHandle fd): slaveFd { unbox<int>(fd) } {}
+        explicit Slave(PtySlaveHandle fd): _slaveFd { unbox<int>(fd) } {}
         ~Slave() override;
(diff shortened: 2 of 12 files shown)


src/vtpty/LinuxPty.h
```diff
@@ -32,8 +32,8 @@ class LinuxPty final: public Pty
     class Slave final: public PtySlave
     {
       public:
-        int _slaveFd;
-        explicit Slave(PtySlaveHandle fd): _slaveFd { unbox<int>(fd) } {}
+        int slaveFd;
+        explicit Slave(PtySlaveHandle fd): slaveFd { unbox<int>(fd) } {}
```
christianparpart, line 36: Actually, this looks more like `slaveFd` should never have been public, but private. So it should remain as `_slaveFd`, with leading underscore, because it's a private field member, but it should be made private in fact then, whereas the constructor is public of course.

src/vtpty/UnixPty.h
```diff
@@ -57,8 +57,8 @@ class UnixPty final: public Pty
     class Slave final: public PtySlave
     {
       public:
-        int _slaveFd;
-        explicit Slave(PtySlaveHandle fd): _slaveFd { unbox<int>(fd) } {}
+        int slaveFd;
```
christianparpart, line 60: same here, like in `LinuxPty::Slave`

src/vtrasterizer/DecorationRenderer.cpp
```diff
@@ -218,7 +218,7 @@ auto DecorationRenderer::createTileData(Decorator decoration, atlas::TileLocatio
             auto const x0 = 0u;
             auto const x1 = unbox<unsigned>(width) / 2;
             auto block = blockElement(ImageSize { width, height });
-            return create(block.downsampledSize(), [&]() -> atlas::Buffer {
+            return create(block.getdownsampledSize(), [&]() -> atlas::Buffer {
```
christianparpart, line 221:
```suggestion
            return create(block.getDownsampledSize(), [&]() -> atlas::Buffer {
```
I think that's more readable if it starts capital lettered after the `get` prefix, what do you think?

src/vtrasterizer/Pixmap.h
```diff
@@ -186,37 +186,37 @@ constexpr void drawEllipseArc(PutPixel putpixel, ImageSize imageSize, crispy::Po
...
+    std::function<int(int, int)> filler = [](int, int) {
         return 0xFF;
     };
-    int _lineThickness = 1;
-    int _baseLine = 0; // baseline position relative to cell bottom.
+    int lineThickness = 1;
+    int baseLine = 0; // baseline position relative to cell bottom.
 
-    [[nodiscard]] constexpr ImageSize downsampledSize() const noexcept { return _downsampledSize; }
+    [[nodiscard]] constexpr ImageSize getdownsampledSize() const noexcept { return downsampledSize; }
```
christianparpart, line 198: Hmm, since it's a struct with all data members being public, we do not need this accessor at all, I think it's even better to simply use `Pixmap.downsampledSize` then, and getting rid of the accessor member methods all together that simply return a data field. What do you think?

christianparpart (changes requested):
overall good, but very few minor's I think is worth changing. :)

src/vtpty/LinuxPty.h
Yaraslaut: Changed it

src/vtpty/UnixPty.h
Yaraslaut: Done

src/vtrasterizer/DecorationRenderer.cpp
Yaraslaut: Changed

src/vtrasterizer/Pixmap.h
Yaraslaut: Done

christianparpart (approved):

PR #1149, opened 2023-07-26
Change default version of Qt to 6

Yaraslaut:
script `install-deps.sh` will install dependencies for qt6 by default, while cmake will try to build with qt5, this PR fix this discrepancy


christianparpart (approved):
LGTM. Many thanks :)

PR #1150, opened 2023-07-27
Fix notcurses github actions

Yaraslaut:
This PR fix notcurses github actions. 
The issue is  failing of notcurses from time to time during startup.
When notcurses fails to startup contour is stuck in infinite loop of waiting for some input since we require at leas one input from the user.
I added additional cmake flag to specify whether we want to wait for initial input or not, so this flag is meant only for actions with notcurses tests


christianparpart:
Why did you undo the merge of external undo build.yml?

Is it possible to avoid the third commit (merge master into this branch)?

Yaraslaut:
> Why did you undo the merge of external undo build.yml?
> 
> Is it possible to avoid the third commit (merge master into this branch)?

I separated them because now we need to package contour separately for notcurses-test, and when we have only one file for all this jobs we are increasing time since number of workers for one action is limited. 
Overall i do not have opinion whether they should be merged or separated, if you think that better to merge them i will update it

christianparpart:
I am still trying to understand why this change is helping notcurses here.

Yaraslaut:
You can see that notcuses-demo sometimes fails to start
fish shell : 
```
for i in (seq 1 100);
       contour early-exit-threshold 0  /usr/bin/notcurses-demo  -l -1 -d 0.3 -p /usr/share/notcurses e;
       echo "========================================== $i";
end;
```          
Then, we have a state when no input for contour, and since we are trying to process input at least once here
https://github.com/contour-terminal/contour/blob/59df9d6bee5702dd014185f7cd27b553434f6518/src/contour/TerminalSession.cpp#L231-L235

we are trying to read from pty, and then after timeout we have `errno == EAGAIN`
we exit here from processInputOnce with return true;
https://github.com/contour-terminal/contour/blob/59df9d6bee5702dd014185f7cd27b553434f6518/src/vtbackend/Terminal.cpp#L253-L258

So, we have infinite loop of waiting for some input

Yaraslaut:
Issue is fixed in https://github.com/contour-terminal/contour/pull/1160

PR #1159, opened 2023-08-10
Makes the bell silent

### LINKED ISSUE #1162: Repeatedly outputting `BEL` doesn't produce any sound
### Contour Terminal version

Contour Terminal Emulator 0.3.13-master-362a2868

### Installer source

Github: source code cloned

### Operating System

Fedora 38

### Architecture

x86-64

### Other Software

_No response_

### Steps to reproduce

Run `for i in $(seq 20);do printf '\a'; sleep 1; done` in bash 

### Expected Behavior

Contour should ring bell every one second.

### Actual Behavior

Contour doesn't produce any sound.

### Additional notes

Other ways this can be repro'd is by doing `printf'\a' ` a bunch of time in shell, after a while no sound is produced for BEL character.

Utkarsh-khambra:
Adds a config option `bell` under profile to configure BEL behaviour and Fixes #1162 ,  #1163


src/contour/Config.h
```diff
@@ -275,8 +276,19 @@ struct Config
...
+        assert(prof);
+        return *prof;
+    }
+
+    TerminalProfile const& profile() const noexcept
+    {
+        auto const* prof = profile(defaultProfileName);
+        assert(prof);
+        return *prof;
+    }
```
Utkarsh-khambra, line 291: This is silence gcc error warning about possible nullptr deference.

Yaraslaut (changes requested):

Yaraslaut:
Also, we need to update documentation here  [profiles.md](https://github.com/contour-terminal/contour/blob/master/docs/configuration/profiles.md)

docs/configuration/profiles.md
```diff
@@ -80,6 +80,16 @@ profiles:
 
 
 
+### `bell`
+configuration option determines whether or not BEL character '\a' or 0x07 produces audible sound or not.
```
christianparpart, line 84:
```suggestion
Configuration option determines whether or not `BEL` character `\a` or `0x07` produces audible sound or not.
```

docs/configuration/profiles.md
```diff
@@ -80,6 +80,16 @@ profiles:
...
+``` yaml
+profiles:
+  profile_name:
+    bell: true
+```
+
+
+
 ### `wm_class`
 configuration option defines the class part of the WM_CLASS property of the terminal window. The WM_CLASS property is a standard X11 property used to identify and classify windows by their class and instance.
```
christianparpart, line 94:
```suggestion
Configuration option defines the class part of the `WM_CLASS` property of the terminal window. The `WM_CLASS` property is a standard X11 property used to identify and classify windows by their class and instance.
```

christianparpart (changes requested):
LGTM, only very minors, that might make sense to apply. What do you think?

Utkarsh-khambra:
I read the suggestion making bell to be `off` , `default` or `path`  to make more configurable. Working on that.

Yaraslaut (changes requested):
Looks like not all wav files are good for a bell sound, i just downloaded random one from https://mixkit.co/free-sound-effects/bell/  and it didn't work :(

Utkarsh-khambra:
@Yaraslaut can link a specific one? I tried a bunch of 'em from here already and they worked.

src/contour/ui.template/Terminal.qml
```diff
@@ -59,9 +59,16 @@ ContourTerminal
...
+        id: bellAudioOutput
+        objectName: "BellAudioOutput"
+    }
+
+    MediaPlayer {
         id: bellSoundEffect
-        source: vtWidget.bellSoundSource
+        objectName: "Bell"
+        source: "qrc:/contour/bell.wav"
+        audioOutput: bellAudioOutput
```
christianparpart, line 71: Only the `.qml.in` files were supposed to be git-tracked in the `ui.template` folder. You can basically drop the changes here and it'll have no affect.

In a future/other PR we should actually get rid of these. :)

src/contour/contour.yml
```diff
@@ -176,6 +176,9 @@ profiles:
         # whether or not to put the window into maximized mode.
         maximized: false
 
+        # Determines if BEL character sound is "off" or "default" or is given by file at "path"
```
christianparpart, line 179: I think we can make this a multi-line comment, specifically describing each value.

src/contour/ContourGuiApp.cpp
```diff
@@ -389,6 +389,16 @@ int ContourGuiApp::terminalGuiAction()
...
     newWindow();
 
+    if (auto const& bell = config().profile().bell; bell == "off")
+    {
+        qmlEngine_->rootObjects().first()->findChild<QObject*>("BellAudioOutput")->setProperty("muted", true);
+    }
+    else if (bell != "default")
+    {
+        QUrl const path(bell.c_str());
+        qmlEngine_->rootObjects().first()->findChild<QObject*>("Bell")->setProperty("source", path);
```
christianparpart, line 399: Can `qmlEngine_->rootObjects().first()->findChild<QObject*>("Bell")` return `nullptr`? If so, we should dereference it only if it's assured to be non-null.

src/contour/ContourGuiApp.cpp
```diff
@@ -389,6 +389,16 @@ int ContourGuiApp::terminalGuiAction()
     // Spawn initial window.
     newWindow();
 
+    if (auto const& bell = config().profile().bell; bell == "off")
+    {
+        qmlEngine_->rootObjects().first()->findChild<QObject*>("BellAudioOutput")->setProperty("muted", true);
```
christianparpart, line 394: same here (like below, sorry for the inverse comments) :)

src/contour/Config.h
```diff
@@ -42,6 +42,8 @@
 #include <unordered_map>
 #include <variant>
 
+#include "crispy/assert.h"
```
christianparpart, line 45:
```suggestion
#include <crispy/assert.h>
```

christianparpart (changes requested):
Sorry, I should have used the review functionality from the start. :)

christianparpart (approved):
Many thanks

PR #1165, opened 2023-08-14
Fixes DECSTBM reset``

### LINKED ISSUE #1164: DECSTBM doesn't work when used to reset margins.
### Contour Terminal version

Contour Terminal Emulator 0.3.13-master-7172b40d

### Installer source

Github: source code cloned

### Operating System

Fedora 38

### Architecture

x86-64

### Other Software

_No response_

### Steps to reproduce

Set some margins first by running `printf '\033[1;20r'` and then reset margins by running `printf '\033[r'`.

### Expected Behavior

After reset margin is run, contour should set margins to pagesize.

### Actual Behavior

Contour doesn't reset margins and keeps the original margins that were given in previous DECSTBM invocation.

### Additional notes

_No response_

Utkarsh-khambra:
Closes #1164


src/vtbackend/Screen_test.cpp
```diff
@@ -1580,6 +1580,11 @@ TEST_CASE("ScrollUp.WithMargins", "[screen]")
                 "KLMNO\n"
                 == screen.renderMainPageText());
     }
+    mock.writeToScreen("\033[r");
+    mock.writeToScreen("\033[s");
+    REQUIRE(screen.margin().vertical == Margin::Vertical { .from = LineOffset(0), .to = LineOffset(4) });
+    REQUIRE(screen.margin().horizontal
+            == Margin::Horizontal { .from = ColumnOffset(0), .to = ColumnOffset(4) });
```
christianparpart, line 1587: This is C++20 syntax. We do also compile with C++20 here and there on CI, but we still need to support C++17, at least selectively.

You could make this test depending on the `__cplusplus` version. Not sure that's the best solution.

On the other hand, we can evaluate if (and if not => when) it makes sense to bump the minimum C++ version to C++20.

Utkarsh-khambra: I'll use the regular syntax here. I just didn't know we should be compatible with c++17.

christianparpart (approved):
Many thanks, @Utkarsh-khambra :-)

PR #1168, opened 2023-08-15
Remove a lot of code duplication in UnixPty vs LinuxPty by introducing `crispy::read_selector` API

christianparpart:
Definitely reduces code duplication by abstracting read selector API and providing platform specific implementations (i.e. for Linux using epoll + eventfd, and select() + pipe() otherwise).


src/crispy/read_selector.h
```diff
@@ -0,0 +1,260 @@
...
+        tv.tv_sec = timeout.value().count() / 1000;
+        tv.tv_usec = (timeout.value().count() % 1000) * 1000;
+
+        auto const result =
+            ::select(_fds.back() + 1, &_reader, &_writer, &_except, timeout.has_value() ? &tv : nullptr);
+
+        if (result <= 0)
+            return std::nullopt;
+
+        if (FD_ISSET(_break_pipe[0], &_reader))
```
Yaraslaut, line 89: we are trying to read -1 descriptor here, we see this crash on freebsd

Issue #960, opened 2022-12-29
streamline naming conventions


christianparpart:
about a year ago or so, the naming scheme has changed in some places. That made it just worse.

This ticket is to track that we've fixed that again as part of the bigger goals: having good looking code :) :hankey: 

## Changes

- function parameters don't use any underscores anymore
- private member variables use leading underscore

```cpp
struct SomeStruct 
{
    Type publicMember;
};

class SomeClass
{
  public:
    SomeType publicMember;

    Type someFunction(Type functionParam)
    {
        auto localVariable = _privateMember; 
        _privateMember = functionParam;
        return localVariable;
    }

  private:
    Type _privateMember;
};
```

## Checklist

- [x] `vtpty`
- [x] `vtbackend`
- [x] `vtrasterizer`
- [x] `crispy`
- [x] `text_shaper`
- [x] contour frontend (not recommended to be done *before* the QML PR merge, see #956).
- [x] add CI check to enforce valid naming in CI
- [x] get rid of stdfs.h and related legacy

PR #1169, opened 2023-08-16
clang tidy naming for some type of  variables

Yaraslaut:
refs #960

### DIFF
--- a/src/crispy/boxed.h
+++ b/src/crispy/boxed.h
@@ -38,7 +38,7 @@ namespace helper
 // }}}
 
 template <typename T>
-constexpr bool is_boxed = helper::is_boxed<T>::value;
+constexpr bool IsBoxed = helper::is_boxed<T>::value;
 
 /**
  * Wrapper to provide strong typing on primitive types.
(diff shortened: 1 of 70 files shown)


src/crispy/boxed.h
```diff
@@ -38,7 +38,7 @@ namespace helper
 // }}}
 
 template <typename T>
-constexpr bool is_boxed = helper::is_boxed<T>::value;
+constexpr bool IsBoxed = helper::is_boxed<T>::value;
```
christianparpart, line 41: boxed.h should retain it's C++/STL like naming conventions, maybe?

christianparpart (changes requested):
only very few minors, I think. Overall looking good. Many thanks.

src/crispy/boxed.h
Yaraslaut: this file will be deleted soon, no need to update PR

src/vtpty/UnixPty.cpp
```diff
@@ -369,39 +367,6 @@ int UnixPty::write(std::string_view data)
...
-
-    if (select(nfds, &rfd, &wfd, &efd, &tv) < 0)
-        return -1;
-
-    if (!FD_ISSET(_masterFd, &wfd))
-    {
-        ptyOutLog()("PTY write of {} bytes timed out.\n", size);
-        return 0;
-    }
-
```
christianparpart: this should never have compiled (`blocking` undefined, it was a function-param that was removed). How did this end up in this commit? 🤔

christianparpart (approved):
many thanks :)

PR #1171, opened 2023-08-17
Switch from crispy::boxed to boxed::boxed

Yaraslaut:
`crispy::boxed` was moved to external project `boxed-cpp::boxed`


christianparpart (approved):
you also included the change from `unbox<T>(v)` to `unbox(v)`; I kind of missed that when you added this in boxed-cpp PR :)

LGTM.

scripts/install-deps.sh
```diff
@@ -561,6 +571,7 @@ main()
             ;;
     esac
 
+    fetch_and_unpack_boxed
```
christianparpart, line 574: this needs to be also done for the PowerShell install-deps.ps1 script, similar to

https://github.com/contour-terminal/contour/blob/master/scripts/install-deps.ps1#L35-L38

that should at least fix the windows CI build error.

christianparpart (changes requested):
not sure why some Fedora versions are breaking. Maybe the files were out of sync? (or some CXXFLAGS changes?) 🤔

christianparpart (approved):
Many thanks!

PR #1176, opened 2023-08-20
naming convention fixes in frontend

christianparpart:
refs #960

### DIFF
--- a/src/contour/Audio.h
+++ b/src/contour/Audio.h
@@ -7,13 +7,20 @@
 #include <qbuffer.h>
 #include <qthread.h>
 
-#if QT_VERSION >= 0x060000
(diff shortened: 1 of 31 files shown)


src/contour/Audio.h
```diff
@@ -14,6 +14,13 @@
 #endif
 namespace contour
 {
+
+#if QT_VERSION >= 0x060000
```
Yaraslaut, line 18: I think we can use QT_VERSION_CHECK macro here, I think we use it already in some places

christianparpart: true. i simply moved that line and didn't think about if this line actually could be written better, will do, thanks.

PR #1179, opened 2023-08-21
Update link to point  to our aur file

Yaraslaut:
Update link to AUR package so users download package from Release package

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ for everyday use. It is aiming for power users with a modern feature mindset.
 sudo dnf install contour-terminal
 ```
 
- - `Arch` use AUR [package](https://aur.archlinux.org/packages/contour-git)
+ - `Arch` use AUR [package](https://aur.archlinux.org/packages/contour)
 
 ### Installing via Flatpak
 
--- a/docs/install.md
+++ b/docs/install.md
@@ -22,7 +22,7 @@ sudo dnf install contour-terminal
 
 ## Arch Linux
 
-Please use the AUR, at [https://aur.archlinux.org/packages/contour-git](https://aur.archlinux.org/packages/contour-git).
+Please use the AUR, at [https://aur.archlinux.org/packages/contour-git](https://aur.archlinux.org/packages/contour).
 
 ## Ubuntu Linux
 


README.md
```diff
@@ -49,7 +49,7 @@ for everyday use. It is aiming for power users with a modern feature mindset.
 sudo dnf install contour-terminal
 ```
 
- - `Arch` use AUR [package](https://aur.archlinux.org/packages/contour-git)
+ - `Arch` use AUR [package](https://github.com/contour-terminal/contour-aur.git)
```
christianparpart, line 52: we should probably mention both?

If we exclusively point to ours, it's our responsibility to keep users happy, I'd prefer the distributions (e.g. Arch Linux) to take care of the packaging for **their** distribution.

Users should only use ours, if distribution packages are not up-to-date enough.

Another note is, the AUR on aur.archlinux.org should probably only include releases and not build against latest master - which we try to keep stable, but as the current case has shown, we may change dependencies, and that will break user experience, which will be bad for us again. We should mention this, that these AUR(s) are building from latest master branch and not reflecting stable releases.


Update: Okay, looks like on the official AUR database, there is a release-based AUR:

https://aur.archlinux.org/packages/contour

We should link that one instead, not mentioning the contour-git from aur.archlinux.org at all (nor ours). As ours is used for internal CI testing only (and for curious users, that know what they're doing and maybe breaking) :)

docs/install.md
```diff
@@ -22,7 +22,7 @@ sudo dnf install contour-terminal
 
 ## Arch Linux
 
-Please use the AUR, at [https://aur.archlinux.org/packages/contour-git](https://aur.archlinux.org/packages/contour-git).
+Please use the AUR, at [https://aur.archlinux.org/packages/contour-git](https://github.com/contour-terminal/contour-aur.git).
```
christianparpart, line 25:
```suggestion
Please use the AUR, at [https://aur.archlinux.org/packages/contour](https://aur.archlinux.org/packages/contour).
```

README.md
Yaraslaut: > 

Yes, you are right, i will update it to point to Release packages

PR #1181, opened 2023-08-23
Action to check package for Ubuntu

Yaraslaut:
Adds check for build action to check deb package installation


christianparpart (approved):
Many thanks

PR #1182, opened 2023-08-26
Contrain available vt sequences to current VTLevel

Utkarsh-khambra:
Now available sequences are constrained by the current VTLevel and current terminal state. Currently on SCOSC and DECSLRM are affected by terminal state only. When DECLRMM is enabled then only DECSLRM is available otherwise SCOSC is enabled.


src/vtbackend/Functions.h
```diff
@@ -8,11 +8,15 @@
...
 #include <fmt/format.h>
 
+#include <gsl/span>
+
 #include <array>
 #include <optional>
 #include <string>
 #include <vector>
 
+#include "crispy/indexed.h"
```
christianparpart, line 18:
```suggestion
#include <crispy/indexed.h>
```

src/vtbackend/Functions.h
```diff
@@ -461,9 +465,13 @@ constexpr inline auto CaptureBufferCode = 314;
 
 // clang-format on
 
-inline auto const& functions() noexcept
+#if __cpp_consteval >= 201811L
+consteval auto functions() noexcept
+#else
+constexpr auto functions() noexcept
```
christianparpart, line 471: I think it makes more sense to introduce a `CRISPY_CONSTEVAL` define that falls back to `constexpr` if `consteval` isn't supported.

Maybe this function should now be renamed to `allFunctions`, or simply `all`, because `functions` is now a little bit too synonymous.

src/vtbackend/Functions.h
```diff
@@ -618,10 +623,82 @@ inline auto const& functions() noexcept
     return funcs;
 }
 
+class SupportedSequences
```
christianparpart, line 626: Can we get some documentation above this class name?

src/vtbackend/Functions.h
```diff
@@ -618,10 +623,82 @@ inline auto const& functions() noexcept
...
+    SupportedSequences()
+    {
+        for (auto [index, seq]: crispy::indexed(functions()))
+        {
+            _supportedSequences[index] = seq;
+        }
+        crispy::sort(
+            _supportedSequences,
+            [](FunctionDefinition const& a, FunctionDefinition const& b) constexpr { return compare(a, b); });
+        _lastIndex = _supportedSequences.size();
```
christianparpart, line 638: This looks like you're basically copying ALL of the result of `functions()` into the field member `_supportedSequences`. 

The result of `functions()` should be already sorted, then we should rely on that fact.

We could probably ensure that writing `_supportedSequences = functions()` works as well, therefore, you could even further optimize this and get rid of the constructor all along by declaring the field member like this:

```cpp
std::array<FunctionDefinition, functions().size()> _supportedSequences = functions(); // mind, I proposed to rename functions() :)
```

src/vtbackend/Functions.h
```diff
@@ -618,10 +623,82 @@ inline auto const& functions() noexcept
     return funcs;
 }
 
+class SupportedSequences
+{
+  public:
+    SupportedSequences()
```
christianparpart, line 629:
```suggestion
    constexpr SupportedSequences() noexcept
```
in case we cannot avoid this ctor

src/vtbackend/Functions.h
```diff
@@ -618,10 +623,82 @@ inline auto const& functions() noexcept
...
+    void updateVTLevel(VTType vt) noexcept
+    {
+        auto itr = std::partition(
+            _supportedSequences.begin(),
+            _supportedSequences.end(),
+            [vt](const FunctionDefinition& value) noexcept { return value.conformanceLevel <= vt; });
+
+        _lastIndex = std::distance(_supportedSequences.begin(), itr);
+        gsl::span<FunctionDefinition> availableDefinition{_supportedSequences.begin(), _lastIndex};
+        crispy::sort(
```
christianparpart, line 661: C++20 finally provides a `constexpr` `std::sort()` implementation. We could try using that one, in case all platforms provide it.

src/vtbackend/Functions.h
```diff
@@ -618,10 +623,82 @@ inline auto const& functions() noexcept
...
+            availableDefinition,
+            [](FunctionDefinition const& a, FunctionDefinition const& b) constexpr { return compare(a, b); });
+    }
+
+    void disableSequence(FunctionDefinition seq) noexcept
+    {
+        auto seqIter = std::find(_supportedSequences.begin(), _supportedSequences.begin() + _lastIndex, seq);
+        if (seqIter != _supportedSequences.begin() + _lastIndex)
+        {
+            std::rotate(seqIter, seqIter + 1, _supportedSequences.end());
```
christianparpart, line 671: I keep messing with `std::rotate` (also `std::partition`), maybe we want to add a comment above both to explain what we're doing here exactly? :)

src/vtbackend/Functions.h
```diff
@@ -618,10 +623,82 @@ inline auto const& functions() noexcept
...
+        }
+    }
+
+    void enableSequence(FunctionDefinition seq) noexcept
+    {
+        auto* seqIter = std::find(_supportedSequences.begin() + _lastIndex, _supportedSequences.end(), seq);
+        if (seqIter != _supportedSequences.end())
+        {
+            // Maybe could be done better since rest of the data is sorted
+            std::iter_swap(_supportedSequences.begin() + _lastIndex, seqIter);
```
christianparpart, line 682: things like `_supportedSequences.begin()` as well as `_supportedSequences.begin() + _lastIndex` are written so often, it actually makes sense to provide helper members for them:

```cpp
constexpr auto begin() noexcept { return _supportedSequences.begin(); }
constexpr auto end() noexcept { return std::next(_supportedSequences.begin(), _lastIndex); }
```

And then use these everywhere where we'd have manually written it like this.

src/vtbackend/Functions.h
```diff
@@ -618,10 +623,82 @@ inline auto const& functions() noexcept
...
+            crispy::sort(arr, [](FunctionDefinition const& a, FunctionDefinition const& b) constexpr {
+                return compare(a, b);
+            });
+            ++_lastIndex;
+        }
+    }
+
+  private:
+    std::array<FunctionDefinition, functions().size()> _supportedSequences;
+    size_t _lastIndex;
```
christianparpart, line 694:
```suggestion
    size_t _lastIndex = functions().size();
```
Turns out to be that by default (when looking at the ctor above.

Also, `_lastIndex` does NOT point to the last valid item but to the first item next to the last element. Therefore, we probably want to call that `_endIndex` or similar, to denote this?

src/vtbackend/Functions.h
```diff
@@ -618,10 +623,82 @@ inline auto const& functions() noexcept
...
+                                              _supportedSequences.begin() + _lastIndex + 1);
+            crispy::sort(arr, [](FunctionDefinition const& a, FunctionDefinition const& b) constexpr {
+                return compare(a, b);
+            });
+            ++_lastIndex;
+        }
+    }
+
+  private:
+    std::array<FunctionDefinition, functions().size()> _supportedSequences;
```
christianparpart, line 693:
```suggestion
    std::array<FunctionDefinition, functions().size()> _supportedSequences = functions();
```
That should work (or we could make sure to get it work), effectively eliminating the ctor.

src/vtbackend/Functions_test.cpp
```diff
@@ -17,45 +17,74 @@ inline std::ostream& operator<<(std::ostream& os, FunctionDefinition const& f)
 using namespace std;
 using namespace terminal;
 
+SupportedSequences seq;
```
christianparpart, line 20: I personally think that globals should be used only if we have to. I know that this will add a little bit of code repetition, but it's a trivial line to add to the test function bodies. :)

christianparpart: The name `seq` is not really self explaining. Maybe `sequences` or `availableSequences` or similar?

src/vtbackend/Functions.h
```diff
@@ -618,10 +623,82 @@ inline auto const& functions() noexcept
...
+            [](FunctionDefinition const& a, FunctionDefinition const& b) constexpr { return compare(a, b); });
+        _lastIndex = _supportedSequences.size();
+    }
+
+    [[nodiscard]] gsl::span<FunctionDefinition const> getAllSequences() const noexcept
+    {
+        return gsl::span<FunctionDefinition const>{_supportedSequences.cbegin(), _supportedSequences.cend()};
+    }
+
+    [[nodiscard]] gsl::span<FunctionDefinition const> getSupportedSequencesAtCurrentVTLevel() const noexcept
```
christianparpart, line 646:
```suggestion
    [[nodiscard]] gsl::span<FunctionDefinition const> activeSequences() const noexcept
```

I think, since this class is all about restricting the set of available VT sequences based on VT level, we do not need to name this member like that but simply give it a name that denotes what you're accessing, an array of (active) sequences. We could also get rid of the `get...` prefixes in both functions.

src/vtbackend/Functions.h
```diff
@@ -618,10 +623,82 @@ inline auto const& functions() noexcept
...
+        {
+            _supportedSequences[index] = seq;
+        }
+        crispy::sort(
+            _supportedSequences,
+            [](FunctionDefinition const& a, FunctionDefinition const& b) constexpr { return compare(a, b); });
+        _lastIndex = _supportedSequences.size();
+    }
+
+    [[nodiscard]] gsl::span<FunctionDefinition const> getAllSequences() const noexcept
```
christianparpart, line 641: isn't that function basically the same as the original `functions()` one, except that now it's not sorted across the full range anymore? We should probably return simply the result of `functions` here. 
But then we could simply make this function here static, and move the implementation of `functions()` into this function, and get rid of the other all together.

christianparpart: Based on my comment below (which I wrote first, sorry), we could then also get rid of the get-prefix and all this `allSequences()`

src/vtbackend/Functions.h
```diff
@@ -618,10 +623,82 @@ inline auto const& functions() noexcept
...
+        return gsl::span<FunctionDefinition const>{_supportedSequences.cbegin(), _supportedSequences.cend()};
+    }
+
+    [[nodiscard]] gsl::span<FunctionDefinition const> getSupportedSequencesAtCurrentVTLevel() const noexcept
+    {
+        return gsl::span<FunctionDefinition const>{_supportedSequences.cbegin(),
+                                                   _supportedSequences.cbegin() + _lastIndex};
+    }
+
+    void updateVTLevel(VTType vt) noexcept
```
christianparpart, line 652:
```suggestion
    void reset(VTType vt) noexcept
```
We're resetting the set of available VT sequences based on the given VT level. Since this class is all about that, maybe simply call it `reset` then?

src/vtbackend/Functions.h
```diff
@@ -618,10 +623,82 @@ inline auto const& functions() noexcept
...
+    {
+        auto seqIter = std::find(_supportedSequences.begin(), _supportedSequences.begin() + _lastIndex, seq);
+        if (seqIter != _supportedSequences.begin() + _lastIndex)
+        {
+            std::rotate(seqIter, seqIter + 1, _supportedSequences.end());
+            --_lastIndex;
+        }
+    }
+
+    void enableSequence(FunctionDefinition seq) noexcept
```
christianparpart, line 676: Can you try making this (and the disable-pendant) function `constexpr`? As of C++20, all member functions should (tm) be constexpr aware.

src/vtbackend/Functions_test.cpp
```diff
@@ -17,45 +17,74 @@ inline std::ostream& operator<<(std::ostream& os, FunctionDefinition const& f)
...
 TEST_CASE("Functions.OSC777", "[Functions]")
 {
-    FunctionDefinition const* osc = terminal::selectOSCommand(777);
+    FunctionDefinition const* osc =
+        terminal::selectOSCommand(777, seq.getSupportedSequencesAtCurrentVTLevel());
     REQUIRE(osc);
     CHECK(*osc == NOTIFY);
 }
+
+TEST_CASE("Functions.VTLevelConstrain", "[Functions]")
```
christianparpart, line 68: Many thanks for adding also a test for this! :+1:

src/vtbackend/Terminal.cpp
```diff
@@ -25,6 +25,8 @@
 #include <utility>
 #include <variant>
 
+#include "crispy/logstore.h"
+#include "vtbackend/Functions.h"
```
christianparpart, line 29:
```suggestion
#include <crispy/logstore.h>
#include <vtbackend/Functions.h>
```

src/vtbackend/Terminal.cpp
```diff
@@ -163,6 +165,8 @@ Terminal::Terminal(Events& eventListener,
     setMode(DECMode::TextReflow, true);
     setMode(DECMode::SixelCursorNextToGraphic, true);
 #endif
+    _supportedVTSequences.enableSequence(SCOSC);
+    _supportedVTSequences.disableSequence(DECSLRM);
```
christianparpart, line 169:
```suggestion
    setMode(DECMode::LeftRightMargin, false);
```

This way we could avoid code duplication of actually important logic, reducing potential bugs in the future. :)

christianparpart (changes requested):
I think I'm done reviewing. I went through all of it, mostly quality improvement suggestions, but overall actually looking good. Many thanks!

src/vtbackend/Functions.h
Utkarsh-khambra: One of the ubuntu ci doesn't have new enough version so constexpr algorithms are not present there. But maybe I should also add another macro to conditionally add constexpr?

src/vtbackend/Functions.h
Utkarsh-khambra: Yep will do by conditonally marking it constexpr with macro, since not all compiler in CI support constexpr std algos at the moment.

src/vtbackend/Functions.h
christianparpart: Hm, maybe, but then use this macro only where we know it's not working otherwise, I'd suggest.

The place to put both macros would probably be best in `<crispy/defines.h>`

src/vtbackend/Functions.h
```diff
@@ -461,154 +465,161 @@ constexpr inline auto CaptureBufferCode = 314;
 
 // clang-format on
 
-inline auto const& functions() noexcept
+// Hack to get older compiler work
```
christianparpart, line 468: which older compiler are you referring here to? Might maybe make sense to add that to the commend, such that we know when we can get rid of this hack) :)

src/vtbackend/Functions.h
```diff
@@ -461,154 +465,161 @@ constexpr inline auto CaptureBufferCode = 314;
...
-            RCOLORMOUSEBG,
-            RCOLORHIGHLIGHTFG,
-            RCOLORHIGHLIGHTBG,
-            NOTIFY,
-            DUMPSTATE,
-        };
-        crispy::sort(f, [](FunctionDefinition const& a, FunctionDefinition const& b) constexpr {
+    static auto const funcs = []() constexpr {
+        auto funcs = allFunctionsArray();
+        crispy::sort(funcs, [](FunctionDefinition const& a, FunctionDefinition const& b) constexpr {
```
christianparpart, line 618: this is the workaround you're referring to? One function for declaring the array and this to make it sorted?

src/crispy/defines.h
```diff
@@ -19,3 +19,17 @@
...
+#else
+    #define CRISPY_CONSTEVAL constexpr
+#endif
+
+// Use this only when constexpr std algorithm is not supported but we still
+// wanna mark function constexpr using the std algorithms in their body
+#if (defined(__cpp_lib_constexpr_algorithms) && __cpp_lib_constexpr_algorithms >= 201806L)
+    #define CRISPY_LIB_CONSTEXPR constexpr
+#else
+    #define CRISPY_LIB_CONSTEXPR /**/
```
christianparpart, line 34:
```suggestion
    #define CRISPY_CONSTEXPR constexpr
#else
    #define CRISPY_CONSTEXPR /**/
```

I'd propose to keep it in line with the other macros above.

src/vtbackend/Functions.h
```diff
@@ -618,10 +629,85 @@ inline auto const& functions() noexcept
...
+            availableDefinition,
+            [](FunctionDefinition const& a, FunctionDefinition const& b) constexpr { return compare(a, b); });
+    }
+
+    CRISPY_LIB_CONSTEXPR void disableSequence(FunctionDefinition seq) noexcept
+    {
+        auto seqIter = std::find(begin(), end(), seq);
+        if (seqIter != end())
+        {
+            // Move the disable sequence to the end of array, keep the rest of active sequences sorted
```
christianparpart, line 680:
```suggestion
            // Move the disabled sequence to the end of array, keep the rest of active sequences sorted
```

christianparpart (changes requested):
only minors left. Many thanks :-)

src/vtbackend/Functions.h
Utkarsh-khambra: Yes, Since gcc 9.4 doesn't have std::algorithms we can't call crispy::sort in constexpr contexts, becuase it uses std::swap. So instead of doing constexpr re-implementation of std::swap we can separate sorting and actual array temporarily until we move away from gcc 9.4.

PR #1183, opened 2023-08-27
Fixes bold_italic front not inheriting from regular

christianparpart:
@Yaraslaut if you look a few lines above, we'll already inherit bold from regular and italic from regular (as used to be), but somehow missed that line for bold+italic.

I'm adding this line back. Would you mind having a look over it?


Yaraslaut (approved):
Oh, right, i missed this, seems like my eye filter was looking for two lines and this two lines were filled xDD

PR #1186, opened 2023-09-02
Add image of sixel support of plots inside julia repl

Yaraslaut:
Adds image with julia repl and plot using sixel


christianparpart (approved):
Many thanks

PR #1192, opened 2023-09-08
Fix crash on text selection

Yaraslaut:
After qml merge contour started to crash on text selection, this PR fix this crash. 
As for the reason why, @christianparpart  maybe you have idea ?

### DIFF
--- a/src/contour/display/TerminalWidget.cpp
+++ b/src/contour/display/TerminalWidget.cpp
@@ -1260,12 +1260,9 @@ void TerminalWidget::scheduleRedraw()
         _lastHistoryLineCount = currentHistoryLineCount;
     }
 
-    if (setScreenDirty())
-    {
-        // QCoreApplication::postEvent(this, new QEvent(QEvent::UpdateRequest));
-        if (window())
-            post([this]() { window()->update(); });
-    }
+    // QCoreApplication::postEvent(this, new QEvent(QEvent::UpdateRequest));
+    if (window())
+        post([this]() { window()->update(); });
 }
 
 void TerminalWidget::renderBufferUpdated()
--- a/src/vtbackend/Line.cpp
+++ b/src/vtbackend/Line.cpp
@@ -215,6 +215,7 @@ InflatedLineBuffer<Cell> inflate(TrivialLineBuffer const& input)
     }
 
     assert(columns.size() == unbox<size_t>(input.usedColumns));
+    assert(unbox(input.displayWidth) > 0);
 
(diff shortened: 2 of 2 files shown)


src/vtbackend/Line.cpp
```diff
@@ -215,6 +215,7 @@ InflatedLineBuffer<Cell> inflate(TrivialLineBuffer const& input)
     }
 
     assert(columns.size() == unbox<size_t>(input.usedColumns));
+    assert(unbox(input.displayWidth) > 0);
```
christianparpart, line 218: this is just a quality addition, right?

src/contour/display/TerminalWidget.cpp
```diff
@@ -1260,12 +1260,9 @@ void TerminalWidget::scheduleRedraw()
         _lastHistoryLineCount = currentHistoryLineCount;
     }
 
-    if (setScreenDirty())
-    {
-        // QCoreApplication::postEvent(this, new QEvent(QEvent::UpdateRequest));
-        if (window())
-            post([this]() { window()->update(); });
-    }
+    // QCoreApplication::postEvent(this, new QEvent(QEvent::UpdateRequest));
```
christianparpart, line 1263: So your change here is simply to not mark the screen as dirty, right?

christianparpart: which simply calls `RenderStateManager.touch()`, so that itself doesn't have any side effects.
Since this however is done using atomic bool operations, I really wonder why that has an effect. I sadly cannot reproduce this (maybe my PC too fast? who knows... ah, and it's ARM64 on my end, who knows)....

I looked into into where we use that information. Looks like an optimization relic where I tried to do as minimal renders as possible. Looking at `paint()` impl, we are always rendering, regardless of the current dirty-state on the GUI side.

I am fine in dropping this line (or have it commented out?), but in the long run, we should evaluate if we do  still need this design idea of minimizing render operations (does it still have an effect).

christianparpart (approved):
:+1: I'll leave it up to you on the one question I had. Many thanks ! :-)

src/vtbackend/Line.cpp
Yaraslaut: Yes, just to cover this bug

src/contour/display/TerminalWidget.cpp
Yaraslaut: i will try to get rid of this atomic

PR #1193, opened 2023-09-08
Add cmake target for unit tests

Yaraslaut:
This PR adds cmake target to run unit tests with `cmake --build build --target test`

Also, option for unit tests inherit value of `CONTOUR_TESTING`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,6 +88,9 @@ else()
     message(STATUS "No sanitizer enabled.")
 endif()
 
+if(CONTOUR_TESTING)
+    enable_testing()
+endif()
 # ----------------------------------------------------------------------------
 
 include(ContourThirdParties)
--- a/src/crispy/CMakeLists.txt
+++ b/src/crispy/CMakeLists.txt
@@ -109,7 +109,7 @@ endif()
 # --------------------------------------------------------------------------------------------------------
 # crispy_test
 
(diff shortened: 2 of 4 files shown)


src/CMakeLists.txt
```diff
@@ -14,3 +14,10 @@ add_subdirectory(vtrasterizer)
 if(CONTOUR_FRONTEND_GUI)
     add_subdirectory(contour)
 endif()
+
+if(CONTOUR_TESTING)
+    add_custom_target(test
+    COMMAND crispy_test
+    COMMAND vtbackend_test
+    COMMAND vtparser_test )
+endif()
```
christianparpart, line 23:
```suggestion
```

The cool thing is, we don't need any of it. For real :)

Let's just add the following into the project root's CMakeLists.txt:

```cmake
if(CONTOUR_TESTING)
    enable_testing()
endif()
```

and also of course keep CONTOUR_TESTING, but then we should be already ready to go.

CMake (CTest) automatically finds all test applications, because we added them deeper in the tree via `add_test` already.

christianparpart (changes requested):
LGTM, great idea, but see comment :)

src/CMakeLists.txt
Yaraslaut: Updated

christianparpart (approved):
Many thanks! :-)

PR #1202, opened 2023-09-15
Small refactor of variant access for line

Yaraslaut:
Small update of variant use inside `Line`


christianparpart (approved):
Thanks for the PR.

As a general note, we did the bottom change in an interactive debugging/bug-reproduction session. And rewriting the lines like this turned out to not trigger the bug that @Yaraslaut somehow has gotton during selection spamming.

PR #1206, opened 2023-09-19
Fix OSC 888 after qml merge

Yaraslaut:
Update  OSC 888 implementation to handle QML

### DIFF
--- a/src/contour/display/TerminalWidget.cpp
+++ b/src/contour/display/TerminalWidget.cpp
@@ -684,6 +684,11 @@ void TerminalWidget::paint()
 
         terminal().tick(steady_clock::now());
         _renderer->render(terminal(), _renderingPressure);
+        if (_doDumpState)
+        {
+            doDumpStateInternal();
+            _doDumpState = false;
+        }
     }
     catch (exception const& e)
     {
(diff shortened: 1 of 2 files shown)


src/contour/display/TerminalWidget.cpp
```diff
@@ -684,6 +684,11 @@ void TerminalWidget::paint()
 
         terminal().tick(steady_clock::now());
         _renderer->render(terminal(), _renderingPressure);
+        if(_doDumpState)
```
christianparpart, line 687:
```suggestion
        if (_doDumpState)
```
Didn't clang-format check catch this? hmmm 🤔

src/contour/display/TerminalWidget.cpp
```diff
@@ -996,6 +1001,12 @@ void TerminalWidget::inspect()
 
 void TerminalWidget::doDumpState()
 {
+    _doDumpState = true;
+}
+
+void TerminalWidget::doDumpStateInternal()
+{
+
```
christianparpart, line 1009:
```suggestion
```

src/contour/display/TerminalWidget.cpp
```diff
@@ -1012,6 +1023,7 @@ void TerminalWidget::doDumpState()
         return;
     }
 
+
```
christianparpart, line 1026:
```suggestion
```

christianparpart (commented):
LGTM, but I feel like the only things that didn't look right are things that clang-format should have cought. 🤔 ...

Yaraslaut:
> LGTM, but I feel like the only things that didn't look right are things that clang-format should have cought. 🤔 ...

I also realized that qml image grab does not grab OpenGL part but only qml itself, so I need to update it a bit

christianparpart (dismissed):
Many thanks. :)

PR #1211, opened 2023-09-22
Check both ubuntu packages for startup

### LINKED ISSUE #1209: Contour doesn't start with Qt5
### Contour Terminal version

0.3.13-master-4189e91a

### Installer source

Github: source code cloned

### Operating System

Gentoo

### Architecture

x86-64

### Other Software

_No response_

### Steps to reproduce

Build `contour` with Qt5, when you attempt to run it you will be greeted by this lovely message:

```
Warning: QQmlApplicationEngine failed to load component ((null):0, (null))
Warning: qrc:/contour/ui/main.qml:30:5: Type Terminal unavailable (qrc:/contour/ui/main.qml:30, (null))
Warning: qrc:/contour/ui/Terminal.qml:92:5: AudioOutput is not a type (qrc:/contour/ui/Terminal.qml:92, (null))
```

### Expected Behavior

Contour runs

### Actual Behavior

Contour doesn't run

### Additional notes

The reason for this is because `AudioOutput` doesn't exist on Qt5, it's Qt6 QML type, for Qt5 there is something called `Audio`

Qt5: https://doc.qt.io/qt-5/qml-qtmultimedia-audio.html
Qt6: https://doc.qt.io/qt-6/qml-qtmultimedia-audiooutput.html

Yaraslaut:
Add startup check ob both ubuntu versions and as a consequence both qt versions

Have to disable dump of the state on exit since it does not work on Ub 20.04
Closes #1209


src/contour/CMakeLists.txt
```diff
@@ -63,14 +63,25 @@ function(DeclareQmlImport name import version)
...
 DeclareQmlImport(qml_import_QtQuick "QtQuick" "2.6")
-DeclareQmlImport(qml_import_QtQuickControls "QtQuick.Controls" "2.6")
+DeclareQmlImport(qml_import_QtQuickControls "QtQuick.Controls" "2.2")
 DeclareQmlImport(qml_import_QtQuickLayouts "QtQuick.Layouts" "1.11")
 DeclareQmlImport(qml_import_QtQuickWindow "QtQuick.Window" "2.6")
 DeclareQmlImport(qml_import_QtQuicklabs "Qt.labs.platform" "1.1")
 DeclareQmlImport(qml_import_QtMultimedia "QtMultimedia" "5.12")
 DeclareQmlImport(qml_import_QtGraphicalEffects "QtGraphicalEffects" "1.12" "Qt5Compat.GraphicalEffects")
+DeclareQML(qml_audio "Audio" "AudioOutput")
+DeclareQML(qml_media_player "" "audioOutput: bellAudioOutput")
```
christianparpart, line 83: interesting take. 🤔

christianparpart (commented):
comment for now, due to the one question, otherwise LGTM :)

scripts/ci/Xvfb-contour-run.sh
```diff
@@ -35,7 +35,6 @@ $CONTOUR_PREFIX \
         debug "$LOG" \
         display ${DISPLAY} \
         early-exit-threshold 0 \
-        dump-state-at-exit "${DUMP_DIR}" \
```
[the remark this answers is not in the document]
Yaraslaut: we have some issue with ub 20.04 dump state, but I can not repro it inside vm (on real display and wm), so i disabled it just to check that we can run contour on ub with qt5, since dump is available for notcurses gui if needed

src/contour/CMakeLists.txt
Yaraslaut: do you think that we should do this in some other way? my intent was to keep it as simple as possible so we dont polute qml files itslelf

christianparpart: i think it's still better than going for our own solution. so staying in the domain of CMake I do not know how to do it better. (I mean, we could introduce named arguments to our own CMake functions/macros, but that would probably be too much)

christianparpart (approved):

PR #1212, opened 2023-09-22
Adds VT sequence DECSSCLS and properly handle DECSCLM

### LINKED ISSUE #1204: Add VT sequence: DECSCLM - Scrolling Mode
We currently only implement fast mode, but it might make sense to also implement smooth (slow) mode.
This mode only affects adding new lines at a slow rate, but - in modern age - we could correlate that feature with smooth scrolling (of new lines to be added), see #145.

One open question I am having is how long the linefeed should wait until resuming terminal I/O processing.

### How to implement

One design idea I am having in my mind is, to actually suspend processing for N msecs on received linefeeds, if slow scrolling mode is enabled. This should effectively also suspend reading from the TTY to avoid flooding the internal buffers.

### References

- https://vt100.net/docs/vt510-rm/DECSCLM.html

christianparpart:
Closes #1204

### DIFF
--- a/src/contour/Config.cpp
+++ b/src/contour/Config.cpp
@@ -1778,6 +1778,23 @@ namespace
             }
         }
 
+        if (auto value = profile["slow_scrolling_time"])
+        {
+            usedKeys.emplace(basePath + ".slow_scrolling_time");
+            if (value.IsScalar())
+            {
+                auto time = value.as<unsigned>();
+                if (time < 10 || time > 2000)
+                {
+                    logger("slow_scrolling_time must be between 10 and 2000 milliseconds. Defaulting to 100");
+                    time = 100;
+                }
+                terminalProfile.smoothLineScrolling = chrono::milliseconds(time);
+            }
+            else
+                logger("Invalid value for config entry {}", "slow_scrolling_time");
+        }
+
         if (auto frozenDecModes = profile["frozen_dec_modes"]; frozenDecModes)
         {
             usedKeys.emplace(basePath + ".frozen_dec_modes");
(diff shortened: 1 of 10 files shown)


src/contour/Config.cpp
```diff
@@ -1778,6 +1778,23 @@ namespace
             }
         }
 
+        if (auto value = profile["slow_scrolling_time"])
+        {
+            usedKeys.emplace(basePath + ".slow_scrolling_time");
+            if (value.IsScalar())
+            {
+                auto time = value.as<unsigned>();
+                if (time < 10 || time > 2000)
```
Yaraslaut, line 1787: maybe we can add `=` as well here

src/vtbackend/Screen.cpp
```diff
@@ -797,6 +837,35 @@ void Screen<Cell>::backspace()
...
+        _terminal.settings().smoothLineScrolling = {};
+        return;
+    }
+
+    std::array<float, 9> constexpr NumberOfLinesPerSecond = { {
+        1,  // 0
+        2,  // 1
+        9,  // 2 | defined by spec to be 18 lines per second
+        13, // 3
+        18, // 4 | defined by spec to be 18 lines per second
```
j4james, line 856: This is not a big deal, but if you are revisiting this code, I'd suggest maybe using 3, 6, 9, 12, and 18 for these first 5 settings. That way you'd also be able to select the original speeds used on some of the early DEC terminals: 6 for the VT1xx/VT2xx models; 3, 6, and 12 for the VT340; and you still have 9 and 18 in their standard positions for the VT5xx models.

christianparpart: I better take care of it now, before I forget: #1231 :)

Issue #975, opened 2023-01-09
Display size when resizing


cqexbesd:
## Abstract

When resizing a terminal window via my window manager I would like to see what the new size will be in characters.

## Motivation

I miss this feature from ROXTerm. It's probably a symptom of neurosis but I like to resize windows to round numbers of characters.
Not high priority.

## Specification

![image](https://user-images.githubusercontent.com/6312149/211360413-02ec7bd7-6119-4dd9-a4d5-779babebdc99.png)

This is ROXTerm being resized by dragging the window manager decoration in the corner. The sise updates in real time.

christianparpart:
didn't we have that fixed? @Yaraslaut ? I barely remember that one of you had been working on this. Am i remembering wrong?

On the other hand, I'm having my one grid-resize [refactor branch](https://github.com/contour-terminal/contour/tree/improvement/refactor-inflated-lines). 

Thanks for stressing me @cqexbesd. Please gimme some time. :)

christianparpart:
Lol. I realized I was answering to the wrong topic. Nevertheless, it still applies (what I said). and your wish to display the current screen size can be surely implemented, too. - In various ways. Didn't decide yet to what would be best.

1. Indicator statusline could permanently display the current screen size in characters
2. An overlay could be rendered in the middle of the screen *during* resize.
3. Actual GUI elements could be used to show the current grid screen size *during* resize

cqexbesd:
I think 2 or 3 would be good. 1 is fine but I think it is a different feature.

There is really no hurry to implement my suggestions. I am already quite happy with contour and I can wait for my every whim to be implemented :-)

Yaraslaut:
Implemented in #1203 
@cqexbesd   you now can give it a try now

cqexbesd:
Works great! Thanks!

PR #1214, opened 2023-09-23
Add config option to enable/disable size indicator on resize and make resize indicator small

Utkarsh-khambra:


Yaraslaut (dismissed):
LGTM
Not sure about default value but can change later
Refs https://github.com/contour-terminal/contour/issues/975

Utkarsh-khambra:
I'd like it to be false by default but just kept it to true for no particular reason. What you guys think.

Yaraslaut:
> I'd like it to be false by default but just kept it to true for no particular reason. What you guys think.

I think that false is also good, personally i dont see this indicator to much, but i guess if  it was bothering you we can set it to false

christianparpart (approved):
LGTM.

I think true as default is not a bad default. People being annoyed can always disable the setting. And others have the instant feedback. 
Other than that, the default can always be changed in future versions, especially since we want to rework configuration soonish anyways.

PR #1217, opened 2023-09-26
Adds config option `profile.*.option_as_meta` to allow remapping Option key to Alt on MacOS

christianparpart:
On MacOS, using the Option key with another (e.g. A, B, ...), will result in other Unicode characters. It is typical to remap the Option key as if it would be an Alt key however.
This config option allows the remapping.


Yaraslaut:
Missing info in the docs

christianparpart:
> Missing info in the docs 

Damn, I was thinking about it

Yaraslaut:
> > Missing info in the docs
> 
> Damn, I was thinking about it

I think that we need some tool to gather info from source to insert in different places, like docs or man page, list of VT sequencies

PR #1220, opened 2023-09-27
Fix formatting 

Yaraslaut:
Somehow clang-format 15 started to giving errors while they dont exist

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -39,7 +39,7 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - name: Run clang-format style check for C/C++/Protobuf programs.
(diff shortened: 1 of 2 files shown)


christianparpart (dismissed):

christianparpart:
Do we also want to apply the changes as required by clang-format v16?

Yaraslaut:
> Do we also want to apply the changes as required by clang-format v16?

Yes, i will apply them, which version of clang-format are you using? since if its 15 then formatting can be wrong in other PRs

Yaraslaut:
Also, i am not sure about this changes since locally clang-format does not see problems

Yaraslaut:
I will fix formatting that we have at the moment
 github action doing something wrong since locally clang-format (18.0) does not see this issues

PR #1221, opened 2023-09-28
Initial bits to auto-document VT sequences

christianparpart:
- [x] refactor function API to allow more extensive compile time documentation
- [x] create `vtbackend-doc-tool` CLI to dump out markdown for use with mkdocs


src/vtbackend/Functions_documentation.h
```diff
@@ -0,0 +1,82 @@
...
+
+// NOTE: This file currently serves more as an experiment, on how we could actually start documenting
+//       the various functions. It is not yet used anywhere.
+//
+//       The goal is to have it all constexpr, and that we can generate documentation from it
+//       for markdown, website, and man-pages.
+
+struct Documentation
+{
+    FunctionDefinition sequence;
```
Yaraslaut, line 18: We want to include documentation in config entry as well, so not sure that `FunctionDefinition` can be inside here, maybe some base class and then some additional class for documentation of VT seqs with `FunctionDefinition`

Yaraslaut:
Also, i think that it would be great if we can declare configuration something like this
```
auto constexpr config = Configuration {
     ConfigEntry{ "pty_buffer_size", "104124" , "Size in bytes per PTY Buffer Object"},
     ....
     ConfigEntry{ "profile.font.size", "12", "Initial font size in pixels"},
     ConfigEntry{ "profile.font.locator", "native", "Font locator API", 
                         "Selects an engine to use for locating font files on the system.
                          This is implicitly also responsible for font fallback
                          Possible values are:
                              - native          : automatically choose the best available on the current platform
                              - fontconfig      : uses fontconfig to select fonts
                              - CoreText        : uses OS/X CoreText to select fonts.
                              - DirectWrite     : selects DirectWrite engine (Windows only)"
     },


};
// to load
config.load(path);

```

src/vtbackend/Functions_documentation.h
christianparpart: I should probably not call it `Documentation` in a too common namespace then. you are right.
But config entry definition and documentation should be done in the frontend code and, thus, not conflict with what is in vtbackend.

PR #1223, opened 2023-09-29
Check size of available_sizes to not get SEGV

Yaraslaut:
This PR fixes bug in bad access of `face->available_sizes[0]` because number_of_elements can be zero

### DIFF
--- a/src/text_shaper/open_shaper.cpp
+++ b/src/text_shaper/open_shaper.cpp
@@ -12,6 +12,7 @@
 #include <range/v3/view/iota.hpp>
 
 #include <limits>
+#include <optional>
 
 #include <libunicode/convert.h>
 #include <libunicode/ucd_fmt.h>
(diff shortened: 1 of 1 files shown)


src/text_shaper/open_shaper.cpp
```diff
@@ -206,6 +206,9 @@ namespace
     {
         auto const targetLength = static_cast<int>(pt * dpi.y / 72.0);
         int bestIndex = 0;
+        if(face->num_fixed_sizes == 0 )
+            return bestIndex;
```
christianparpart, line 210: Actually, I think we should return nullopt in such case instead. Because there is sisterly no index, then we should not return index 0. 
This requires sissy a small change to the caller side as well

christianparpart (changes requested):
Good catch but I think we should change function signature and return an optional<size_t> and then return nullopt instead.

src/text_shaper/open_shaper.cpp
Yaraslaut: Updated

src/text_shaper/open_shaper.cpp
```diff
@@ -258,7 +262,11 @@ namespace
 
         if (FT_HAS_COLOR(ftFace))
         {
-            auto const strikeIndex = ftBestStrikeIndex(ftFace, fontSize.pt, dpi);
+            auto const strikeIndexOpt = ftBestStrikeIndex(ftFace, fontSize.pt, dpi);
+            if (!strikeIndexOpt.has_value())
+                return nullopt;
```
christianparpart, line 267:
```suggestion
            {
                FT_Done_Face(ftFace);
                return nullopt;
            }
```

Yaraslaut: Done

christianparpart (approved):

PR #1224, opened 2023-09-30
Fix contour close

Yaraslaut:
When closing from window manager GUI will disappear but process was still alive, this PR fix this behavior

### DIFF
--- a/src/contour/TerminalSession.cpp
+++ b/src/contour/TerminalSession.cpp
@@ -478,7 +478,7 @@ void TerminalSession::onClosed()
     auto const now = steady_clock::now();
     auto const diff = std::chrono::duration_cast<std::chrono::seconds>(now - _startTime);
 
-    if (auto*  localProcess = dynamic_cast<terminal::Process *>(&_terminal.device()))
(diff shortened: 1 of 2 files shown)


src/contour/TerminalSession.cpp
```diff
@@ -478,8 +478,9 @@ void TerminalSession::onClosed()
     auto const now = steady_clock::now();
     auto const diff = std::chrono::duration_cast<std::chrono::seconds>(now - _startTime);
 
-    if (auto const* localProcess = dynamic_cast<terminal::Process const*>(&_terminal.device()))
+    if (auto*  localProcess = dynamic_cast<terminal::Process *>(&_terminal.device()))
```
christianparpart, line 481:
```suggestion
    if (auto* localProcess = dynamic_cast<terminal::Process *>(&_terminal.device()))
```

christianparpart (commented):
I'm surprised that clang-format didn't catch this? 🤔

src/contour/TerminalSession.cpp
Yaraslaut: xD ye, clang format shows weird behavior lately

christianparpart (approved):
Nice. Many thanks. :)

PR #1225, opened 2023-09-30
Clean up C++ namespaces per library

christianparpart:
The idea is to not have the kinda legacy `terminal` global namespace across all libraries. Especially since vtpty and vtparser are two libraries that may very well be used by other projects.

Also, giving each library its own namespace makes it easier to spot inter-library dependencies.

So, each library gets its own equally named prefix.


src/vtbackend/primitives.h
```diff
@@ -18,9 +15,13 @@
 // - [ ] rename all History to Scrollback
 // - [ ] make sense out of all the semantically different line primitives.
 
-namespace terminal
+namespace vtbackend
 {
 
+using LineCount = vtpty::LineCount;
```
Yaraslaut, line 21: I would expect it to be inside vtbackend itself

src/vtrasterizer/DecorationRenderer.cpp
```diff
@@ -157,7 +157,7 @@ auto DecorationRenderer::createTileData(Decorator decoration, atlas::TileLocatio
             auto const y1 = max(0u, underlinePosition() + thickness);
             // y1 - 3 thickness can be negative
             auto const y0 = max(0, static_cast<int>(y1) - 3 * static_cast<int>(thickness));
-            auto const height = Height(y1 + thickness);
+            auto const height = vtbackend::Height(y1 + thickness);
```
Yaraslaut, line 160: shouldn't it be from `crispy` ?

src/vtrasterizer/RenderTarget.h
```diff
@@ -9,23 +9,24 @@
...
+namespace vtbackend
 {
 struct BackgroundImage;
 }
 
-namespace terminal::rasterizer
+namespace vtrasterizer
 {
 
+using ImageSize = crispy::image_size;
```
Yaraslaut, line 28: You changed it to direct type `crispy::image_size` inside `Pixmap.h`

src/vtrasterizer/TextureAtlas.h
```diff
@@ -382,17 +382,17 @@ struct TileSliceIndex
 
 /// Constructs a container to conveniently iterate over sliced tiles of the given
 /// input `bitmapSize`.
-constexpr auto sliced(Width tileWidth, uint32_t offsetX, ImageSize bitmapSize)
+constexpr auto sliced(crispy::width tileWidth, uint32_t offsetX, crispy::image_size bitmapSize)
```
Yaraslaut, line 385: `Width` -> `vtbackend::Width` maybe ?

Yaraslaut (changes requested):
Great, now namespaces much cleaner  
Just few minor things left to address

examples/watch-mouse-events.cpp
```diff
@@ -132,15 +132,16 @@ struct MouseTracker final: public BasicParserEvents
...
-    MouseTracker() noexcept: savedTermios { detail::getTerminalSettings(STDIN_FILENO) }, vtInputParser { *this }
+    MouseTracker() noexcept:
+        savedTermios { vtpty::detail::getTerminalSettings(STDIN_FILENO) }, vtInputParser { *this }
     {
         auto tio = savedTermios;
         tio.c_lflag &= static_cast<tcflag_t>(~(ECHO | ICANON));
         tio.c_cc[VMIN] = 1;  // Report as soon as 1 character is available.
         tio.c_cc[VTIME] = 0; // Disable timeout (no need).
-        detail::applyTerminalSettings(STDIN_FILENO, tio);
+        vtpty::detail::applyTerminalSettings(STDIN_FILENO, tio);
```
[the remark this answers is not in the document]
christianparpart: true, i just didn't know it might be more useful than initially anticipated ;)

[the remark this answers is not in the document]
christianparpart: i suggest to rename it to `util` instead, because it's platform dependant interface (POSIX)

src/vtrasterizer/RenderTarget.h
christianparpart: i know. i just didn't decide yet :)

src/vtbackend/primitives.h
christianparpart: LineCount and ColumnCount along with PageSize are in vtpty, because a PTY exposes and configures the page size. this is why we're having it there

src/vtrasterizer/RenderTarget.h
christianparpart: i now decided to move ImageSize/Width/Height into vtpty, because - just like PageSize - it's needed to configure as well as expose the pixel dimensions of the terminal.

only one problem remains: `text_shaper` doesn't make sense to get that from vtpty, for obvious reasons. I think I'll simply dup it here.

src/vtrasterizer/TextureAtlas.h
christianparpart: adapted.

src/vtpty/UnixUtils.cpp
```diff
@@ -3,25 +3,26 @@
...
 {
 
 UnixPipe::UnixPipe(unsigned flags): pfd { -1, -1 }
 {
 #if defined(__linux__)
-    if (pipe2(pfd, static_cast<int>(flags)) < 0)
+    if (pipe2(pfd.data(), static_cast<int>(flags)) < 0)
         throw std::runtime_error { "Failed to create PTY pipe. "s + strerror(errno) };
 #else
     if (pipe(pfd) < 0)
```
Yaraslaut, line 20: what about this `pfd`? should it be also `.data()`?

Yaraslaut (changes requested):
I think that all namespaces super clear now.
Just few comments about other things

src/vtpty/UnixUtils.cpp
christianparpart: damn, nice catch.

Yaraslaut (approved):
Great :)

PR #1227, opened 2023-10-01
Better check inside install-deps.sh script

Yaraslaut:
`which` does not always return proper exit status 
See https://stackoverflow.com/questions/592620/how-can-i-check-if-a-program-exists-from-a-bash-script

### DIFF
--- a/scripts/install-deps.sh
+++ b/scripts/install-deps.sh
@@ -42,11 +42,11 @@ fetch_and_unpack()
     FULL_DISTFILE="$SYSDEPS_DIST_DIR/$DISTFILE"
 
     if ! test -f "$FULL_DISTFILE"; then
-        if which wget &>/dev/null; then
+        if command -v wget &>/dev/null; then
             wget -O "$FULL_DISTFILE" "$URL"
-        elif which curl &>/dev/null; then
+        elif command -v curl &>/dev/null; then
             curl -L -o "$FULL_DISTFILE" "$URL"
-        elif which fetch &>/dev/null; then
+        elif command -v fetch &>/dev/null; then
             # FreeBSD
             fetch -o "$FULL_DISTFILE" "$URL"
         else


scripts/install-deps.sh
```diff
@@ -42,11 +42,11 @@ fetch_and_unpack()
     FULL_DISTFILE="$SYSDEPS_DIST_DIR/$DISTFILE"
 
     if ! test -f "$FULL_DISTFILE"; then
-        if which wget &>/dev/null; then
+        if command -v wget &>/dev/null; then
```
christianparpart, line 45: Does this work on FreeBSD / Mac as well?

christianparpart: I'm asking because at least for FreeBSD there's usually no bash installed, so it has to be sh syntax.

Yaraslaut: FreeBSD build is green, so this should work

christianparpart: true. just checked it's not a false negative. but all good. yay

christianparpart (approved):

scripts/install-deps.sh
herrhotzenplotz: I don't know why, but a notification about this thread showed up in my e-mail and I felt like chiming in because I believe there are misconceptions here.

Recent versions of contour are totally broken on FreeBSD which is why I did delete the WIP patch for the FreeBSD Ports and lost interest in contour - which is kind of sad given the work I put into it.

Thus:

If you want to check for bashisms please use the "checkbashisms" tool. It prints a couple of warnings about said script. If you have wget installed the script will probably work, if you don't it attempts to invoke wget nevertheless.

> FreeBSD build is green, so this should work

Well no, I doubt anyone tested this and the CI tests are broken because [wget gets installed](https://github.com/contour-terminal/contour/blob/master/.github/workflows/build.yml#L206). If you remove the line it will probably cause the CI to fail. For exactly the right reason - which is also why I believe this patch is wrong. Also, the above linked best SO answer's example use is wrong too.

In any case, fixing the bashisms gets you past the dependency installation script.

The test suite runs fine, however I believe it is broken too:

```console
$ build/src/contour/contour
[error] Missing key .logging.enabled. Using default: false.
[error] Missing key .logging.file. Using default: "".
[error] Missing key background_image.path. Using default: "".
[error] Missing key .shell. Using default: "".
[error] Missing key .refresh_rate. Using default: 0.
[error] Superfluous config key found: profiles.main.frozen_dec_modes.2026
Info: libpng warning: iCCP: known incorrect sRGB profile ((null):0, (null))
Warning: Invalid audio device ((null):0, (null))
Warning: Failed to create a gst element for the audio device, using a default audio sink ((null):0, (null))
[error] Attempt to freeze batched rendering. Ignoring.
Bus error (core dumped)
$
```

There's multiple things that stand out to me in the above log:

- There is no configuration file in my home directory, so why all the errors? 
- Invalid audio device? I guess this is an assumption about pulseaudio or alsa. The first of which may not exist and the second definitely doesn't exist.
- Bus error? haven't seen that in a while on x86. Might be a nasty bug.

Anyways, this is what I've been hitting multiple times and I didn't wanna continue spamming bug reports just because things kept breaking and weren't tested appropriately.

PR #1231, opened 2023-10-03
Match speeds for DECSSCLS as defined by old DEC VT1xx and VT2xx terminals

christianparpart:
References @j4james comment in https://github.com/contour-terminal/contour/pull/1212/files#r1344674416

I hope I did understand you correctly there?

### DIFF
--- a/src/vtbackend/Screen.cpp
+++ b/src/vtbackend/Screen.cpp
@@ -847,11 +847,13 @@ void Screen<Cell>::setScrollSpeed(int speed)
         return;
     }
 
+    // NB: Match speeds as defined by old DEC VT1xx and VT2xx terminals.
+    // See https://github.com/contour-terminal/contour/pull/1212/files#r1344674416
     std::array<float, 9> constexpr NumberOfLinesPerSecond = { {
-        1,  // 0
-        2,  // 1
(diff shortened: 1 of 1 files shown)


src/vtbackend/Screen.cpp
```diff
@@ -847,11 +847,13 @@ void Screen<Cell>::setScrollSpeed(int speed)
...
+    // See https://github.com/contour-terminal/contour/pull/1212/files#r1344674416
     std::array<float, 9> constexpr NumberOfLinesPerSecond = { {
-        1,  // 0
-        2,  // 1
+        3,  // 0
+        6,  // 1
         9,  // 2 | defined by spec to be 18 lines per second
-        13, // 3
+        12, // 3
         18, // 4 | defined by spec to be 18 lines per second
```
j4james, line 857: That's exactly what I meant. Thank you.

However, I just noticed the comment for 2 is wrong. You've got both the 2 and 4 comments saying "defined by spec to be **18** lines per second". The first one should assumedly be "**9** lines per second".

christianparpart: Ooops, copy'n'paste bug. thx ;)

j4james (approved):

PR #1233, opened 2023-10-06
Use color scheme based on OS system dark/light mode preferences

### LINKED ISSUE #604: default config's color scheme should respect dark/light mode OS settings
Windows and OSX have it. And at least KDE version 5.24 has it too .

The default color scheme from there default config should respect that setting and pick some appropriate colors.

christianparpart:
Closes #604 

### Checklist

- [x] get dark/light theme mode events on Linux
- [x] get dark/light theme mode events on MacOS
- [x] get dark/light theme mode events on Windows ([how](https://learn.microsoft.com/en-us/windows/apps/desktop/modernize/apply-windows-themes))
- [x] config: extend config to allow specifying dark and light mode themes (then the system setting will decide what to chose from) or just one and then this is always used
- [x] apply color scheme based on that.
- [x] propagate dark/light mode changes to the app (VT extension to query state and get realtime notifications on state changes)
- [x] reflect changes in documentation
- [x] changelog entry added to metainfo.xml

https://github.com/contour-terminal/contour/assets/56763/ea4b5867-defc-489b-bdf2-9aa635275ff0

### DIFF
--- a/src/contour/ContourGuiApp.cpp
+++ b/src/contour/ContourGuiApp.cpp
@@ -12,7 +12,12 @@
 #include <crispy/utils.h>
 
 #include <QtCore/QProcess>
+#if !defined(__APPLE__) && !defined(_WIN32)
+    #include <QtDBus/QDBusConnection>
+#endif
+#include <QtDBus/QtDBus>
 #include <QtGui/QGuiApplication>
+#include <QtGui/QStyleHints>
 #include <QtGui/QSurfaceFormat>
 #include <QtQml/QQmlApplicationEngine>
 #include <QtQml/QQmlContext>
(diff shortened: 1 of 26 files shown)


src/contour/ContourGuiApp.cpp
```diff
@@ -347,6 +363,36 @@ int ContourGuiApp::terminalGuiAction()
...
+            const uint result = reply.arguments()
+                                    .first()
+                                    .value<QDBusVariant>()
+                                    .variant()
+                                    .value<QDBusVariant>()
+                                    .variant()
+                                    .toUInt();
+
+            _colorPreference = getColorPreferenceFromFDO(result);
+            fmt::print("color theme mode: {}\n", _colorPreference); // TODO: remove debug print
```
Yaraslaut, line 392: This

src/contour/ContourGuiApp.cpp
```diff
@@ -409,6 +455,22 @@ int ContourGuiApp::terminalGuiAction()
...
+{
+    if (nameSpace == QLatin1String("org.freedesktop.appearance") && key == QLatin1String("color-scheme"))
+    {
+        const uint result = value.variant().toUInt();
+        auto newValue = getColorPreferenceFromFDO(result);
+
+        if (_colorPreference != newValue)
+        {
+            _colorPreference = newValue;
+            fmt::print("color preferenced changed to {} mode\n", _colorPreference); // TODO: remove debug print
```
Yaraslaut, line 468: This

src/contour/contour.yml
```diff
@@ -452,7 +452,15 @@ profiles:
...
-        colors: "default"
+        #
+        # This can be either the name to a single colorscheme to always use,
+        # or a map with two keys (dark and light) to determine the color scheme to use for each.
+        # The dark color scheme is used when the system is configured to prefer dark mode and light theme otherwise.
+        #
+        # Default: "default"
+        colors:
+            light: onelight
+            dark: onedark
```
Yaraslaut, line 463: It is not clear what `onelight` and `onedark` means

Yaraslaut (commented):
Documentation is missing :)

src/contour/contour.yml
christianparpart: just two different color themes that I happen to have locally present. I did not come up with nice default values here. Also, we must provide a default for dark theme and light theme (so not just one) with this PR in the future as well.

christianparpart:
> Documentation is missing :)

@Yaraslaut I sadly didn't decide yet how the configuration should look like. What you see is my proposal. But if you have an idea on how it could be done better, I'm absolutely open - then of course we must document that within this PR as well :)

christianparpart:
MacOS bits as provided by @uspasojevic96:

```objc++
#import <Cocoa/Cocoa.h>
- (BOOL)isDarkMode {
    NSAppearance *currentAppearance = [NSAppearance currentAppearance];
    if (currentAppearance) {
        // Check if the current appearance is dark
        if (@available(macOS 10.14, *)) {
            return [currentAppearance bestMatchFromAppearancesWithNames:@[NSAppearanceNameDarkAqua]] == NSAppearanceNameDarkAqua;
        }
    }
    
    // Default to light mode if appearance cannot be determined
    return NO;
}
```

For event subscription:

```objc++
[[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(appearanceDidChange:) name:NSAppearanceDidChangeNotification object:nil];
```

I will probably need some help to clue that into the code base, as I've never touched ObjC/++ just yet :|

PR #1234, opened 2023-10-07
Fix infinite loop when next line does not exist

Yaraslaut:
Fix error found by @Utkarsh-khambra  on vim `f` press. 
Issue is that when looking for next character to the right not always we can go to the next line


christianparpart (approved):
:+1: many thanks

PR #1235, opened 2023-10-07
Check naming of class/struct members

Yaraslaut:
Add clang-tiy checks for naming of data mamebers inside struct/class
I was confused previouslu and was thinking that `ClassMemberCase` will check naming convention for the class/struct members, but it only checks `static` members. 
This PR adds clang-tidy checks for the names of private and public members


christianparpart (approved):
looks good to me. many thanks :)

Issue #1238, opened 2023-10-08
Building dependencies for Debian: qml6-module-qt5compat-graphicaleffects


tkapias:
### Contour Terminal version

Contour Terminal Emulator 0.4.0-master-ce5d3bb

### Installer source

Github: source code cloned

### Operating System

Debian GNU/Linux trixie/sid Linux 6.5.0-1-amd64

### Architecture

x86-64

### Steps to reproduce

Build contour following the readme.

### Expected Behavior

The terminal window should open when running.

### Actual Behavior

``` bash
No window and the console output [error] Missing key .logging.enabled. Using default: false.                                                                                               
[error] Missing key .logging.file. Using default: "".      
[error] Missing key background_image.path. Using default: "".    
[error] Missing key .refresh_rate. Using default: 0.                                                                                                      
[error] Superfluous config key found: profiles.main.frozen_dec_modes.2026    
Info: libpng warning: iCCP: known incorrect sRGB profile ((null):0, (null))  
Warning: QQmlApplicationEngine failed to load component ((null):0, (null))   
Warning: qrc:/contour/ui/main.qml:30:5: Type Terminal unavailable (qrc:/contour/ui/main.qml:30, (null))
Warning: qrc:/contour/ui/Terminal.qml:8:1: module "Qt5Compat.GraphicalEffects" is not installed (qrc:/contour/ui/Terminal.qml:8, (null))
```

### Additional notes

The issue is resolved by installing `qml6-module-qt5compat-graphicaleffects` with apt.

This dependency is in `./scripts/install-deps.sh` for Ubuntu, but not for Debian.

Yaraslaut:
Thanks @tkapias  for taking the time to create the issue and let us know which package is missing, we switched to qml recently and there have been no releases with qml yet, so not all distros/platforms are thoroughly tested, I will fix this issue in `install-deps.sh` script

PR #1239, opened 2023-10-08
Move qml package to common list

Yaraslaut:
closes https://github.com/contour-terminal/contour/issues/1238


christianparpart (approved):
:+1:

PR #1245, opened 2023-10-10
feat: Add OpenSelection action, and honor case in the configuration bindings

Ne02ptzero:
## Description

This patch adds the `OpenSelection` action, which can be used to quickly open a link in the buffer. With an appropriate word selector, one can select an URL and open it with a keybind.

It is a variation of this patch:
https://st.suckless.org/patches/open_copied_url/, which I use daily to quickly open URLs in chat, mails, diffs, etc.

I have not tested it, but it should work with other URIs, like file:// and such.

I've also fixed a minor behavior I encountered while developing this patch: By default, the Config will upper() the keybind used when it is a simple key, but for simple Alt+Key binds (in my case, I use Alt+o for this feature), the event sent to the terminal is Alt+o (Alt+6F), while the configuration stores Alt+O (Alt+4F). I've corrected this behavior, which might break stuff; but it seemed more logical to me this way.

- [X] I have read the [**`CONTRIBUTING`**](https://github.com/contour-terminal/contour/blob/master/CONTRIBUTING.md) document in my spoken language, and understand the terms
- [X] I have updated (or added) the documentation accordingly.
- [X] I have gone through all the steps, and have thoroughly read the instructions

### DIFF
--- a/metainfo.xml
+++ b/metainfo.xml
@@ -135,6 +135,7 @@
           <li>Adds capital `A` and `I` keys to switch from normal mode back to insert mode, too.</li>
           <li>Adds size indicator window on resize (#1203).</li>
           <li>Adds config entry `profile.*.size_indicator_on_resize` to control size indicator on resize and makes resize indicator small.</li>
+          <li>Adds OpenSelection action, and honor case in the configuration bindings</li>
         </ul>
       </description>
     </release>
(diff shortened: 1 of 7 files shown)


christianparpart (commented):
This is a great addition, and should not be a problem to still keep around when we will eventually (hopefully) have proper clickable hyperlink support for auto-detected URLs and local files.

Many thanks.

Only one addition: Would you mind adding a changelog text to /metainfo.xml ?

christianparpart (approved):
:+1: Yes, this was correct. Many thanks, @Ne02ptzero

Issue #1247, opened 2023-10-10
Consider integrating window alerting with the bells


Ne02ptzero:
Hello!

I currently have this patch running locally for my terminals:
```diff
diff --git a/src/contour/ui.template/Terminal.qml.in b/src/contour/ui.template/Terminal.qml.in
index 05dca56a..ada218fd 100644
--- a/src/contour/ui.template/Terminal.qml.in
+++ b/src/contour/ui.template/Terminal.qml.in
@@ -194,6 +194,7 @@ ContourTerminal
         if(bellSoundEffect.playbackState === MediaPlayer.PlayingState)
             bellSoundEffect.stop()
         bellSoundEffect.play()
+        Window.window.alert(0)
     }

     function updateFontSize() {
```

Which enables the visual cues with my window manager when a bell is ringed from a terminal (Please note that I use the `bell: "off"` in my configuration, as I do not want any sounds). This feature is used by weechat (When an highlights pops, or a DM), mutt (When I receive a new mail), or many scripts where I added `echo "\a"` at the end to notify me that the script is done (Big compilations are the prime example of this).

![2023-10-10_264x39_scrot](https://github.com/contour-terminal/contour/assets/6032832/3f4a1c46-9725-404e-ba74-cd960241eac0)
(In this example, I'm in another workspace and the terminal is raising an alert).

I'm fully aware that the above diff is pretty basic, and cannot be merged as-is. I'm open to directions/cues for properly implementing this upstream, as I'm sure I'm not the only one used to this feature from different terminal emulators.

uspasojevic96:
Thanks for the idea and willingness to implement this, here is a checklist that I came up with on the fly:

- [ ] Create config option for window alerts
- [ ] Create function in QML which would be called to run window alerts
- [ ] Create C++ function which would trigger QML function on specific event

PR #1251, opened 2023-10-11
feat: Enable window alert on terminal bells

Ne02ptzero:
With a slight rework of the configuration, as discussed in https://github.com/contour-terminal/contour/issues/1247:

Before:

    bell: "off"

After:

    bell:
      sound: "off"
      alert: true

The original value was moved to bell.sound, and a boolean was added, bell.alert; when this boolean is true, a window alert will be raised each time a bell is ringed.

Closes https://github.com/contour-terminal/contour/issues/1247


Yaraslaut (changes requested):
I think that by default we can enable this behavior. 
Also, at the moment we do not create documentation for the web page automatically http://contour-terminal.org/
So we need to update it manually, for the configuration this is  `/docs/configuration/profiles.md`

Another issue is update of existing configuration files, so we might want to add this info somewhere so users aware of breaking changes

Ne02ptzero:
> I think that by default we can enable this behavior. 

Sure, thing!

> Another issue is update of existing configuration files, so we might want to add this info somewhere so users aware of breaking changes

How would I do that? Is there any file in the repo to notify the users of breaking changes?

Yaraslaut:
> > I think that by default we can enable this behavior.
> 
> Sure, thing!
> 
> > Another issue is update of existing configuration files, so we might want to add this info somewhere so users aware of breaking changes
> 
> How would I do that? Is there any file in the repo to notify the users of breaking changes?

No, we dont have such,  we can add this info in metainfo file so user can see that configuration entry was moved somewhere

Ne02ptzero:
All right, done. I think since the bell original configuration was never released, the impact should be limited to users using unstable/master builds. Feel free to propose a better change for `metainfo.xml` if my change is not enough.

Yaraslaut:
> All right, done. I think since the bell original configuration was never released, the impact should be limited to users using unstable/master builds. Feel free to propose a better change for `metainfo.xml` if my change is not enough.

Oh, indeed, its been a while since last release and i missed that we added it after previous release, sorry for that, i think better to omit then this comment

Yaraslaut (approved):
Many thanks @Ne02ptzero

christianparpart:
> Oh, indeed, its been a while since last release and i missed that we added it after previous release, sorry for that, i think better to omit then this comment

Yeah, I really hope we can release 0.4.0 as soon as possible (as in: we feel confident with having master as stable as possible)

christianparpart (approved):

PR #1255, opened 2023-10-11
Fix install-deps script to handle freeBSD

Yaraslaut:
Thanks @herrhotzenplotz  
I updated install-deps script to fix warnings from `checkbashisms`


christianparpart (approved):

herrhotzenplotz:
lgtm, why didn't you remove the wget install step on CI? That would ultimately test exactly the change.

Yaraslaut:
> lgtm, why didn't you remove the wget install step on CI? That would ultimately test exactly the change.

You are right, this will be a good check, will change it in following PR for freeBSD

Issue #1125, opened 2023-06-22
Update contour.yml config file to match actually existing keys


Marenz:
Through constant updates, config files become out-dated and start missing keys or contain keys that have been removed or renamed.

This issue is about providing a way to auto-fix those keys, either automatically or by specific user request..

Yaraslaut:
We wanted to make it happen at some point,  the issue with that is that at the moment `yaml-cpp` API does not allow  user to write comments to the files, meaning that if we try to update config file of the user it will delete all comments in the file.
Maybe @christianparpart  have some ideas about that

Yaraslaut:
Will be implemented as part of  https://github.com/contour-terminal/contour/issues/1282

PR #1264, opened 2023-10-13
Move config errors in config log

Yaraslaut:
Move logging of config into configLog so users do not see them as errors if something is missing, for example in the case
when user defined only one key to change default behavior 
Refs https://github.com/contour-terminal/contour/issues/1125


christianparpart (approved):
this was disturbing me all long already, too. Many thanks :-)

christianparpart (dismissed):

PR #1288, opened 2023-10-17
Make github actions faster for PRs

Yaraslaut:
Atempt to make github actions complete in some reasonable time for pull requests 

Longest actions 
 * Windows 20 min
 * Fedora 16-17 min
While we can move fedora builds to freebsd build so we check them only after merge, windows build is more essential and unfortunately slow

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,6 +36,7 @@ jobs:
 
   # {{{ ArchLinux
   archlinux:
+    if: github.ref == 'refs/heads/master'
     strategy:
       matrix:
         arch:
@@ -93,7 +94,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os_version: [38, 39]
+        os_version: [39]
         arch:
           [
             "linux/amd64 x86_64"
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -1138,7 +1048,6 @@ jobs:
     name: Create Github release
     runs-on: ubuntu-latest
     needs:
-      - archlinux
```
christianparpart, line 1141: I just realized. With that change, we will NOT build archlinux binaries for ArchLinux upon release!


Maybe we can keep it all in (not creating a secondary CI pipeline) and make some individual jobs conditional. What do you think?
I think it should go something like this:

```yaml
jobs:
    archlinux:
        if: github.ref == 'refs/heads/master'
        runs-on: ubuntu-latest
```

that should do the trick. this way we can still keep it all in one build CI pipeline without potentially complicating release pipeline again :)

Yaraslaut: Oh, i didn't know that this is possible :( 
will update PR now

christianparpart (approved):
yay :)

PR #1295, opened 2023-10-20
Fix error with out of bounds access

Yaraslaut:
Fix sixel parser bug

### DIFF
--- a/src/vtbackend/SixelParser.cpp
+++ b/src/vtbackend/SixelParser.cpp
@@ -279,7 +279,7 @@ void SixelParser::leaveState()
                 auto const pad = _params[1];
 
                 auto const imageSize =
-                    _params.size() > 2
+                    _params.size() > 3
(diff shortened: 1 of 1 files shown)


src/vtbackend/SixelParser.cpp
```diff
@@ -279,7 +279,7 @@ void SixelParser::leaveState()
                 auto const pad = _params[1];
 
                 auto const imageSize =
-                    _params.size() > 2
+                    _params.size() > 3
```
christianparpart, line 282: how did you find it? Did this cause a visible bug to the app? If so, we might want to mention that in the changelog. :-)

Yaraslaut: This caused a crush when trying to run [background_select.sh](https://github.com/hackerb9/vt340test/blob/main/j4james/background_select.sh), so not a visible bug

christianparpart: nice. thx.

Issue #1292, opened 2023-10-19
in case of bad font config, default to something existing


christianparpart:
Just in case somebody misconfigures a font (by accident) and that font cannot be found, then Contour should still boot up. with a default font that definitely exists on this operating system.

in such case, an error should still be reported (via `errorlog()` as first step)

### Future improvements

We do not have a good error reporting API for that yet, apart from `errorlog()`, but that could be hooked into the frontend to show the messages also on the GUI instance's terminal window itself (either directly on screen, if no shell spawned yet, or in the statusline area)

Issue #1282, opened 2023-10-16
Refactor configuration file loading


christianparpart:
The current way of loading the config file has been incrementally growing, and thus, also its technical debt. It's time to clean the last(?) dark corner of the source code up.

### Main design goals

- [x] enable the user to have a minimal configuration file (all values should be defaulting)
- [ ] easily maintainable config file syntax (stick with yaml for now?, later toml/...?)
- [x] allow generating website-documentation out of the config file's key definitions
- [ ] allow easy upgrading, possible auto-updating config entries, if necessary/possible
- [x] allow auto-constructing a full non-defaulting config file template
- [x] integrate auto-generated website into ghpages CI hook

### Config file syntax

At first I think it should remain YAML, but i'm not hard-lining here.
Later on the following come into my mind, iff it turns out to be an all-inclusive upgrade:

- YAML: stick with it, known already, but some might not like forced indentation
- TOML: looks like a mix of yaml and ini, so not too alienating, but easy to understand
- LUA: do we really want to go for a more dynamic config language to be more powerful?

Yaraslaut:
Info about missing entries implemented in https://github.com/contour-terminal/contour/pull/1585
At the moment, we do not upgrade user config file to preserve comments and if implemented in simplest way upgrade will cause user profiles to specify all values while we allow to set specific values in the profile.  Implementation of complete upgrade will require some time and can be done in the future

Yaraslaut:
At the moment config loading let easily add config entries and we can generate default config and missing entries from executable, extending to other formats will lead to copy of read/write logic and can be postpones until reflection will appear in c++ (hopefully 26)

PR #1300, opened 2023-10-22
Refactor configuration loading

Yaraslaut:
Closes https://github.com/contour-terminal/contour/issues/1292
Refs https://github.com/contour-terminal/contour/issues/1282

### DIFF
--- a/src/contour/Config.h
+++ b/src/contour/Config.h
@@ -2,14 +2,21 @@
 #pragma once
 
 #include <contour/Actions.h>
+#include <contour/ConfigDocumentation.h>
 #include <contour/display/ShaderConfig.h>
 
 #include <vtbackend/Color.h>
 #include <vtbackend/ColorPalette.h>
+#include <vtbackend/ControlCode.h>
 #include <vtbackend/InputBinding.h>
+#include <vtbackend/InputGenerator.h>
+#include <vtbackend/MatchModes.h>
 #include <vtbackend/Settings.h>
+#include <vtbackend/VTType.h>
 #include <vtbackend/primitives.h> // CursorDisplay
 
+#include <vtpty/ImageSize.h>
+#include <vtpty/PageSize.h>
 #include <vtpty/Process.h>
 #include <vtpty/SshSession.h>
 
@@ -21,14 +28,35 @@
 
 #include <crispy/StrongLRUHashtable.h>
 #include <crispy/assert.h>
+#include <crispy/flags.h>
+#include <crispy/logstore.h>
 #include <crispy/size.h>
+#include <crispy/utils.h>
+
+#include <fmt/core.h>
+
+#include <yaml-cpp/emitter.h>
+#include <yaml-cpp/node/detail/iterator_fwd.h>
+#include <yaml-cpp/ostream_wrapper.h>
+#include <yaml-cpp/yaml.h>
+
+#include <range/v3/range/conversion.hpp>
+#include <range/v3/view/concat.hpp>
+#include <range/v3/view/join.hpp>
+#include <range/v3/view/transform.hpp>
 
 #include <chrono>
+#include <cstdint>
+#include <exception>
 #include <filesystem>
+#include <limits>
+#include <memory>
 #include <optional>
+#include <regex>
 #include <set>
 #include <string>
 #include <system_error>
+#include <type_traits>
 #include <unordered_map>
 #include <variant>
 
@@ -101,6 +129,7 @@ namespace helper
                && testMatchMode(actualModeFlags, expected, Flag::Search)
                && testMatchMode(actualModeFlags, expected, Flag::Trace);
     }
+
 } // namespace helper
 
 template <typename Input>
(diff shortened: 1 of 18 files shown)


src/contour/Config.h
```diff
@@ -144,138 +150,332 @@ struct SimpleColorConfig
...
+        // Defaults to 1 MB, that's roughly 10k lines when column count is 100.
+        size_t ptyBufferObjectSize = 1024lu * 1024lu;
+
+        bool reflowOnResize = true;
+
+        std::unordered_map<std::string, vtbackend::ColorPalette> colorschemes;
+        std::unordered_map<std::string, TerminalProfile> profiles;
+        std::string defaultProfileName;
+
+        TerminalProfile* profile(std::string const& name)
```
christianparpart, line 281:
```suggestion
        TerminalProfile* profile(std::string const& name) noexcept
```

src/contour/Config.h
```diff
@@ -144,138 +150,332 @@ struct SimpleColorConfig
...
+        {
+            assert(!name.empty());
+            if (auto i = profiles.find(name); i != profiles.end())
+                return &i->second;
+            assert(false && "Profile not found.");
+            return nullptr;
+        }
+
+        TerminalProfile const* profile(std::string const& name) const
```
christianparpart, line 290:
```suggestion
        TerminalProfile const* profile(std::string const& name) const noexcept
```

src/contour/Config.h
```diff
@@ -144,138 +150,332 @@ struct SimpleColorConfig
...
+    ~ConfigEntry() = default;
 
-    // Configures the size of the PTY read buffer.
-    // Changing this value may result in better or worse throughput performance.
-    //
-    // This value must be integer-devisable by 16.
-    size_t ptyReadBufferSize = 16384;
+  private:
+    std::string _documentation;
+    T _val;
```
christianparpart, line 366:
```suggestion
    T _value;
```
let's please avoid abbreviations, if possible :)

src/contour/Config.h
```diff
@@ -144,138 +150,332 @@ struct SimpleColorConfig
...
+    [[nodiscard]] constexpr T const& get() const { return _val; }
+    [[nodiscard]] constexpr T& get() { return _val; }
 
-    // Number of hashtable slots to map to the texture tiles.
-    // Larger values may increase performance, but too large may also decrease.
-    // This value is rounted up to a value equal to the power of two.
-    //
-    // Default: 4096
-    crispy::strong_hashtable_size textureAtlasHashtableSlots = crispy::strong_hashtable_size { 4096 };
+    [[nodiscard]] constexpr std::string makeDoc() const
```
christianparpart, line 353:
```suggestion
    [[nodiscard]] constexpr std::string makeDocumentation() const
```

src/contour/Config.h
```diff
@@ -144,138 +150,332 @@ struct SimpleColorConfig
...
+    {
+        return fmt::format(fmt::runtime(_documentation), _val);
```
christianparpart, line 355:
```suggestion
        return fmt::format(fmt::runtime(_documentation), _value);
```

nice, TIL (`fmt::runtime...` :) )

src/contour/Config.h
```diff
@@ -144,138 +150,332 @@ struct SimpleColorConfig
...
+
+        std::set<std::string> experimentalFeatures;
+    };
+} // namespace old
+
+template <typename T>
+struct ConfigEntry
 {
-    std::filesystem::path backingFilePath;
+    using val_type = T;
```
christianparpart, line 335:
```suggestion
    using value_type = T;
```

I noticed this a bit more often, but only clicked on it _some_ times. But let's generally try to avoid abbreviations, unless there's a good reason :)

src/contour/Config.h
```diff
@@ -144,138 +150,332 @@ struct SimpleColorConfig
...
+};
 
-    // Size in bytes per PTY Buffer Object.
-    //
-    // Defaults to 1 MB, that's roughly 10k lines when column count is 100.
-    size_t ptyBufferObjectSize = 1024lu * 1024lu;
+template <typename T>
+concept con_configEntry = requires(T t) {
+    t.makeDoc();
+    t.val;
```
christianparpart, line 372:
```suggestion
    t.makeDocumentation();
    t.value;
```

christianparpart (commented):
first bits of thoughts

Yaraslaut:
refs https://github.com/contour-terminal/contour/issues/1282

src/contour/Config.h
```diff
@@ -144,263 +189,3020 @@ struct SimpleColorConfig
...
+};
 
-    bool autoScrollOnUpdate;
+template <typename T>
+struct ConfigEntry
+{
+    using value_type = T;
 
-    vtrasterizer::FontDescriptions fonts;
+    constexpr ConfigEntry(string d): _value {}, _documentation { std::move(d) } {}
```
christianparpart, line 204:
```suggestion
    constexpr ConfigEntry(std::string d): _value {}, _documentation { std::move(d) } {}
```

src/contour/Config.h
```diff
@@ -144,263 +189,3020 @@ struct SimpleColorConfig
...
 
-    struct
+    constexpr ConfigEntry(T&& in, string d): _value { std::forward<T>(in) }, _documentation { std::move(d) }
```
christianparpart, line 206:
```suggestion
    constexpr ConfigEntry(T&& in, std::string d): _value { std::forward<T>(in) }, _documentation { std::move(d) }
```

I'm not going to comment that now everywhere below. This is most likely because you imported std::string into the current namespace. The problem I see here is, that this is a public header, and that should not have happened - IMHO at least :)

src/contour/Config.h
```diff
@@ -144,263 +189,3020 @@ struct SimpleColorConfig
...
+    T _value;
 };
 
-enum class RenderingBackend
-{
-    Default,
-    Software,
-    OpenGL,
+template <typename T>
+concept con_configEntry = requires(T t) {
```
christianparpart, line 232:
```suggestion
concept ConfigEntryConcept = requires(T t) {
```
Maybe we can give it a little better name, with fewer abbreviations? 🤔 hmmmm

src/contour/Config.h
christianparpart: NB: initialization order is wrong

src/contour/Config.h
```diff
@@ -144,263 +189,3020 @@ struct SimpleColorConfig
...
+                                                "#   - VT320\n"
+                                                "#   - VT420\n"
+                                                "#   - VT510\n"
+                                                "#   - VT520\n"
+                                                "#   - VT525\n"
+                                                "terminal_id: {}\n"
+                                                "\n" };
+    ConfigEntry<vtbackend::MaxHistoryLineCount> maxHistoryLineCount {
+        vtbackend::LineCount(1000),
+        "# Number of lines to preserve (-1 for infinite).\n"
```
christianparpart, line 328: can we get the syntax of a potential serialization format (like YAML) out of the documentation?

The config file writer, that is about to write the comments above each key, should make sure it's proper syntax in the target format, like YAML. so a comment like `"this makes foo"` should become `"    # this makes foo\n"` with leading spaces to do the indentation proper, and `#` to introduce the YAML comment. the last line of a multi line comment should not require to have an LF. that's easy to add.

This makes it free for us to exchange YAML in the future with something else, say, ... LUA? or TOML etc :)

src/contour/Config.h
```diff
@@ -144,263 +189,3020 @@ struct SimpleColorConfig
...
+    }
+    void load(Config& c)
+    {
+        loadFromEntry("platform_plugin", c.platformPlugin);
+        // TODO(pr)
+        if (c.platformPlugin.get() == "auto")
+        {
+            c.platformPlugin.get() = "";
+        }
+        loadFromEntry("default_profile", c.defaultProfileName);
```
christianparpart, line 2374: I think it should be possible to encode the key name into the `ConfigEntry` to not just hold the entry type `T`.

`using ThatConfigEntry = ConfigEntry<That, "that">;` 

Having a look at https://ctrpeach.io/posts/cpp20-string-literal-template-parameters/ - we should probably be able to make it. :)

src/contour/Config.h
```diff
@@ -144,263 +189,3020 @@ struct SimpleColorConfig
...
+        process(c.live);
+        process(c.experimentalFeatures);
+
+        // inside images:
+        doc.append("\nimages: \n");
+        level++;
+        process(c.sixelScrolling);
+        process(c.maxImageColorRegisters);
+        process(c.maxImageSize);
+        level--;
```
christianparpart, line 2458: maybe we can use some RAII helper to avoid `x++; ...; x--;` that would be a bit more error prone.

src/contour/Config.h
```diff
@@ -144,263 +189,3020 @@ struct SimpleColorConfig
...
+struct YAMLConfigWriter
+{
+
+    std::string addOffset(string doc, size_t off)
+    {
+        auto offset = std::string(off, ' ');
+        return std::regex_replace(doc, std::regex(".+\n"), offset + "$&");
+    }
+
+    string createString(Config& c)
```
christianparpart, line 2416:
```suggestion
    string createString(Config const& c)
```

src/contour/Config.h
```diff
@@ -144,263 +189,3020 @@ struct SimpleColorConfig
...
+        loadFromEntry("profiles", c.profiles);
+
+        loadFromEntry("color_schemes", c.colorschemes);
+    };
+};
+
+struct YAMLConfigWriter
+{
+
+    std::string addOffset(string doc, size_t off)
```
christianparpart, line 2410:
```suggestion
    std::string addOffset(string const& doc, size_t off)
```

christianparpart (commented):
some more bits of looking through :)

src/contour/Config.h
Yaraslaut: Hm, TOML has the same comment character, LUA does not, maybe we can add this generalization in the future, what do you think?

Yaraslaut: but we should not apply this is to every line, for example here:
```
"# whether or not to hide the scrollbar when in alt-screen.\n"
"hide_in_alt_screen: {}\n"
 ```
 we need to add comment only in first line, for that we need to have something like 
```
"{comment} whether or not to hide the scrollbar when in alt-screen.\n"
"hide_in_alt_screen: {}\n"
 ```
 and then format it with `fmt::arg("comment", "#")` maybe its not that difficult so i can add it then

christianparpart: I think defaults should not be mentioned in C++'s documentation at all. In fact, that could be inferred and auto-generated in the docs, IMHO. No need to encode that into the documentation field, if it's already expressed in the C++ struct. That should work, shouldn't it?

Yaraslaut: > I think defaults should not be mentioned in C++'s documentation at all. In fact, that could be inferred and auto-generated in the docs, IMHO. No need to encode that into the documentation field, if it's already expressed in the C++ struct. That should work, shouldn't it?

You are right, better not place default values there explicitly, will fix it, default construction of config gives exactly the default config that we have in `contour.yml` file

src/crispy/flags.h
```diff
@@ -150,7 +150,7 @@ struct fmt::formatter<crispy::flags<Enum>>: public fmt::formatter<std::string>
                 continue;
 
             if (!result.empty())
-                result += '|';
+                result += ',';
```
christianparpart, line 153: was there a reason why you did it or had to do this? If you rely on that in the YAML output, than we should not use the fmt formatter here to encode for the YAML output. this is formatting only for human readability. :)

Yaraslaut: This is not needed anymore, we wrap format inside yaml writer

christianparpart (approved):
I am really curious now. Many thanks, man :)

PR #1328, opened 2023-11-07
Improve Numpad input mapping (relevant for CSIu protocol)

### LINKED ISSUE #1325: Ctrl+Keypad Plus seem not to work
### Contour Terminal version

0.4.0.r4132.31248d53

### Installer source

Github: source code cloned

### Operating System

Arch

### Architecture

x86-64

### Other Software

none

### Steps to reproduce

I have the following lines in my input mapping:
    - { mods: [Control],        key: Plus,                action: IncreaseFontSize }
    - { mods: [Control],        key: Minus,             action: DecreaseFontSize }

Press ctrl + (Numpad) Plus or ctrl + (Numpad) Minus

### Expected Behavior

Resizing the Font Size 

### Actual Behavior

Nothing

### Additional notes

Works with the regular Minus and Plus and worked with the last version i  used.

christianparpart:
Kinda hard to test and verify blindingly - while not having a keyboard that is having a keypad, but I think this is going into the right direction.

The problem of keypad not working was introduced by implementing CSIu, which strongly differenciates between numbers and numbers typed in from numpad.

## EDIT:

Now, also fixes #1325!

NB: No changelog needed because this problem was introduced post last release and only affects CSIu (new)

### DIFF
--- a/src/contour/Config.cpp
+++ b/src/contour/Config.cpp
@@ -594,6 +594,23 @@ namespace
             pair { "PrintScreen"sv, Key::PrintScreen },
             pair { "Pause"sv, Key::Pause },
             pair { "Menu"sv, Key::Menu },
+            pair { "Numpad_0"sv, Key::Numpad_0 },
+            pair { "Numpad_1"sv, Key::Numpad_1 },
+            pair { "Numpad_2"sv, Key::Numpad_2 },
+            pair { "Numpad_3"sv, Key::Numpad_3 },
+            pair { "Numpad_4"sv, Key::Numpad_4 },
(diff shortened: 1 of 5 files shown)


Yaraslaut:
I can check it

Yaraslaut (changes requested):
I checked that it works now. (Was checking numpad inside nvim)
bug  that i noticed is that if i press numpad key inside `normal` mode it gets written to the prompt with output from contour:
```
[error] ViInputHandler: Unhandled key: Numpad_5 ()
[error] ViInputHandler: Unhandled key: Numpad_6 ()
```

PR #1337, opened 2023-11-16
Fixes setting window class (`app_id`) on Wayland

### LINKED ISSUE #1336: setting window class doesn't work on wayland
### Contour Terminal version

0.3.1.200/0.3.12.262/0.4.0.6089

### Installer source

### Operating System

Nixos unstable

### Architecture

x86-64

### Other Software

hyprland 0.32.0

### Steps to reproduce

either set the wm_class via `profiles.<profilename>. wm_class` or pass it via `contour class`

### Expected Behavior

`hyprctl clients` should display the new class

### Actual Behavior

Nothing changes, although the result is slightly different on the old/stable and the version obtained from github actions. In the former case class is set to `contour`; on the git version it is simply goes blank if specified via an argument, and doesn't change when specified in the configuration.

### Additional notes

Regarding the versions, the oldest one was obtained from nixpkgs (built from source), the newer versions were "repackaged" from .debs for ubuntu 22.04 (obtained from releases and github actions' artifacts).

christianparpart:
On Wayland, there's no such thing as WM_CLASS (X11), but a user can use the `app_id` to identify and group Application windows. However, Qt sets that not via QCoreApplication::setApplicationName but rather uses QCoreApplication::setOrganizationDomain's information in reverse name order.

Closes #1336.

### DIFF
--- a/src/contour/ContourGuiApp.cpp
+++ b/src/contour/ContourGuiApp.cpp
@@ -312,10 +312,6 @@ int ContourGuiApp::terminalGuiAction()
         errorLog()("Could not access configuration profile.");
         return EXIT_FAILURE;
     }
-    auto appName = QString::fromStdString(profile->wmClass);
-    QCoreApplication::setApplicationName(appName);
-    QCoreApplication::setOrganizationName("contour");
-    QCoreApplication::setApplicationVersion(CONTOUR_VERSION_STRING);
 
     vector<string> qtArgsStore;
     vector<char const*> qtArgsPtr;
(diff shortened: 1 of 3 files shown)


src/contour/ContourGuiApp.cpp
```diff
@@ -445,6 +443,48 @@ int ContourGuiApp::terminalGuiAction()
...
+    }
+    else
+    {
+        QCoreApplication::setOrganizationDomain(defaultOrgDomain);
+        QCoreApplication::setOrganizationName(defaultOrgName);
+        QCoreApplication::setApplicationName(defaultAppName);
+        QCoreApplication::setApplicationVersion(CONTOUR_VERSION_STRING);
+    }
+}
+
```
uspasojevic96, line 487: We don't need to duplicate 4 same lines for each if statement here

within xcb if check i would set appName to wmClass if it exists

within wayland check i would set desktop file name if wmClass exists

everything else should just default i believe

uspasojevic96 (changes requested):

src/contour/ContourGuiApp.cpp
christianparpart: yeah you're right. at first I didn't know how to approach it, so i was writing it a little bit more naively. thanks ;)

uspasojevic96 (approved):
Lgtm

PR #1345, opened 2023-11-26
Fix scroll freeze

Yaraslaut:
Fix scroll freeze due to error in conversion from float to int while 532.9999 -> 532


christianparpart (approved):
Many thanks for investigating this! :-)

Issue #1350, opened 2023-12-01
error: cannot bind non-const lvalue reference of type ‘int&’ to an rvalue of type ‘int’ with fmt 10


topazus:
### Contour Terminal version

6df0a2f9ef83e0b8666a26403115f769073ac435

### Installer source

Github: source code cloned

### Operating System

Fedora 40

### Architecture

x86-64

### Steps to reproduce

<details>
<summary>build log</summary>

```bash
❯ git log
commit 6df0a2f9ef83e0b8666a26403115f769073ac435 (grafted, HEAD -> master, origin/master, origin/HEAD)
Author: Christian Parpart <christian@parpart.family>
Date:   Mon Nov 27 22:54:24 2023 +0100

    Merge pull request #1347 from contour-terminal/improvement/image-impl-cleanup
    
    Minor cleanup on how to instanciate a RasterizedImage

❯ rm -rf build/ && cmake . -B build/ -GNinja
-- [Version] Git branch: master
-- [Version] Git SHA: 6df0a2f
-- [Version] version source: git & /home/ruby/test/contour/metainfo.xml
-- [Version] version triple: 0.4.0
-- [Version] version string: 0.4.0-master-6df0a2f
-- The CXX compiler identification is GNU 13.2.1
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/lib64/ccache/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- [ccache] Enabled: /usr/bin/ccache
-- [clang-tidy] Disabled.
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- addr2line: /usr/bin/addr2line
-- No sanitizer enabled.
-- No 3rdparty libraries found at /home/ruby/test/contour/_deps/sources
-- Found HarfBuzz: /usr/lib64/libharfbuzz.so  
-- Found Freetype: /usr/lib64/libfreetype.so (found version "2.13.1") 
-- Found libunicode, version: 0.4.0
-- Found boxed-cpp, version: 1.0.1
-- Performing Test fdiagnostics-color=always
-- Performing Test fdiagnostics-color=always - Success
-- Adding compiler flag: -fdiagnostics-color=always.
-- Enabling pedantic compiler options: no
-- Looking for backtrace
[...]

PR #1352, opened 2023-12-01
Fix fmt compilation.

Yaraslaut:
from fmt release page :  
https://github.com/fmtlib/fmt/releases/tag/10.1.0
```
Disallowed passing temporaries to make_format_args to improve API safety by preventing dangling references.
```
closes https://github.com/contour-terminal/contour/issues/1350

### DIFF
--- a/src/vtbackend/Terminal.h
+++ b/src/vtbackend/Terminal.h
@@ -631,7 +631,7 @@ class Terminal
 #if defined(__APPLE__) || defined(_MSC_VER)
         reply(fmt::vformat(message, fmt::make_format_args(args...)));
 #else
-        reply(fmt::vformat(message, fmt::make_format_args(std::forward<Ts>(args)...)));
+        reply(fmt::vformat(message, fmt::make_format_args(args...)));
 #endif
     }
 


src/vtbackend/Terminal.h
```diff
@@ -631,7 +631,7 @@ class Terminal
 #if defined(__APPLE__) || defined(_MSC_VER)
         reply(fmt::vformat(message, fmt::make_format_args(args...)));
 #else
-        reply(fmt::vformat(message, fmt::make_format_args(std::forward<Ts>(args)...)));
+        reply(fmt::vformat(message, fmt::make_format_args(args...)));
```
christianparpart, line 634: that is interesting. I thought it's usually a good idea to do perfect forwarding. but here it maybe not the case. hmmm. ok

christianparpart (approved):

src/vtbackend/Terminal.h
Yaraslaut: So the idea is that we want to prevent rvalue reference there, so that is why we can not forward things there anymore, few issues from fmt repo: https://github.com/fmtlib/fmt/issues/3443 , https://github.com/fmtlib/fmt/issues/3589

christianparpart: ok thx

PR #1362, opened 2023-12-16
Bump boxed-cpp to version 1.1.0

christianparpart:
such that future contour release depends on a fixed release of boxed-cpp (1.1.0)


Yaraslaut (approved):
Great :)

PR #1363, opened 2023-12-18
Update linksafe action

Yaraslaut:
Update linksafe


src/vtrasterizer/DecorationRenderer.cpp
```diff
@@ -185,7 +185,7 @@ auto DecorationRenderer::createTileData(Decorator decoration, atlas::TileLocatio
                 for (int x = 0; x < unbox<int>(width); ++x)
                 {
                     // Using Wu's antialiasing algorithm to paint the curved line.
-                    // See: https://www-users.mat.umk.pl//~gruby/teaching/lgim/1_wu.pdf
+                    // See: https://dl.acm.org/doi/pdf/10.1145/127719.122734
```
christianparpart, line 188: wow. I think it's good to have a link-checker. So we know as early as possible when links become broken :)

christianparpart (approved):

PR #1365, opened 2023-12-18
Fix window resize events not respecting content scale 

### LINKED ISSUE #1299: DPI adjustment at least sometimes broken
At least I noticed that on MacOS, where I am clearly having an HiDPI screen (192x192 DPI) and I frequently attach/detach/re-attach a different screen with a different DPI value.

What should happen is, that the font size remains the same, and thus, the cell dimensions are not touched at all.

What does happen is, that the front size is either crazy big or crazy small.

This also leads to bad calculation of mouse position to grid cell mapping.

I am experiencing this issue on MacOS/X the most, but it's most likely an issue on any platform, which can adapt to DPI changes live (e.g. KDE Plasma / Windows)

christianparpart:
this is mostly affecting HiDPI screens, anything with a non-standard DPI setting.

no changelog entry needed, because it was introduced with the QML rework

Closes #1299.

### DIFF
--- a/src/contour/TerminalSession.cpp
+++ b/src/contour/TerminalSession.cpp
@@ -696,7 +696,8 @@ void TerminalSession::requestWindowResize(LineCount lines, ColumnCount columns)
 
 void TerminalSession::requestWindowResize(QJSValue w, QJSValue h)
 {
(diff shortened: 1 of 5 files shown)


Yaraslaut (approved):
Nice, hope we can release now

PR #1370, opened 2023-12-19
[scripts] Fix OS misdetection in install-deps.sh for FreeBSD

herrhotzenplotz:
The script used to say that FreeBSD is unsupported despite there
being an explicit case for it.

Fix by correcting the casing.


Yaraslaut:
Thanks for fixing it

christianparpart (approved):
Many thanks, @herrhotzenplotz :)

PR #1376, opened 2023-12-20
Fix collision of bold and regular font files

Yaraslaut:
Some fonts do not provide separate files for regular and bold weight but use the same, this make impossible to distinguish this two possibilities when comparing file paths and font size

### DIFF
--- a/src/text_shaper/open_shaper.cpp
+++ b/src/text_shaper/open_shaper.cpp
@@ -16,6 +16,7 @@
 
 #include <limits>
 #include <optional>
+#include <string>
 
 // clang-format off
 #include <ft2build.h>
@@ -66,15 +67,16 @@ using namespace std::string_view_literals;
 namespace
 {
 
-struct FontPathAndSize // NOLINT(readability-identifier-naming)
+struct FontInfo // NOLINT(readability-identifier-naming)
 {
     string path;
     text::font_size size;
+    text::font_weight weight;
 };
 
-[[maybe_unused]] bool operator==(FontPathAndSize const& a, FontPathAndSize const& b) noexcept
+[[maybe_unused]] bool operator==(FontInfo const& a, FontInfo const& b) noexcept
 {
-    return a.path == b.path && a.size.pt == b.size.pt;
+    return a.path == b.path && a.size.pt == b.size.pt && a.weight == b.weight;
 }
 
 } // namespace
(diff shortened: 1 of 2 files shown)


classabbyamp:
tested this applied on top of the tagged 0.4.0 release, appears to work.

0x5c:
Tried the same, works perfectly

src/text_shaper/open_shaper.cpp
```diff
@@ -423,9 +423,6 @@ struct open_shaper::Private // {{{
     optional<font_key> getOrCreateKeyForFont(font_source const& source, font_size fontSize)
     {
         auto const sourceId = identifierOf(source);
-        if (auto i = fontPathAndSizeToKeyMapping.find(FontPathAndSize { sourceId, fontSize });
-            i != fontPathAndSizeToKeyMapping.end())
-            return i->second;
```
christianparpart, line 428: Myconcern I am having (right now) is, that it's not a get-or-create anymore, but you'll ALWAYS create a new key, right?

Maybe the cache key is wrong then. If a font file can contain more than one face (or weight), but a font_key should represent a specific "font face", then we need to encode the weight into it too.

Why I didn't do that so far is, because (to my knowledge), a font file can only contain one weight (and style). 🤔

classabbyamp: > Why I didn't do that so far is, because (to my knowledge), a font file can only contain one weight (and style). 🤔

[variable fonts](https://en.wikipedia.org/wiki/Variable_font) are fairly new, and that is usually a reasonable assumption.

0x5c: Variable fonts are fonts that support arbitrary weights though interpolation, but the problem also happens with fonts distributed as TrueType collections (usually `.ttc`), like Iosevka.

Yaraslaut: yes, so basically we need to include calculation of weight in the font_face or just renema this function and create new key always

0x5c: > include calculation of weight in the font_face

Would that work with TrueType Collections too? Typically those contain normal styles and fixed weights, just bundled in one file for file size reasons

Yaraslaut: Yes, we can distinguish between them, at the moment we only use font name and size to check if we already loaded this font, so we 
 need to add font weight into comparison

Yaraslaut: So i updated PR and now we check for caching but taking into account weight of the font

christianparpart (approved):
In the future, we could extend that over font style (like italics) or weight value for variable width fonts (values between 100, 200, ..., 900 - IIUIC).

PR #1388, opened 2023-12-26
contour.yml: fix incorrect default for frozen dev modes

0x5c:
I assume the proper to do this is to comment out the key, which works without error messages.
I initially tried leaving the key with no value (based on the comment) but that caused `[error] Invalid frozen_dec_modes entry.`. Using a value of `{}` worked but seemed more confusing than anything.

### DIFF
--- a/src/contour/contour.yml
+++ b/src/contour/contour.yml
@@ -276,8 +276,7 @@ profiles:
         #         2027: true
         #
         # Default: (empty object)
-        frozen_dec_modes:
-            2026: false
+        #frozen_dec_modes:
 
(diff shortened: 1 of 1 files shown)


christianparpart (changes requested):
Many thanks.

christianparpart (approved):
Smartphone error... 🌲

0x5c:
I just now realise that I wrote `frozen dev modes` in the commit message haha

christianparpart:
Not the worst thing that ever happened. :)

PR #1391, opened 2023-12-27
Properly manage scrollbar and offset

Yaraslaut:
Fixes mistake with rounding of scrollbar position and loop dependency between scrollbar position and offset


christianparpart (approved):
Many thanks 🎄

PR #1392, opened 2023-12-27
Attempting to fix assertion failure when running in RHEL Linux

### LINKED ISSUE #1384: Support for a margin around the terminal area
## Abstract

Currently, contour draws all cells directly at the window border. This makes the content of the terminal very cramped to the edges, especially in small font sizes.

## Motivation

This feature would likely take form of a setting allowing the user to specify a minimum size in px of margin between the window border (or screen border for fullscreen) and the character cells.
The margin's background would be the default background colour.

For reference, see Konsole's margin feature

christianparpart:
This is probably related to building using Qt 5 rather than Qt 6.

This should fix #1384's comment on failing to run Contour on RHEL. I think maybe it's building against Qt5 rather than Qt6 on that platform, and that Qt5 us inducing somewhat different execution order on Contour-side. 🤔

### DIFF
--- a/src/contour/helper.cpp
+++ b/src/contour/helper.cpp
@@ -598,8 +598,8 @@ void applyResize(vtbackend::ImageSize newPixelSize,
     vtbackend::Terminal& terminal = session.terminal();
     vtbackend::ImageSize cellSize = renderer.gridMetrics().cellSize;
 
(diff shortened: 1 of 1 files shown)


Yaraslaut (approved):
LGTM, 
thanks for fixing it

Issue #1360, opened 2023-12-11
Extra fast scrolling when using trackpad


0x5c:
### Contour Terminal version

Contour Terminal Emulator 0.4.0

### Installer source

Github: source code cloned

### Operating System

VoidLinux

### Architecture

x86-64

### Other Software

Built with Qt 6

### Steps to reproduce

Open contour, use trackpad to scroll

### Expected Behavior

The rate of scroll events matches other terminals, and the scrolling isn't jittery (no flashing through up/down/up/down/...)

### Actual Behavior

The rate of scroll events is too fast, to the point of being mostly unusable. The scroll events are also frequently alternating up/down fast when the movement of the fingers on the trackpad is at the limit of the trackpad's sensing precision.

### Additional notes

This also happens with the latest release (`v0.3.12.262`).
Normal mouse scrollwheels don't seem to be affected. The trackpoint on my laptop might also be affected but it's much harder to tell.

christianparpart:
@0x5c: we've just merged a fix to master. Is it possible for you to confirm that it is also working for **you**?

Otherwise, we're about to release ASAP 0.4.1 as well.

0x5c:
I can confirm that the fix works perfectly, applied as a patch on top of 0.4.0 along with the margins one.

PR #1393, opened 2023-12-29
Accumulate scroll distance instead of immediately trigger scroll

Yaraslaut:
Closes https://github.com/contour-terminal/contour/issues/1360

Qt backend sends a lot of mouse events on scroll with touchpad, this PR introduce time restiction between mouse events that we apply to terminal, this must affect only scroll on touchpads since all other manuall events can not be separated in time for such small duration as 50 ms

### DIFF
--- a/metainfo.xml
+++ b/metainfo.xml
@@ -107,6 +107,7 @@
     <release version="0.4.1" urgency="medium" type="development">
       <description>
         <ul>
+          <li>Fixes fast scrolling when using trackpad (#1360)</li>
           <li>Fixes variable fonts loading</li>
           <li>Fixes Command modifier for input mappings, such as Command+C or Command+V on on MacOS (#1379).</li>
           <li>Fixes CSIu encoding of shift modifier produced characters (#1373).</li>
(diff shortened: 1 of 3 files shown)


src/contour/TerminalSession.cpp
```diff
@@ -826,6 +828,15 @@ void TerminalSession::sendMousePressEvent(Modifiers modifiers,
     inputLog()("Mouse press received: {} {}\n", modifiers, button);
 
     terminal().tick(steady_clock::now());
+    auto const passed =
+        std::chrono::duration_cast<std::chrono::milliseconds>(_terminal.currentTime() - _previousMouseEvent);
+
+    if (passed.count() < 50)
```
christianparpart, line 834: This feels like we really want that `50` configurable. Because most likely I'd like to set it to `0` and maybe that also depends on the frequency of mouse event reporting on the OS side that Qt is blindingly forwarding (just guessing, but I know I'm having such a gamer mouse that can report in different frequencies)

Yaraslaut: I agree that we should make it configurable, maybe we can postpone it until config refector PR? at the moment this number is quite arbitrary and based only on my machine, but i would not expect actual mouse to show this behaviour since usually there is a tick mechanism there to trigger scroll event

christianparpart (approved):
Many thanks, @Yaraslaut 🎄

PR #1394, opened 2023-12-31
docs: fixed markup for backticks text object

0x5c:
Markdown has a lot of weird obscure corners; TIL there's a specific markup for inline code that can contain backticks.

## Motivation and Context

image


christianparpart (approved):
Oh yeah, many thanks for fixing it, @0x5c! Happy new year!

PR #1400, opened 2024-01-06
Fix Alt+WheelEvents not captured correctly

### LINKED ISSUE #394: Alt+WheelEvents not captured correctly
Related Qt-BUG tracking ticket: https://bugreports.qt.io/browse/QTBUG-30948

Thing is, that behavior seems intended by Qt, but I explicitly do NOT want that cleverness. I want the actual, unaltered(!) information. With `orientation()` being dropped in Qt 6 there is literally no way (as far as I can tell) to detect that.

Not sure how to fix this unless someone in the related tracker has a clever answer for me.

https://github.com/qutebrowser/qutebrowser/issues/3840 seems to suffer from the same problem and didn't have a fix just yet (?).

christianparpart:
closes #394.

### DIFF
--- a/metainfo.xml
+++ b/metainfo.xml
@@ -108,6 +108,7 @@
       <description>
         <ul>
           <li>Fixes some text run segmentation problems, e.g., not making use of some programming ligatures (#395).</li>
+          <li>Fixes mouse scroll wheel events going into the wrong direction if `Alt` modifier was pressed at the same time (#394).</li>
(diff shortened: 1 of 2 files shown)


Yaraslaut (approved):
Lgtm, great that we fixed it. One unfortunate thing is that I cannot find documentation for this behavior on the Qt side. What they mention here, https://bugreports.qt.io/browse/QTBUG-30948 not very helpful  :(

Issue #1402, opened 2024-01-07
CI tests for GUI (notcurses-demo) are flaky again


christianparpart:
We need to fix them - again. I forgot what the problem was the last time, we've fixed it. Do you happen to remember, @Yaraslaut?

Yaraslaut:
Last time this was due to failing of notcurses to start, so we fixed it and added new action `GUI: Quick Shell Exit ` to check it

PR #1403, opened 2024-01-07
Fix notcurses github action

Yaraslaut:
It looks like actions are working well now, even though nothing significant has been changed, apart from the logging level for these tests. Locally, I notice that notucrses-demo can fail from time to time due to events like terminal resizing. I think it is sufficient for us to have this PR merged at the moment and if needed to investigate further how to deal with it. for example to build notcurses from source inside actions as @christianparpart  few times proposed 

Closes https://github.com/contour-terminal/contour/issues/1402

### DIFF
--- a/src/vtbackend/Terminal.cpp
+++ b/src/vtbackend/Terminal.cpp
@@ -265,10 +265,10 @@ bool Terminal::processInputOnce()
 
     if (!readResult)
     {
+        terminalLog()("PTY read failed. {}", strerror(errno));
         if (errno == EINTR || errno == EAGAIN)
             return true;
 
-        terminalLog()("PTY read failed. {}", strerror(errno));
         _pty->close();
         return false;
     }
--- a/scripts/ci/Xvfb-contour-run.sh
+++ b/scripts/ci/Xvfb-contour-run.sh
@@ -7,6 +7,7 @@ set -x
 LIBGL_ALWAYS_SOFTWARE="${LIBGL_ALWAYS_SOFTWARE:-true}"
 CONTOUR_BIN=${CONTOUR_BIN:-contour}
 LOG="error,config,pty,gui.session,gui.display,vt.renderer,font.locator" # "all"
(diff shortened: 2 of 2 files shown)


src/vtbackend/Terminal.cpp
```diff
@@ -263,12 +263,12 @@ bool Terminal::processInputOnce()
 
     auto const readResult = readFromPty();
 
-    if (!readResult)
+    if (!readResult.has_value())
```
christianparpart, line 266: this should be equivalent. I am surprised and wondering why you did that. :)

src/vtbackend/Terminal.cpp
```diff
@@ -263,12 +263,12 @@ bool Terminal::processInputOnce()
...
     {
+        terminalLog()("PTY read failed. {}", strerror(errno));
```
christianparpart, line 268: that's a good thing to log, but maybe for everything except `EINTR` and `EAGAIN`. Maybe 2 lines below? Or was it intentionally placed above the if-condition?

src/vtbackend/Terminal.cpp
Yaraslaut: i was thinking that pointer to bool conversion been happening, was it  internal conversion to bool from std::optional  side ?

src/vtbackend/Terminal.cpp
Yaraslaut: I used it for debug what is happening, i was not sure whether we were waiting for response or were infinitely looping somewhere else, so i would like to have it as a log ouput, maybe we can move it in another log tag ?

src/vtbackend/Terminal.cpp
Yaraslaut: Oh, I see, indeed it has operator bool()

Yaraslaut: Updated

Issue #1413, opened 2024-01-17
Selecting words after tabs is off by one


andrew-stevenson-frequenz:
### Contour Terminal version

0.4.2-master-20f76b90

### Installer source

Github: source code cloned

### Operating System

Arch Linux

### Architecture

x86-64

### Steps to reproduce

printf “\tword\n”

then double click on word. Note that the highlighted area is off by one. Pasting the value however does the right thing.

If you drag with your mouse to highlight what looks to be right then when you paste you miss the last character and get a leading space.

if you drag with your mouse, starting at the second character and overshooting by one then you can paste the whole string.

This seems to require a leading tab. 

### Actual Behavior

See above

### Additional notes

There are other issues with tab, e.g. sometimes they don’t get printed in the output of ‘ls’ so I wonder if there is an off by one error when calculating which column a tab puts you in. 
![Uploading image.jpg…]()

Yaraslaut:
bisect shows that https://github.com/contour-terminal/contour/pull/1401 broke it

cqexbesd:
I'm assuming this is the same issue so I will check again with Christian's PR but as I have already recorded it:

https://github.com/contour-terminal/contour/assets/6312149/4114c3d6-cc69-4572-a72f-70358959b911

Note that when searching the string I enter, apart from being misspelt, is one position too far to the left. Then when the search gets to `$io->flush`, `flush` is offset one space to the right. That didn't happen for the earlier matches and it goes away when moving to the next match.

Yaraslaut:
> I'm assuming this is the same issue so I will check again with Christian's PR but as I have already recorded it:
> contour.mp4
> 
> Note that when searching the string I enter, apart from being misspelt, is one position too far to the left. Then when the search gets to `$io->flush`, `flush` is offset one space to the right. That didn't happen for the earlier matches and it goes away when moving to the next match.

Yes, we expect it to be fixed in https://github.com/contour-terminal/contour/pull/1414 so if you can update us whether PR fixed this behaviour we can merge it asap

cqexbesd:
Yes I can confirm #1414 seems to fix everything I reported here.

Yaraslaut:
Close it as complete via #1414

PR #1414, opened 2024-01-17
Fixes text render bug that was just introduced

christianparpart:
FYI: @Marenz just in case if that was the reason for your rollback :)


Yaraslaut:
This PR closes https://github.com/contour-terminal/contour/issues/1413

christianparpart:
> I'll test it a bit

it works for me (tm), but please report anyways, either here or in Discord/etc. :)

PR #1417, opened 2024-01-20
Fix legacy keyboard input protocol not reporting non-Alt modifier

### LINKED ISSUE #1411: [Ctrl+arrow] no longer works in some contexts
### Contour Terminal version

Contour Terminal Emulator 0.4.1

### Installer source

### Operating System

VoidLinux

### Architecture

x86-64

### Other Software

zsh 5.9
micro 2.0.13
contour from voidlinux's package

### Steps to reproduce

In zsh, [Ctrl+V] [Up/Down/Left/Right] [Ctrl+V] [Ctrl+Up/Down/Left/Right]
The shell receives the same key sequence for both the unmodified and modified arrows. Shell keybinds that use the ctrl-modified arrow keys simply don't fire.

In micro's input debug mode (`raw` command), typing unmodified and ctrl-modified arrow keys shows the received sequences are different and correct.

### Expected Behavior

All programs can receive a ctrl-modified arrow key

### Actual Behavior

Some programs receive [Ctrl+Up/Down/Left/Right] as [Up/Down/Left/Right].

### Additional notes

Since micro does not use neither [`modifyOtherKeys`](https://github.com/zyedidia/micro/issues/2590) nor [Kitty's](https://github.com/zyedidia/micro/issues/1176) [CSI u](https://github.com/zyedidia/micro/issues/2868), the main difference seems to be either raw/normal mode or usage of the alternate buffer.

I was too busy with packaging contour, reporting bugs, testing patches, and requesting features between `0.4.0` and `0.4.1` to notice which one broke it.

christianparpart:
Fixes #1411. 🙏

### DIFF
--- a/metainfo.xml
+++ b/metainfo.xml
@@ -111,6 +111,7 @@
           <li>Fixes mouse scroll wheel events going into the wrong direction if `Alt` modifier was pressed at the same time (#394).</li>
           <li>Fixes `scripts/install-deps.sh` for openSuSE (Tumbleweed) to install the correct dependencies.</li>
           <li>Fixes missing dependencies for release .deb packages (#1397).</li>
+          <li>Fixes legacy keyboard input protocol not reporting non-Alt modifier (#1411).</li>
         </ul>
       </description>
     </release>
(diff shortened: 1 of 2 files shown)


0x5c:
Can confirm, this is a successful fix when applied as a patch on top of 0.4.1

PR #1422, opened 2024-01-24
Small cleanup of cmake

Yaraslaut:
Small cleanup of cmake files

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,7 +1,3 @@
-if(("${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU") OR ("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang"))
-    #(TODO: first get rid of these warnings) add_compile_options(-Werror)
-endif()
-
 include(PedanticCompiler)
 
 add_subdirectory(crispy)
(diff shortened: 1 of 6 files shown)


Yaraslaut:
I realized that we use older versions of cmake in some distros and can not use target sources with FILE_SET yet, thus closing PR

christianparpart:
@Yaraslaut would it be possible to upgrade cmake on such platforms - for CI?

IMHO it should be all good for any supported platform as long as we can build for that platform **on CI**.

Ubuntu 18.04 also doesn't have a proper C++20 aware compiler out of the box, and yet we compile using one for Ubuntu 18.04 as well. I would be fine with using a more recent cmake as long as it doesn't break CI **and** existing downstream distributions (SuSE, Fedora, Void, NixOS). :)

Yaraslaut:
> @Yaraslaut would it be possible to upgrade cmake on such platforms - for CI?
> 
> IMHO it should be all good for any supported platform as long as we can build for that platform **on CI**.
> 
> Ubuntu 18.04 also doesn't have a proper C++20 aware compiler out of the box, and yet we compile using one for Ubuntu 18.04 as well. I would be fine with using a more recent cmake as long as it doesn't break CI **and** existing downstream distributions (SuSE, Fedora, Void, NixOS). :)

i think this is pointless at the moment to use newer versions of cmake  since we do not get any benefits and potentially loosing support of some distros

Issue #1408, opened 2024-01-11
key mappings have stopped working


T-727:
### Contour Terminal version

0.4.1.6292

### Installer source

something else (please specify below)

### Operating System

Windows 11 Home 23H2

### Architecture

x86-64

### Steps to reproduce

- Open contour
- Use any keymap (e.g Ctrl+Shift+Space for vi-mode)
- See that nothing happens

### Actual Behavior

According to the logs they do get detected:
```shell
Debug[default]: sendKeyEvent QKeyEvent(KeyPress, Key_Space, ShiftModifier|ControlModifier, text=" ") ((null):0, (null))
```
but absolutely nothing happens after

### Additional notes

- last working version for me is `0.3.12.262`
- tried both with default configuration and my own

I know this isn't the most helpful report but I can't find anything to explain this 😅

Yaraslaut:
Maybe something is wrong with config file, you can try `contour debug config` and  check the output

T-727:
That was my first thought, hence why I made sure it happens with the default config (`contour generate config`).

I tried `contour debug config` now with the default config and there doesn't seem to be anything interesting:
![image](https://github.com/contour-terminal/contour/assets/74924917/f52624cf-8003-41f0-9072-3022768f9fa6)

Yaraslaut:
Alright, can you try then 
`contour debug "gui.input,gui.session"`
this is what i see for ctr+shift+shape
```
Debug[default]: sendKeyEvent QKeyEvent(KeyPress, Key_Shift, ShiftModifier) ((null):0, (null))
[2024-01-11 17:24:37.336898549.336898] [gui.input] Input not handled for mods Shift key 0x1000020
Debug[default]: sendKeyEvent QKeyEvent(KeyPress, Key_Control, ShiftModifier|ControlModifier) ((null):0, (null))
[2024-01-11 17:24:37.361670882.361670] [gui.input] Key Press event received: Shift|Control LeftControl
Debug[default]: sendKeyEvent QKeyEvent(KeyPress, Key_Space, ShiftModifier|ControlModifier, text="\u0000") ((null):0, (null))
[2024-01-11 17:24:38.395820813.395820] [gui.input] Character Press event received: Shift|Control
[2024-01-11 17:24:38.395836923.395836] [gui.session] executeAction: ViNormalMode
```

T-727:
Ctrl-Shift-Space:
```
[...]

PR #1434, opened 2024-01-28
Fix Windows alt/control key modifiers

Yaraslaut:
Closes https://github.com/contour-terminal/contour/issues/1408

Also, fixes my locall build with libssh support, see target names: https://github.com/libssh2/libssh2/pull/1104

### DIFF
--- a/src/crispy/flags.h
+++ b/src/crispy/flags.h
@@ -52,9 +52,11 @@ class flags
         _value &= ~static_cast<value_type>(flags._value);
     }
 
+    // Tests for existence of all given flags to be present.
+    // @return true if all flags are set in this flags set, false otherwise.
     [[nodiscard]] constexpr bool contains(flags<flag_type> flags) const noexcept
     {
-        return _value & flags.value();
+        return (_value & flags.value()) == flags.value();
     }
 
     [[nodiscard]] constexpr bool test(flag_type flag) const noexcept { return contains(flag); }
--- a/metainfo.xml
+++ b/metainfo.xml
@@ -107,6 +107,7 @@
     <release version="0.4.2" urgency="medium" type="development">
       <description>
         <ul>
+          <li>Fixes Windows alt/control key modifiers (#1408).</li>
           <li>Fixes Windows built accidentally not including builtin-SSH support (#1427).</li>
           <li>Fixes some text run segmentation problems, e.g., not making use of some programming ligatures (#395).</li>
           <li>Fixes mouse scroll wheel events going into the wrong direction if `Alt` modifier was pressed at the same time (#394).</li>


src/vtpty/CMakeLists.txt
```diff
@@ -79,7 +79,7 @@ if(LIBSSH2_FOUND)
     target_include_directories(vtpty PRIVATE ${LIBSSH2_INCLUDE_DIRS})
     target_link_libraries(vtpty PRIVATE ${LIBSSH2_LINK_LIBRARIES})
 elseif(WIN32)
-    target_link_libraries(vtpty PRIVATE $<IF:$<TARGET_EXISTS:Libssh2::libssh2_shared>,Libssh2::libssh2_shared,Libssh2::libssh2_static>)
+    target_link_libraries(vtpty PRIVATE Libssh2::libssh2)
```
christianparpart, line 82: This is weird, because i've got it to run. `master` branch should be already working. I don't mind this change regardless, because we should not link against the shared version but only against the static one. :)

src/crispy/flags.h
```diff
@@ -52,6 +52,17 @@ class flags
...
+    [[nodiscard]] constexpr bool containsAll(T ...args)
+    {
+        return ((_value & args) && ...);
+    }
+
+    [[nodiscard]] constexpr bool contains(flag_type flag) const noexcept
+    {
+      return _value & flags<flag_type>{flag}.value();
+    }
+
```
christianparpart, line 65: Instead of adding those two functions, we should fix the one below, that is:

```cpp
    /// Tests for existence of all given flags to be present.
    /// @return true if all flags are set in this flags set, false otherwise.
    [[nodiscard]] constexpr bool contains(flags<flag_type> flags) const noexcept
    {
        return (_value & flags.value()) == _flags.value();
    }
```

src/contour/helper.h
```diff
@@ -133,7 +133,7 @@ constexpr inline vtbackend::Modifiers makeModifiers(Qt::KeyboardModifiers qtModi
     // that are missing the AltGr key.
     // Microsoft does not recommend using Ctrl+Alt modifier for shortcuts.
     auto constexpr AltGrEquivalent = Modifiers { Modifier::Alt, Modifier::Control };
-    if (modifiers.contains(AltGrEquivalent))
+    if (modifiers.containsAll(Modifier::Alt, Modifier::Control))
```
christianparpart, line 136: With the comment below, I think we should be able to revert this line again. :-)

src/crispy/flags.h
Yaraslaut: Updated to this implementation

christianparpart (approved):
Thank you, @Yaraslaut

christianparpart:
```
CMake Error at src/vtpty/CMakeLists.txt:82 (target_link_libraries):
Error:   Target "vtpty" links to:

    Libssh2::libssh2

  but the target was not found.
```

I think, on windows, we need the old way, but then explicitly link against the `_static` version.

Yaraslaut:
> ```
> CMake Error at src/vtpty/CMakeLists.txt:82 (target_link_libraries):
> Error:   Target "vtpty" links to:
> 
>     Libssh2::libssh2
> 
>   but the target was not found.
> ```
> 
> I think, on windows, we need the old way, but then explicitly link against the `_static` version.

I reverted this commit

christianparpart (approved):

PR #1435, opened 2024-01-28
Allow password retry on priv-key auth and try with empty password first automatically

### LINKED ISSUE #1425: SSH should not ask for password if the key is not protected with one.
### Contour Terminal version

Contour Terminal Emulator 0.4.0.6088

### Installer source

### Operating System

Windows 11

### Architecture

x86-64

### Other Software

_No response_

### Steps to reproduce

Set up an SSH profile with an ssh key that is not password protected.

### Expected Behavior

_No response_

### Actual Behavior

If you use an ssh profile contour will ask for a password even if there is no password. Pressing enter on the password prompt works, but it should not be necessary.

### Additional notes

_No response_

christianparpart:
closes #1425.


Yaraslaut (changes requested):
In a such way users will have only two oportunities to type correct password, while 3  is desired number

Yaraslaut (approved):
Oh, i am tired and do not see properly :(
LGTM

Issue #1443, opened 2024-01-31
Compilation failure with LLVM 17: explicitly defaulted three-way comparison operator is implicitly deleted


herrhotzenplotz:
I cannot build contour at commit `b610c52`:

```console
[ 49%] Building CXX object src/vtbackend/CMakeFiles/vtbackend.dir/RenderBufferBuilder.cpp.o
In file included from /usr/home/nico/src/contour/src/vtbackend/ColorPalette.cpp:3:
In file included from /usr/home/nico/src/contour/src/vtbackend/ColorPalette.h:5:
In file included from /usr/home/nico/src/contour/src/vtbackend/Image.h:5:
/usr/home/nico/src/contour/src/vtbackend/primitives.h:132:20: warning: explicitly defaulted three-way comparison operator is implicitly deleted [-Wdefaulted-function-deleted]
  132 |     constexpr auto operator<=>(CellLocation const&) const noexcept = default;
      |                    ^
/usr/home/nico/src/contour/src/vtbackend/primitives.h:111:16: note: defaulted 'operator<=>' is implicitly deleted because there is no viable three-way comparison function for member 'line'
  111 |     LineOffset line {};
      |                ^
/usr/home/nico/src/contour/src/vtbackend/primitives.h:132:70: note: replace 'default' with 'delete'
  132 |     constexpr auto operator<=>(CellLocation const&) const noexcept = default;
      |                                                                      ^~~~~~~
      |                                                                      delete
In file included from /usr/home/nico/src/contour/src/vtbackend/ColorPalette.cpp:3:
In file included from /usr/home/nico/src/contour/src/vtbackend/ColorPalette.h:5:
/usr/home/nico/src/contour/src/vtbackend/Image.h:221:30: error: object of type 'CellLocation' cannot be compared because its 'operator<=>' is implicitly deleted
  221 |                && a.offset() < b.offset());
      |                              ^
/usr/home/nico/src/contour/src/vtbackend/primitives.h:132:20: note: explicitly defaulted function was implicitly deleted here
  132 |     constexpr auto operator<=>(CellLocation const&) const noexcept = default;
      |                    ^
/usr/home/nico/src/contour/src/vtbackend/primitives.h:111:16: note: defaulted 'operator<=>' is implicitly deleted because there is no viable three-way comparison function for member 'line'
  111 |     LineOffset line {};
      |                ^
1 warning and 1 error generated.
[...]

PR #1446, opened 2024-01-31
Bump boxed-cpp version and ensure it's always the correct one being built against

christianparpart:
Refs #1443.

### DIFF
--- a/cmake/ContourThirdParties.cmake
+++ b/cmake/ContourThirdParties.cmake
@@ -2,6 +2,27 @@
 # and is used to inject all the dependencies the operating system's
 # package manager did not provide (not found or too old version).
 
+# {{{ helper: subproject_version(<subproject-name> <result-variable>)
+#
+# Extract version of a sub-project, which was previously included with add_subdirectory().
+function(subproject_version subproject_name VERSION_VAR)
+    # Read CMakeLists.txt for subproject and extract project() call(s) from it.
+    file(STRINGS "${${subproject_name}_SOURCE_DIR}/CMakeLists.txt" project_calls REGEX "[ \t]*project\\(")
+    # For every project() call try to extract its VERSION option
+    foreach(project_call ${project_calls})
+        string(REGEX MATCH "VERSION[ ]+([^ )]+)" version_param "${project_call}")
+        if(version_param)
+            set(version_value "${CMAKE_MATCH_1}")
+        endif()
+    endforeach()
+    if(version_value)
+        set(${VERSION_VAR} "${version_value}" PARENT_SCOPE)
+    else()
+        message("WARNING: Cannot extract version for subproject '${subproject_name}'")
+    endif()
+endfunction(subproject_version)
+# }}}
+
 set(ContourThirdParties_SRCDIR ${PROJECT_SOURCE_DIR}/_deps/sources)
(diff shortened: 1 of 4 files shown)


cmake/ContourThirdParties.cmake
```diff
@@ -124,11 +145,20 @@ else()
     set(THIRDPARTY_BUILTIN_termbench "(bench-headless disabled)")
 endif()
 
+set(BOXED_CPP_VERSION "1.2.2")
```
Yaraslaut, line 148:
```suggestion
set(BOXED_CPP_MINIMAL_VERSION "1.2.2")
```
maybe we can rename it so we see what we mean here

cmake/ContourThirdParties.cmake
```diff
@@ -124,11 +145,20 @@ else()
...
+    message(STATUS "Found boxed-cpp: ${boxed-cpp_VERSION} (system package)")
 else()
     ContourThirdParties_Embed_boxed_cpp()
+    subproject_version(boxed-cpp boxed_cpp_version)
+
+    # Assuming the variable is named boxed-cpp_VERSION, adjust if necessary
+    if(NOT DEFINED boxed_cpp_version OR boxed_cpp_version VERSION_LESS BOXED_CPP_VERSION)
+        message(FATAL_ERROR "Embedded boxed-cpp version must be at least ${BOXED_CPP_VERSION}, but found ${boxed_cpp_version}")
+    endif()
+
```
Yaraslaut, line 161: We need to check version if this is a system package as well
```suggestion
    subproject_version(boxed-cpp boxed-cpp_VERSION)
endif()
    # Assuming the variable is named boxed-cpp_VERSION, adjust if necessary
if(NOT DEFINED boxed-cpp_VERSION OR boxed-cpp_VERSION VERSION_LESS BOXED_CPP_MINIMAL_VERSION)
    message(FATAL_ERROR "${THIRDPARTY_BUILTIN_boxed_cpp} boxed-cpp version must be at least ${BOXED_CPP_MINIMAL_VERSION}, but found ${boxed-cpp_VERSION}")


```

Yaraslaut (changes requested):
We need to check version of package for system package as well

cmake/ContourThirdParties.cmake
christianparpart: Ok right, but we can't use `subproject_version()` for checking the version of a system installed dependency, because `subproject_version()` is checking the sub project's CMakeLists.txt the manual way to extract the version information. We don't have that for a system installed version.

If the system does not provide boxed-cpp **or** it's not a sufficient version, then we get into the else() branch and expect it to be an embedded dependency. 🤔

Yaraslaut: Indeed, we need to try to find version that we need and if not then check embed, so something like 
```
find_package(boxed-cpp ${BOXED_CPP_MINIMAL_VERSION} QUIET)
if(boxed-cpp_FOUND)
....
endif()
// try embeded and check version
```

Yaraslaut (approved):
Great :)
did not know about 'if(COMMAND ...)' existence :)

christianparpart:
> Great :)
> did not know about 'if(COMMAND ...)' existence :)

ChatGPT was a great assistance here ;)

PR #1450, opened 2024-02-02
Fix thread safety issue on mouse move event handling

### LINKED ISSUE #1444: Crash: Cursor (63, 0) does not match clamp to screen (63, 56).
# Short overview

I am currently seeing repeated crashes of contour:

```console
===============================================================================================================================================================================================================================================
Cursor (63, 0) does not match clamp to screen (63, 56).
===============================================================================================================================================================================================================================================
Rendered screen at the time of failure
main page size       : 239x64
history line count   : 1000 (max 1000)
cursor position      : (63, 0)
vertical margins     : 0..63
horizontal margins   : 0..238
main page lines: scrollback cur 1000 max 1000, main page lines 64, used lines 1064, zero index 412
===============================================================================================================================================================================================================================================
<snip>
Image pool:
global image stats: 0 instances, 0 raster, 0 fragments
=============================================================
Hashtale: ImagePool name-to-image mappings
-------------------------------------------------------------
hashslot collisions : 0 (0%)
stats               : 0 hits, 0 misses, 0 evictions, 0% hit rate
hash table capacity : 1024 (0% utilization)
entry count         : 0
entry capacity      : 100 (0% utilization)
-------------------------------------------------------------
===============================================================================================================================================================================================================================================
Abort trap (core dumped)
$
```

Again, this is FreeBSD 14.0-RELEASE-p4 amd64 with FreeBSD clang 16 and Qt 6. The entire thing is running in a pure Wayland environment.

# Reproduce

The issue seems to be quickly reproducible by producing output to the terminal and moving the mouse at the same time.

I did this by running: `while :; do ls -lapho; done` and then moving the mouse. Contour then crashes within milliseconds.

# Logs / Debugger trace

See log at https://gist.githubusercontent.com/herrhotzenplotz/4695175147c1b77c4b6f09aee38e98c5/raw/0a661996166c9dffb7322eee0478a25c5af27be2/log
If you wish that I re-run the same thing again with fancy debug options, please let me know. :-)

christianparpart:
Fixes #1444.

### DIFF
--- a/metainfo.xml
+++ b/metainfo.xml
@@ -114,6 +114,7 @@
           <li>Fixes `scripts/install-deps.sh` for openSuSE (Tumbleweed) to install the correct dependencies.</li>
           <li>Fixes missing dependencies for release .deb packages (#1397).</li>
           <li>Fixes legacy keyboard input protocol not reporting non-Alt modifier (#1411).</li>
+          <li>Fixes thread safety issue on mouse move event handling (#1444).</li>
           <li>Digitally sign MacOS release binaries using a valid Apple ID.</li>
           <li>Add support for authenticating to SSH servers with a private key without a password and without requesting for one - also allow password retry (#1425).</li>
           <li>Default history limit is now 1000 lines rather than 0 lines.</li>
--- a/src/contour/TerminalSession.cpp
+++ b/src/contour/TerminalSession.cpp
@@ -14,6 +14,7 @@
 
 #include <crispy/StackTrace.h>
 #include <crispy/assert.h>
+#include <crispy/utils.h>
 
 #include <QtCore/QDebug>
 #include <QtCore/QFileInfo>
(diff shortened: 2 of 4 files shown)


Yaraslaut (commented):
Unfortunately i can not reproduce behaviour from https://github.com/contour-terminal/contour/issues/1444 so we need to ask @herrhotzenplotz to try this patch
Also, what about mouse press events? they are not protected with lock at the moment, should we guard them as well?

christianparpart:
Good point. I took the time to introduce a `locked(Lockable, Lambda)` helper to avoid not nice looking scoping just to keep the lock at a minimum.

herrhotzenplotz:
Seems to fix #1444 :+1: I'll try to switch my default terminal emulator to contour and see if the smoke comes out :laughing:

PR #1451, opened 2024-02-03
Fix packaging for osx inside actions

Yaraslaut:
MacOS packaging is failing lateley in actions, this PR is atempt to fix it
See https://github.com/actions/runner-images/issues/7522


christianparpart (approved):
Oh wow. Yeah, I also noticed that. I typically trigger a rebuild until it eventually fixes itself. This looks like a workaround though (which I don't understand deeply enough). Approving and hoping for the runner to get fixed instead. :)

PR #1453, opened 2024-02-04
Fix dead lock of terminal

Yaraslaut:
Fixes bug introduced in https://github.com/contour-terminal/contour/pull/1450

### DIFF
--- a/src/vtbackend/Terminal.cpp
+++ b/src/vtbackend/Terminal.cpp
@@ -1253,8 +1253,6 @@ namespace
 
 string Terminal::extractSelectionText() const
 {
-    auto const _ = std::scoped_lock { *this };
-
     if (!_selection || _selection->state() == Selection::State::Waiting)
         return "";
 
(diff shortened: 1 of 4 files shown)


src/vtbackend/Terminal.h
```diff
@@ -391,17 +391,9 @@ class Terminal
     void updateInputMethodPreeditString(std::string preeditString);
     // }}}
```
christianparpart, line 393: Looks like we want to get rid of the inner lock then

Yaraslaut: Deleted _innerLock

src/vtbackend/Terminal.cpp
```diff
@@ -1274,8 +1272,6 @@ string Terminal::extractSelectionText() const
 
 string Terminal::extractLastMarkRange() const
 {
-    auto const _ = std::lock_guard { *this };
```
christianparpart, line 1277: if we remove this, we need to apply the lock from outside, i.e. in

```cpp
bool TerminalSession::operator()(actions::CopyPreviousMarkRange)
```

src/vtbackend/Terminal.cpp
```diff
@@ -1253,8 +1253,6 @@ namespace
 
 string Terminal::extractSelectionText() const
 {
-    auto const _ = std::scoped_lock { *this };
```
christianparpart, line 1256: Wen removed here, we need to add one from outside here:

```cpp
bool TerminalSession::operator()(actions::CopySelection copySelection);
bool TerminalSession::operator()(actions::OpenSelection);
```

All the other cases I found were indirectly governed by the caller (/of a caller) already.

christianparpart (approved):
LGTM. Many thanks, @Yaraslaut

PR #1457, opened 2024-02-05
Add CMakePresets.json

christianparpart:
Move more libs from embedded to vcpkg.json (on Windows) and add a CMakePresets.json.

CMake presets are a common way to configure your project. Most IDEs have picked this up so far, so no special IDE-specifics will be needed (anymore), but it can also be done on the CLI of course.

## Configure the project

```sh
cmake --preset windows-x64
```

## Build the project

```sh
cmake --build --preset windows-x64-Debug 
```

## Checklist

- [x] windows build support (MSVC)
- [x] windows build support (CLANG)
- [x] OS/X build support
- [x] Linux build
- [x] add run/debug target (per platform?)
- [x] add ctest target (per platform?)
- [x] add cpack target (per platform?)
- [x] Adapt CI to make use of CMake Presets


Yaraslaut:
Some bits of info between difference between cmake --target package and cpack 
https://discourse.cmake.org/t/are-the-commands-cmake-build-target-package-and-cpack-equivalent/612

Issue #1049, opened 2023-02-23
Insert of second image corrupts sixel graphics


Yaraslaut:
### Contour Terminal version

Contour Terminal Emulator 0.3.11-master-cd0aa937

### Installer source

Github: source code cloned

### Operating System

RHEL

### Architecture

x86-64

### Other Software

julia

### Steps to reproduce

https://user-images.githubusercontent.com/44506630/221343690-788f3798-f458-4d73-8e12-e5bebfc2bc9a.mp4

### Actual Behavior

sixel rendering is broken

### Additional notes

To reproduce this behavior size of the terminal in pixels must exceed 1080p

[figure](https://user-images.githubusercontent.com/44506630/221111967-293fd7e5-a82f-485d-bbfa-a7e6407b92fd.png)
[lines](https://user-images.githubusercontent.com/44506630/221111970-1fddf1a7-48b8-40d4-b351-7f9a20e29623.png)

christianparpart:
I am looking at this from my phone, sorry for this. Is the problem that at the end of the scroll, the background image is shining through?

christianparpart:
I wonder if this is only happening on qml branch or also on master. 🤔

Yaraslaut:
> I wonder if this is only happening on qml branch or also on master. thinking

No, I am seeing same behaviour on current master as well `Contour Terminal Emulator 0.3.11-master-99d6e65a`

christianparpart:
Ok thx. This was just a hunch.

Yaraslaut:
> I am looking at this from my phone, sorry for this. Is the problem that at the end of the scroll, the background image is shining through?

The problem is that after i insert second image first becomes corrupted and i see this:
![image](https://user-images.githubusercontent.com/44506630/221489254-7aeef5af-a22a-4dc1-b45c-6b639b37f656.png)

Yaraslaut:
Can be reproduced with one image, example julia code 
`f = Figure(resolution=(2000,1100)); ax = Axis(f[1,1]); lines!(ax,rand(100)); f`

christianparpart:
We figured out that #1458 is only a interim workaround but not an ultimate fix.

What should be done to close this ticket is to probably have a atlas texture size with the dimension of the current screen this window is living in.

- if the window is moved to another screen (or the physical screen size is changed), then the texture atlas needs to be re-allocated
- if the font size is changed only, then the texture atlas itself only needs to be re-addressed and re-initialized but not re-created.

PR #1458, opened 2024-02-05
Some small corrections and fix automatic adjustement of atlas size

Yaraslaut:
Refs https://github.com/contour-terminal/contour/issues/1049

Issue was with atlas size, magic factor of 3 is enough for 2k resolution to not get this bug 

I am not sure about 4k resolution though since i can test only on 2k, so @christianparpart  maybe you can give it a try?

### DIFF
--- a/metainfo.xml
+++ b/metainfo.xml
@@ -107,7 +107,7 @@
     <release version="0.4.4" urgency="medium" type="development">
       <description>
         <ul>
-          <li> ... </li>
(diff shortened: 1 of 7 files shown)


christianparpart:
Uff, many thanks. Let me try (to understand) that tomorrow first. We can talk about it in our call then, as well. :)

christianparpart (dismissed):

PR #1464, opened 2024-02-12
Modularize CMakePresets.json using includes

christianparpart:


Yaraslaut (changes requested):
Maybe we can also propagate condition for OS to the very top preset file?

christianparpart:
> Maybe we can also propagate condition for OS to the very top preset file?

i actually considered that as well. for some reason I believed it not being possible. 🤔

Issue #1466, opened 2024-02-12
No window shown


cqexbesd:
### Contour Terminal version

3faac53

### Installer source

Github: source code cloned

### Operating System

Arch Linux as of this morning

### Architecture

x86-64

### Steps to reproduce

`contour`

### Expected Behavior

A window to appear

### Actual Behavior

No window appears although the process doesn't exit. Logged to stdout is:

```
Warning: QQmlApplicationEngine failed to load component ((null):0, (null))
Warning: qrc:/contour/ui/main.qml: No such file or directory (qrc:/contour/ui/main.qml:4294967295, (null))
```

For some reason the version string is also strange...`Contour Terminal Emulator 0.4.4-v0.4.3.6442-12-g3faac536-3faac536` I rm'ed the build directory before compiling however.

With some debug logging on:

```
[2024-02-12 16:01:21.624716214.624716] [config] Loading configuration from file: /home/andrew/.config/contour/contour.yml
[2024-02-12 16:01:21.627497301.627497] [config] Missing key background_image.path. Using default: "".
[2024-02-12 16:01:21.627527267.627527] [config] Missing key .shell. Using default: "".
[2024-02-12 16:01:21.627562242.627562] [config] Missing key .refresh_rate. Using default: 0.
[2024-02-12 16:01:21.627594953.627594] [config] Defaulting TERM to contour.
[2024-02-12 16:01:21.627605052.627605] [config] Missing key .margins.horizontal. Using default: 0.
[2024-02-12 16:01:21.627611153.627611] [config] Missing key .margins.vertical. Using default: 0.
[2024-02-12 16:01:21.627657960.627657] [config] Missing key .option_as_alt. Using default: false.
[2024-02-12 16:01:21.627709116.627709] [config] Missing key status_line.position. Using default: none.
[2024-02-12 16:01:21.627716419.627716] [config] Missing key status_line.sync_to_window_title. Using default: false.
[2024-02-12 16:01:21.628003875.628003] [config] Superfluous config key found: profiles.main.font.dpi_scale
[2024-02-12 16:01:21.639262211.639262] [gui.display] Color theme mode at startup: Light
Warning: QQmlApplicationEngine failed to load component ((null):0, (null))
Warning: qrc:/contour/ui/main.qml: No such file or directory (qrc:/contour/ui/main.qml:4294967295, (null))
```

strace output attached.
[Uploading strace.txt…]()

### Additional notes

[...]

PR #1467, opened 2024-02-12
Update presets to handle build type and directories for release correctly 

Yaraslaut:
Closes https://github.com/contour-terminal/contour/issues/1466

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,11 +103,11 @@ can be insalled via the system package manager.
 #### Compile
 
 ```sh
-cmake -S . -B build -G Ninja
-cmake --build build/
+cmake --preset linux-release
+cmake --build --preset linux-release
 
 # Optionally, if you want to install from source
-cmake --build build/ --target install
+cmake --build --preset linux-release --target install
 ```
 
 #### Windows 10 or newer
--- a/cmake/presets/common.json
+++ b/cmake/presets/common.json
@@ -7,7 +7,6 @@
             "description": "Common settings for all configurations",
             "binaryDir": "${sourceDir}/build/${presetName}",
             "cacheVariables": {
-                "CMAKE_INSTALL_PREFIX": "${sourceDir}/out/install/${presetName}",
                 "CONTOUR_INSTALL_TOOLS": "ON",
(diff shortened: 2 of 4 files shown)


cmake/presets/common.json
```diff
@@ -20,14 +19,16 @@
             "name": "release",
             "hidden": true,
             "cacheVariables": {
-                "CMAKE_BUILD_TYPE": "RelWithDebInfo"
+                "CMAKE_BUILD_TYPE": "RelWithDebInfo",
+                "CMAKE_INSTALL_PREFIX": "/usr/local"
```
christianparpart, line 23: this should actually be only in `os-linux.json` (which reminds me of, we didn't think about FreeBSD, damn, there it should also be like that)

docs/install.md
```diff
@@ -66,11 +66,11 @@ can be insalled via the system package manager.
...
 ```sh
-cmake -S . -B build -G Ninja
-cmake --build build/
+cmake --preset linux-release
+cmake --build --preset linux-release
 
 # Optionally, if you want to install from source
-cmake --build build/ --target install
+cmake --build --preset linux-release --target install
 ```
```
christianparpart, line 74: this assumes Linux OS.

Maybe we could come up with a `default` target that is `linux-release` on linux, `macos-release` on mac, etc. I don't have a perfect answer to this one yet either (doesn't have to be addressed in this PR). 🤔

README.md
```diff
@@ -103,11 +103,11 @@ can be insalled via the system package manager.
...
 ```sh
-cmake -S . -B build -G Ninja
-cmake --build build/
+cmake --preset linux-release
+cmake --build --preset linux-release
 
 # Optionally, if you want to install from source
-cmake --build build/ --target install
+cmake --build --preset linux-release --target install
 ```
```
christianparpart, line 111: this also only addresses Linux OS (see below)

Yaraslaut: I wil add info mac os, and for freebsd i suppose that one can use linux one as well

christianparpart (dismissed):
Many thanks, @Yaraslaut :)

PR #1469, opened 2024-02-13
Fix qt resources call in cmake

Yaraslaut:
Fix call to add qt resources, see documentation for `qt_add_resources` https://doc.qt.io/qt-6/qt-add-resources.html

Refs https://github.com/contour-terminal/contour/issues/1466 
i was getting same error


cqexbesd:
I can confirm this is what fixes it for me.

Issue #1475, opened 2024-02-17
TERM variable no longer set


cqexbesd:
### Contour Terminal version

0.4.4-master-ee4d25ce

### Installer source

Github: source code cloned

### Operating System

Arch Linux as of this morning

### Architecture

x86-64

### Steps to reproduce

Start a new Contour process. Note that the shell does not inherit a `TERM` env variable (or `COLORTERM`)

### Expected Behavior

TERM would be set to contour as it always has before.

### Actual Behavior

Variable is not set

### Additional notes

I don't do anything specifically in my config file, I have just been relying on default behaviour up until . Confirmed with strace that the env variable isn't passed to exec. IIRC 5f9a1edfbb1902b97c463bd561635849a0e6b86f worked OK.

Haven't had a chance to delve any deeper.

christianparpart:
Hey @cqexbesd. Sorry for this bad experience. This is most likely due to the very recent merge of #1300. Currently @Yaraslaut knows the most about it. I'm trying to catch up on this topic :)

cqexbesd:
No need to apologise. We have all added our fair share of bugs over the years - it happens. Secondly one of the reasons I build from git is to try and pick up some bugs. I don't have time to learn the code but as I'm running contour anyway I may as well try to test it at the same time, to help in some small way. At least I am (hopefully) slightly more equipped to understand what might be going wrong, and work around the worst problems, than the average user who gets a build from their distro.

Contour is by far the best terminal I have used (sorry Roxterm - you served me well for so many years). It has (almost) all the features I want and the speed that bugs are fixed when I do report them is very impressive. It does you, and @Yaraslaut et al credit.

Yaraslaut:
Hi @cqexbesd  thanks for your feeedback and constant testing of master :)  I created PR to adress this issue https://github.com/contour-terminal/contour/pull/1476 can you please give it a try?

cqexbesd:
I created PR to adress this issue #1476 can you please give it a try?I tried it and it works for me. I’m a bit unsure if it will work for e.g. systems using termcap though. Perhaps it would be better to just try and set TERM and see if it works?My shell settings already have a series of fallbacks that use tcsh’s termname to find a working value. I forget what termname actual calls (though I think it might have been me who added it). I don’t know if the terminal should be doing this logic at all though because the movement you e.g. ssh it is all in vain.I’m currently in a location where GitHub seems to be banned so I don’t know if you will see this…

[...]

PR #1476, opened 2024-02-18
Ammendum to config refactor

Yaraslaut:
Closes https://github.com/contour-terminal/contour/issues/1475

### DIFF
--- a/src/contour/Config.cpp
+++ b/src/contour/Config.cpp
@@ -432,11 +432,13 @@ void YAMLConfigReader::loadFromEntry(YAML::Node const& node,
 {
     logger()("color palette loading {}", entry);
     auto child = node[entry];
+
     if (!child) // can not load directly from config file
     {
 
-        logger()("color paletter not found inside config file, checking colorschemes directory {} \n ",
-                 entry);
+        logger()(
+            "color paletter not found inside config file, checking colorschemes directory for {}.yml file",
+            entry);
         auto const filePath = configFile.remove_filename() / "colorschemes" / (entry + ".yml");
         auto fileContents = readFile(filePath);
         if (!fileContents)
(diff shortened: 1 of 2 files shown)


src/contour/Config.cpp
```diff
@@ -102,6 +102,56 @@ namespace
...
+        if (appTerminfoDir.has_value())
+            locations.emplace_back(appTerminfoDir.value().string());
+
+        locations.emplace_back(Process::homeDirectory() / ".terminfo");
+
+        if (auto const* value = getenv("TERMINFO_DIRS"); value && *value)
+            for (auto const dir: crispy::split(string_view(value), ':'))
+                locations.emplace_back(string(dir));
+
+        locations.emplace_back("/usr/share/terminfo");
```
christianparpart, line 118: not sure this should be hard coded. at least for UNIX installs, we could make use of cmake provided POSIX install-prefix (using GNUInstallDirs or so, in cmake, a variable that should than be preprocessor-defined in the CMakeLists.txt

christianparpart (commented):
@Yaraslaut I remember we had all this already. Did it accidentally go off the board in that refactor PR? :)

Yaraslaut:
Yes, it was lost somewhere during refactor, I will check that I did not introduce another ones :)

src/contour/Config.cpp
Yaraslaut: Fom man page of `tic` it seems that this is default install path for termifo files, and dont think that it will be changed by any distro

christianparpart: we should at least ifdef out WIN32 then 🤔

Yaraslaut: See  getdefaultTERM, we just return "contour" without checking any files

christianparpart (approved):
Thank you, @Yaraslaut :)

PR #1477, opened 2024-02-20
Fix format of mods during config generation

Yaraslaut:
Fix generation of new config file


christianparpart:
image

Also colors need fixing, so it seems ;)

christianparpart (approved):
thank you :)

Issue #1460, opened 2024-02-08
How to remove "Shell terminated too quickly. The window will not be closed automatically."


TheTerrior:
I'm wondering about behavior I'm having with Contour which doesn't happen in other terminals. When I open a new terminal and quickly type "exit", I get the message from the title. I will provide a screenshot here as well. How do I prevent this message from popping up? The terminal doesn't close until another key is pressed or I kill the window. Expected behavior is to exit as soon as the command is inputted, I don't get this message with alacritty, kitty, or wezterm. I'm sure this isn't a bug, but I don't see any reference to this online.

![image](https://github.com/contour-terminal/contour/assets/36648271/592ca30c-abcb-4277-a8ab-1c36edba9040)

System info:
NixOS 23.11

Yaraslaut:
You can see some info from `contour help` page : 
```
            [early-exit-threshold UINT]    If the spawned process exits earlier than the given threshold seconds, an error message will be printed and the window not closed immediately.
                                           [default: 6]

```
This message you see happens only if you try to exit immediately after opening the terminal, so running with `contour early-exit-threshold 0` must solve the problem, also, what version of contour are you using?

TheTerrior:
Seems to be 0.3.12.262. Adding the early-exit-threshold argument fixes it, thank you. I'm not sure why this threshold exists though, what is the logic behind it being set at 6 seconds?

Yaraslaut:
> Seems to be 0.3.12.262. Adding the early-exit-threshold argument fixes it, thank you. I'm not sure why this threshold exists though, what is the logic behind it being set at 6 seconds?

@christianparpart  might tell some story about it, i think that it was introduced on the early stage of development for test purposes.

christianparpart:
Hi @TheTerrior. This was added because it MAY be that whatever application you spawn as main program in your terminal (usually the shell), can exit right at the startup. This leads to the user believing that the terminal is not working, because you don't see that it even was there, and even if you noticed that, the window would be closed too quickly to actually read any potential diagnostic error message, telling you, why the program terminated too quickly.

Contour since then added a guard that prevents too quick terminating apps to also terminate the terminal in order to allow the user to review the screen before finally closing it.

However, on our CI side, we must expect some tests to terminate quickly, so `early-exit-threshold SECONDS` was added to to be able to set it to `0`, especially since in test-automation, there'll be no user to trigger input of any form, just to finish the integration-test.

[...]

PR #1479, opened 2024-02-27
Implement early exit threshold config entry

Yaraslaut:
Closes https://github.com/contour-terminal/contour/issues/1460

### DIFF
--- a/src/contour/ContourGuiApp.cpp
+++ b/src/contour/ContourGuiApp.cpp
@@ -109,7 +109,7 @@ crispy::cli::command ContourGuiApp::parameterDefinition() const
                     "Dumps internal state at exit into the given directory. This is for debugging contour.",
                     "PATH" },
                 CLI::option { "early-exit-threshold",
-                              CLI::value { 6u },
+                              CLI::value { -1 },
                               "If the spawned process exits earlier than the given threshold seconds, an "
                               "error message will be printed and the window not closed immediately." },
                 CLI::option { "working-directory",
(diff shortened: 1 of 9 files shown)


src/contour/ContourGuiApp.cpp
```diff
@@ -149,7 +149,17 @@ crispy::cli::command ContourGuiApp::parameterDefinition() const
 
 std::chrono::seconds ContourGuiApp::earlyExitThreshold() const
 {
-    return std::chrono::seconds(parameters().get<unsigned>("contour.terminal.early-exit-threshold"));
+    auto const configThreshold = config().earlyExitThreshold.value();
+    auto const parameterThreshold = parameters().get<unsigned>("contour.terminal.early-exit-threshold");
+
+    // default threshold is config::documentation::DefaultEarlyExitThreshold seconds
+    if (parameterThreshold != config::documentation::DefaultEarlyExitThreshold)
+        return std::chrono::seconds(parameterThreshold);
```
christianparpart, line 157: what if the user actually explicitly passed an early-exit-threshold that matches the default? Then this branch would not be taken.

src/contour/contour.yml
```diff
@@ -64,6 +64,9 @@ pty_buffer_size: 1048576
 
 default_profile: main
 
+# Time in seconds to check for early threshold
+early_exit_threshold: 5
+
```
christianparpart, line 69: Didn't we make this contour.yml file obsolete?

Yaraslaut: We do not use it to generate config file, this is true, i just placed it there to keep it updated until we have a better way to see full default config file

src/contour/ContourGuiApp.cpp
Yaraslaut: You are right, updated implementation

christianparpart (approved):

Issue #1423, opened 2024-01-25
Other apps try to open folders in Contour rather than the file manager


jobukkit:
### Contour Terminal version

0.4.2-master-dbb8f274

### Installer source

AUR (tried both stable and git)

### Operating System

CachyOS (Arch Linux) with GNOME

### Architecture

x86-64

### Steps to reproduce

When Contour is installed, other apps try to open folders in Contour, rather than Nautilus. Removing Contour fixes the issue.

### Expected Behavior

Having Contour installed shouldn't interfere with the file manager.

### Actual Behavior

Contour opens and interprets the folder path as a shell to use. For example, when trying to browse the files of a Steam game:
```
Failed to spawn /home/myuser/.local/share/Steam/steamapps/common/Unrailed
Trying login shell: "/usr/bin/nu"
```

christianparpart:
Hey @jobukkit, many thanks for this report. This is really weird. I'm mainly using KDE. Would you mind trying out another desktop or window manager, just for the test to see if this is a GNOME-specific behaviour or not? Many thanks.

Yaraslaut:
I have a feeling that this cause by some keywords from our .desktop file 
https://github.com/contour-terminal/contour/blob/f3c3334aa5c861348c5bbe8ffe572c872eef2e08/src/contour/contour.desktop#L12

and i think that @jobukkit you can directly test if this will resolve this issue by modifiing contour.desktop file on your system

jobukkit:
Seems like the line above that is the culprit.

https://github.com/contour-terminal/contour/blob/f3c3334aa5c861348c5bbe8ffe572c872eef2e08/src/contour/contour.desktop#L11

```
> xdg-mime query default inode/directory
org.contourterminal.Contour.desktop
```

`xdg-mime default org.gnome.Nautilus.desktop inode/directory` fixes the problem.

Maybe something other than the .desktop file sets it too.

PR #1480, opened 2024-02-28
Update desktop entry

Yaraslaut:
Closes https://github.com/contour-terminal/contour/issues/1423
Mime type changed to one used in Konsole, also keywords deleted, see https://github.com/KDE/konsole/blob/61264c1917770102a85123d388528f2bcacfd960/desktop/konsolerun.desktop#L4

### DIFF
--- a/metainfo.xml
+++ b/metainfo.xml
@@ -107,13 +107,14 @@
     <release version="0.4.4" urgency="medium" type="development">
       <description>
         <ul>
-          <li>Adds 'early_exit_threshold' config option (#1460)</li>
+          <li>Add 'early_exit_threshold' config option (#1460)</li>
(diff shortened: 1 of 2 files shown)


christianparpart (dismissed):
thank you :)

christianparpart:
I know it is a one-liner PR, but I think it might be worth mentioning in a changelog, especially since we had people noticing this problem.

Yaraslaut:
> I know it is a one-liner PR, but I think it might be worth mentioning in a changelog, especially since we had people noticing this problem.

Updated

christianparpart (approved):

christianparpart (dismissed):
Thank you! :)

PR #1483, opened 2024-03-09
Update of config generation

Yaraslaut:
Fix bug with generation of few entries in the config


christianparpart (dismissed):
interesting. seemed to be more than I expected. thank you :)

christianparpart (approved):
many thanks, @Yaraslaut :)

PR #1491, opened 2024-03-26
Introduce boxed types for margins and handling of input

Yaraslaut:
Add boxed types for window margins and input handling

### DIFF
--- a/src/contour/Config.h
+++ b/src/contour/Config.h
@@ -185,10 +185,20 @@ enum class RenderingBackend
     OpenGL,
 };
 
+struct HorizontalMarginTag
+{
+};
+struct VerticalMarginTag
+{
+};
+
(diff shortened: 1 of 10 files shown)


christianparpart (approved):
For horizontal and vertical margins I totally agree and understand. For `Handled` it took me a while to see the benefit. I think there actually is one, because we encode the meaning of that boolean. So 👍 :)

Yaraslaut:
> For horizontal and vertical margins I totally agree and understand. For `Handled` it took me a while to see the benefit. I think there actually is one, because we encode the meaning of that boolean. So 👍 :)

Concerning `Handled`, yes, i added it to see the meaning of the bool value, since it is not obvious what this bool means when returned from one function into another

PR #1493, opened 2024-04-03
Small refactor to generalize config creation

Yaraslaut:
Small refactor of config creation.

### DIFF
--- a/src/contour/Config.cpp
+++ b/src/contour/Config.cpp
@@ -181,10 +181,15 @@ fs::path configHome()
     return configHome("contour");
 }
 
+std::string createString(Config const& c)
+{
+    return createString(c, YAMLConfigWriter());
+}
+
 std::string defaultConfigString()
 {
     const Config config {};
-    auto configString = YAMLConfigWriter().createString(config);
+    auto configString = createString(config);
 
     auto logger = configLog;
     logger()(configString);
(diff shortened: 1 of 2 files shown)


src/contour/Config.cpp
```diff
@@ -1895,12 +1896,12 @@ std::string YAMLConfigWriter::createString(Config const& c)
...
     doc.append(fmt::format(fmt::runtime(c.profiles.documentation), fmt::arg("comment", "#")));
     {
-        const auto _ = Offset {};
+        const auto _ = typename Writer::Offset {};
         for (auto&& [name, entry]: c.profiles.value())
         {
             doc.append(fmt::format("    {}: \n", name));
             {
-                const auto _ = Offset {};
+                const auto _ = typename Writer::Offset {};
```
christianparpart, line 1904: is Offset needed to be inside Writer? Maybe we can get rid of the not-so-nice `typename` :)

christianparpart: ok, further below I see `Writer::Offset::levels`, I wonder why there it works, but for `Writer::Offset` it doesn't. 🤔

Yaraslaut: we need it here because `Writer` is a template parameter, and name lookup by default looking for class members only, but we need to specify that this is a type

christianparpart (approved):
sorry for the late response. LGTM :)

PR #1497, opened 2024-04-16
Fix crush when status line is enabled

Yaraslaut:
This PR fixed a crush that i notice often when using status line
Closes https://github.com/contour-terminal/contour/issues/1511


src/vtbackend/Terminal.cpp
```diff
@@ -1250,20 +1250,24 @@ struct LocalSequenceHandler // {{{
...
-        instructionCounter += chars.size();
-        targetScreen.writeText(chars, cellCount);
-        return terminal.settings().pageSize.columns.as<size_t>()
-               - terminal.currentScreen().cursor().position.column.as<size_t>();
+        // implementation of targetScreen.writeText(chars, cellCount)
+        // is buggy and does not work correctly
+        // so we do not use optimization for
+        // ParserEvents::print(chars,cellCount)
+        // but write char by char
+        // TODO fix targetScreen.writeText(chars, cellCount)
```
christianparpart, line 1266: I need to think a little bit about that. 🤔

PR #1498, opened 2024-04-16
Drop dead files from project (TerminalState.{cpp,h})

christianparpart:
i wonder how we missed the deletion


Yaraslaut (approved):
LGTM :)

Issue #1047, opened 2023-02-19
Provide statically linked binary through CI


christianparpart:
That should greatly boost Linux distribution independence.

Users should generally always prefer dynamically linked binaries, but Other projects have shown that it is indeed possible to statically link a Qt application.

With that, only Qt platform plugins would need to be embedded into the install archive (tgz, ...) and a custom `qt.conf` would mandate the binary where to find those.

Our CI should create such a binary for regular builds as well as for release and make it then available on the release page as well. 

This could be seen as prep-work for #956, #761.

Yaraslaut:
Now we have static builds for qt and have github action to create them, so i am closing this one as complete

PR #1503, opened 2024-05-04
Add CPM fallback for the dependencies and static build

Yaraslaut:
This PR adds CPM fallback for all packages when they are non embedded using install-deps.sh script or system wide package is not present. 

And add static build of contour

Refs https://github.com/contour-terminal/contour/issues/1047


version.txt
```diff
@@ -0,0 +1 @@
+0.4.4-add_cpm-7f59f0f8
```
christianparpart, line 1: This file should not be version controled. We probably have to add this to .gitignore to not accidentally get it added :)

src/contour/ContourGuiApp.cpp
```diff
@@ -23,6 +25,7 @@
 #include <QtQml/QQmlContext>
 #include <QtWidgets/QApplication>
 
+#include <QtPlugin>
```
christianparpart, line 28:
```suggestion
#include <QtCore/QtPlugin>
```

Always have Qt includes module-prefixed. This should also fix include order.

src/contour/ContourGuiApp.cpp
```diff
@@ -12,6 +12,8 @@
 #include <crispy/utils.h>
 
 #include <QtCore/QProcess>
+
+#include <qqmlextensionplugin.h>
```
christianparpart, line 16:
```suggestion
#include <QtQml/QQmlExtensionPlugin.h>
```

christianparpart (dismissed):

christianparpart:
rebased on top of latest master

christianparpart (approved):
Thank you, @Yaraslaut :)

Issue #1499, opened 2024-04-21
OSC-8 link storage(?) broken for links generated with aerc colorize


ferdinandyb:
### Contour Terminal version

Contour Terminal Emulator 0.4.4-master-521b1408

### Installer source

Github: source code cloned

### Operating System

Ubuntu 22.04

### Architecture

x86-64

### Other Software

[Colorize](https://git.sr.ht/~rjarry/aerc/tree/master/item/filters/colorize.c) filter from aerc filters. Probably shouldn't be hard to build, but if you tell me what logs to make I can paste that as well.

### Steps to reproduce

I was running the below script:

```
#!/bin/bash

for (( i = 0; i < 20; i++ )); do
  echo "http://example$i.com" | ~/.local/libexec/aerc/filters/colorize -8
  # printf '\033]8;;http://example%s.com\033\\This is a link %s\033]8;;\033\\\n' "$i" "$i"
done
```

but actually two different links are enough to reproduce. Contour will consistently open the link corresponding to first link that was output by colorize in the terminal. A subsequent osc-8 with the commented out printf will work fine, but any osc-8 output by colorize will forever open the same (first) link.

### Expected Behavior

The proper links are opened.

### Actual Behavior

Contour will consistently open the link corresponding to first link that was output by colorize in the terminal. A subsequent osc-8 with the commented out printf will work fine, but any osc-8 output by colorize will forever open the same (first) link.

Yaraslaut:
We see this bug because colorize.c sets id for the url, some info about id you can find here: https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda
While initial intention is to use id only for `Hover underlining` we also use it to `open links`. When user opens a link, we search for it  by `id` inside internal url cache, and since `id` is the same for all of them only first link is exist in the cache. 

Possible solution is to ignore user specified `id` since we set it  anyway, and when we count it internally then everything will go smooth  
Maybe @christianparpart  has a better idea

PR #1507, opened 2024-05-23
Fix collision of osc-8 id

Yaraslaut:
Closes https://github.com/contour-terminal/contour/issues/1499


src/vtbackend/Screen.cpp
```diff
@@ -1470,19 +1470,13 @@ void Screen<Cell>::setCurrentWorkingDirectory(string const& url)
 
 template <typename Cell>
 CRISPY_REQUIRES(CellConcept<Cell>)
-void Screen<Cell>::hyperlink(string id, string uri)
+void Screen<Cell>::hyperlink([[maybe_unused]] string id, string uri)
```
christianparpart, line 1473: Thank you, @Yaraslaut, for this PR.

I think, if this parameter is not to be used anymore, we should get rid of it. We could then also give this function a more actionable name, to make it more clear on the caller side, what this function is intending to do.

j4james: If you're not using the id parameter, doesn't that mean you'll fail to highlight both parts of the hyperlink when it gets split (e.g. in something like a wrapped tmux pane)?

Yaraslaut: > If you're not using the id parameter, doesn't that mean you'll fail to highlight both parts of the hyperlink when it gets split (e.g. in something like a wrapped tmux pane)?

Not sure about tmux, but every link has its own unique id and after wrap everything is alright:
image

Yaraslaut: > Thank you, @Yaraslaut, for this PR.
> 
> I think, if this parameter is not to be used anymore, we should get rid of it. We could then also give this function a more actionable name, to make it more clear on the caller side, what this function is intending to do.

I didn't change the signature to have osc-8 compliant function for hyperlinks, if you think that we do not need to keep it that way, i can propagate this change in the code further

j4james: The point is that a multiplexer might have to redraw the wrapped link itself, potentially somewhere in the middle of the screen. In that case you'll end up with two links with the same id, which the multiplexer will output something like this:

```
printf '\e[10C\e]8;id=123;http://example.com/\e\\WRAPPE\e]8;;\e\\\n\e]8;id=123;http://example.com/\e\\D_LINK\e]8;;\e\\\n'
```

Hovering over either half of the link should underline both parts. That's the whole reason for the id parameter.

Yaraslaut: Oh, i see, not sure which of the behaviours we want to break then, i would prefer instead of the id of the link to  have some flag indicating that we need to connect the link with previous one, otherwise i have no idea how this can be overcome, any ideas @j4james, @christianparpart ?

j4james: What I was trying to explain in https://github.com/contour-terminal/contour/issues/1499#issuecomment-2098339025, and what the spec says, is that you should combine the id with the url in order to get a unique identifier for each hyperlink. If you've got two links on the page with the same id and url, they should both be underlined together. If you've got two links with the same id, but with different urls, then they won't be underlined together, and should have separate entries in your cache. The bug in #1499 should not be possible if you follow this approach.

Yaraslaut: I updated PR @j4james, thanks a lot for taking a time and explained  what I was missing in my previous fix. Hope that now everything is correct :)

j4james: @Yaraslaut I don't know the code base, so I can't really review it properly, but the latest changes look good to me.

Issue #1378, opened 2023-12-22
Better default bell sound


erf:
## Abstract

The default bell sound '\u0007' is bad to my ears 

## Motivation

Change default bell sound to default macOS sound

## Specification

Use same sound as on Terminal.app and iTerm2.app

whisperity:
In general, it would be better if the default behaviour of the bell was to trigger the DE-specific default notification sound, with the option to override it in Contour.

Yaraslaut:
What if we use this bell sound ? 
https://commons.wikimedia.org/wiki/File:Bicycle-bell-2.wav

PR #1508, opened 2024-05-23
Update bell sound to use old one in wav format

Yaraslaut:
Closes https://github.com/contour-terminal/contour/issues/1378


src/contour/resources.qrc
```diff
@@ -1,7 +1,7 @@
 <RCC>
     <qresource prefix="/contour">
         <file alias="bell.oga">res/bell.oga</file>
-        <file>bell.wav</file>
+        <file alias="bell.wav">res/bell.wav</file>
```
christianparpart, line 4: why was the source path moved?

christianparpart (approved):

Issue #163, opened 2021-01-24
create shell integration scripts


christianparpart:
For at least bash and zsh

For preexec:
* Add vertical line marker
* Disable text reflow

For preexec:
* Enable text reflow

On init, maybe install functions to:
* display images (like icat)
* change profile to ...
* `vtsend` to send files to host terminal, landing in host's (configurable) Downloads folder

### Checklist

- [x] zsh
- [x] fish
- [x] bash

christianparpart:
Let's do bash at a later time, when we also have users that use bash. :)

whisperity:
Reopen the issue and put them into a milestone bucket called "future features" or something.

christianparpart:
i.e. basically buying some more time without forgetting that one wanted to do that :)

christianparpart:
@data-man I think in on of your screenshots I saw you using fish shell, is that right? Would you mind having a look at how easy it would be to add contour shell integration for fish?

data-man:
> I saw you using fish shell, is that right?

Exactly!

> Would you mind having a look at how easy it would be to add contour shell integration for fish?

Will do!

data-man:
Can we use some template engine for this?
E.g. https://github.com/pantor/inja or https://github.com/kainjow/Mustache (both are header-only).

WSLUser:
Iterm2 has some shell integration scripts but not sure if those are exactly the same as you're looking at.

Yaraslaut:
I suppose fish is done by this PR https://github.com/contour-terminal/contour/pull/1001 ?

christianparpart:
yes, it is. :)

christianparpart:
the question is if bash is relevant at all as of 2023. Most people seem to be using ZSH anyways, and Fish is the new rising star (in my opinion), people are more likely to drive away from bash. We might just close this ticket alltogether and wait for someone actually wanting bash completion. Then it can still be added. No need to create artificial maintenance burdens with no real users behind :)

whisperity:
@christianparpart **Completion** or **integration**? Two different things. Bash integration would still be useful for cases where people just start a small job (e.g. in docker or a remote server) without a full developer environment but would still like to use at least the small features like the "jump to prompt". Really important feature even in maintenance activities, not just development.

(Personally, I'm using ZSH and will continue using ZSH, fish is too arcane for me, so I'd prefer if ZSH were kept in the front of the picture.)

christianparpart:
[...]

PR #1514, opened 2024-06-05
Add bash integration

Yaraslaut:
This PR adds bash integration script to contour and closes https://github.com/contour-terminal/contour/issues/163

### DIFF
--- a/src/contour/shell-integration/shell-integration.bash
+++ b/src/contour/shell-integration/shell-integration.bash
@@ -1,0 +1,389 @@
+# bash-preexec.sh -- Bash support for ZSH-like 'preexec' and 'precmd' functions.
+# https://github.com/rcaloras/bash-preexec
+#
+#
+# 'preexec' functions are executed before each interactive command is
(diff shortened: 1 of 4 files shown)


src/contour/shell-integration/shell-integration.bash
```diff
@@ -0,0 +1,389 @@
...
+fi;
+
+
+
+preexec() {
+    printf "\\e[?2028h";
+}
+precmd() {
+    printf "\\e[?2028l\\e[>M\\e]7;$PWD\\e\\\\";
+}
```
christianparpart, line 389:
```suggestion
# actual customized code (for Contour) starts here:

preexec() {
    printf "\\e[?2028h";
}
precmd() {
    printf "\\e[?2028l\\e[>M\\e]7;$PWD\\e\\\\";
}
```

It seems like all of the above code is taking from the internet, which is totally fine, but we should make it clear where our code starts, such that it's easier to keep it (merely?) up-to-date in case we ever want or need to upgrade the taken code. It's also easier for reviewing our own actual changes in the future. :)

Issue #1023, opened 2023-02-09
Extend word selection


cqexbesd:
## Abstract

Double clicking on a word selects it. One more click and you get the whole line. I would like to have some steps in between, where a different set of word delimiters is used and you can keep growing your selection until you reach the line selection.

(I know this has been discussed but I didn't see it written down hence this issue)

## Motivation

Using the default config, IPv6 addresses have to be dragged over to select with the mouse. URLs are similar. While I could change the delimiters to get them in one go, sometimes I do want to select just a small portion of something. I would prefer just one more click than having to drag or the many key presses needed to enter vi mode and navigate.

## Specification

A choice has to be made if there will be two sets of word delimiters or if it can be any number the user chooses. Only having one more level would probably satisfy 80% of people. Either way the config should allow one or more new sets of delimiters to be specified.

When a user double clicks on some text the first set of delimiters is used to determine what is selected. If the user clicks once more in the same selection then the next set of delimiters are used to find the word boundaries and so on. When there are no more sets configured then the next click selects the whole line as it does now.

* If a user doesn't provide more sets in their config then behaviour would remain as it is now.
* A user could provide a series of smaller and smaller sets so repeated clicks makes the selection larger - but they could also provide larger sets causing the selection to shrink if that's what they desired. I don't see a need to enforce a single direction of growth so it would be possible (if perhaps annoying) to have selections grow and shrink as you click if that's how the user configured it.

christianparpart:
I think supporting one additional selection delimiter configuration apart from word-wise but less than line wise, would make sense, but going beyond that I think, this would probably make things less usable for the user, as in: they might be getting confused by how many clicks they did and end up redoing the click-sequence again, just to get what they want.

So,

- single click: starts linear selection
- double click: selects word wise selection
- if configured: one click more will select an extended word wise selection
- one click more (3 or 4 times therefore) would do line wise selection 

Going for more than one extra layer is certainly possible but I don't see how that could improve workflow. Users should probably be using vim-mode and motion sequences to quickly select a range).

That's just a thought so far.

So now, how would one configure it. Probably with an optional configuration key per profile that would have a name, something like, `alternate_word_selection_delimiters` or so, that accepts a string of all characters that are allowed to be part of such a word. (just like we have already, but for the alternate word selection, which can be triggered using 3-click).

zopieux:
[...]

PR #1518, opened 2024-06-10
Add action to set custom delimiters for selection

Yaraslaut:
Closes https://github.com/contour-terminal/contour/issues/1023

PR adds action to trigger selection with custom delimiters
@cqexbesd you can add following entry in config file:
```
   - { mods: [Control],        mouse: Left,        action: CreateSelection, chars: " " }
```

### DIFF
--- a/src/contour/Config.cpp
+++ b/src/contour/Config.cpp
@@ -1698,6 +1698,16 @@ std::optional<actions::Action> YAMLConfigReader::parseAction(YAML::Node const& n
                 return std::nullopt;
         }
 
+        if (holds_alternative<actions::CreateSelection>(action))
+        {
+            if (auto delimiters = node["delimiters"]; delimiters.IsScalar())
+            {
+                return actions::CreateSelection { delimiters.as<std::string>() };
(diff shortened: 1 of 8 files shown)


src/contour/Config.cpp
```diff
@@ -1698,6 +1698,16 @@ std::optional<actions::Action> YAMLConfigReader::parseAction(YAML::Node const& n
                 return std::nullopt;
         }
 
+        if (holds_alternative<actions::CreateSelection>(action))
+        {
+            if (auto chars = node["chars"]; chars.IsScalar())
```
christianparpart, line 1703:
```suggestion
            if (auto chars = node["delimiters"]; chars.IsScalar())
```

I think here (and all the other related places), the JSON config key for custom delimiters should be named `delimiters` rather than `chars`, to make it more intuitive what the contents of that configuration key actually is.

christianparpart (changes requested):
I'd propose to rename `chars` to `delimiters` to make it more intuitive to use and read

PR #1519, opened 2024-06-11
Add config entry to exit search mode into insert or normal

Yaraslaut:
This PR adds config entry to allow user to configure behavior on search exit,
so user can change in which mode to switch after exit from search


christianparpart (commented):
LGTM

Issue #940, opened 2022-12-22
Sampling of resolution is taking the wrong screen as baseline


whisperity:
I have three screens. The ones on the side are `1024×768` and the one in the middle is `1920×1080`. The primary screen (where the application launcher and the task bar is) is the **middle** screen (the one that is `1920×1080`).

```
$ xrandr --listmonitors
Monitors: 3
 0: +*HDMI-1 1920/531x1080/299+1024+0  HDMI-1
 1: +DP-3-2-1 1024/271x768/203+2944+0  DP-3-2-1
 2: +DP-3-2-2 1024/271x768/203+0+0  DP-3-2-2
```

Let's suppose now I start Contour in this configuration. What is observed, is that the window, when compared against the entire viewport, is of a really weird and unusual shape:

![Screenshot_20221222_180759](https://user-images.githubusercontent.com/1969470/209188799-a7d221ab-b83f-4df3-88f1-a51ded466d37.png)

The relevant log entries:

```
[2022-12-22 18:08:13.356086] [gui.display] TerminalWindow.resizeEvent: size 200x100 (200x100)
[2022-12-22 18:08:13.365841] [gui.display] initializeGL: size=186x100, scale=1
[2022-12-22 18:08:13.365871] [gui.display] Assigning session to terminal widget: shell=/bin/zsh, terminalSize=80x24, fontSize=12pt, contentScale=1
[2022-12-22 18:08:13.397024] [gui.display] Loading vertex shader: :/contour/display/shaders/text.vert
[2022-12-22 18:08:13.399889] [gui.display] Loading fragment shader: :/contour/display/shaders/text.frag
[2022-12-22 18:08:13.400112] [gui.display] Loading vertex shader: :/contour/display/shaders/background_image.vert
[2022-12-22 18:08:13.400131] [gui.display] Loading fragment shader: :/contour/display/shaders/background_image.frag
[2022-12-22 18:08:13.400248] [gui.display] Loading vertex shader: :/contour/display/shaders/background.vert
[2022-12-22 18:08:13.400260] [gui.display] Loading fragment shader: :/contour/display/shaders/background.frag
[2022-12-22 18:08:13.400578] [gui.display] configureAtlas: 1024x2048 RGBA
[2022-12-22 18:08:13.472936] [gui.display] Blurring background image: /home/whisperity/Pictures/Wallpapers/world-of-warcraft-lexith-warcraft-elf-devushka-night-elf.jpg
[...]

PR #1521, opened 2024-06-11
Do screen sampling to choose the bigger one

Yaraslaut:
Closes https://github.com/contour-terminal/contour/issues/940

It seems that we can not know at which window we were spawned, so we use bigger screen to setup initial state


src/contour/display/TerminalDisplay.h
```diff
@@ -44,6 +44,7 @@ class TerminalDisplay: public QQuickItem
     Q_PROPERTY(TerminalSession* session READ getSessionHelper WRITE setSession NOTIFY sessionChanged)
     Q_PROPERTY(QString profile READ profileName WRITE setProfileName NOTIFY profileNameChanged)
     Q_PROPERTY(QString title READ title NOTIFY titleChanged)
+    Q_PROPERTY(QQuickWindow* window READ window)
```
christianparpart, line 47: I think this is not used. Why are we adding it? Is it a leftover of some test code?

Issue #663, opened 2022-05-01
Underline thickness should be overridable per config


uspasojevic96:
## Abstract
When font size is reduced, underline thickness stays the same

## Motivation
Add config option to set underline thickness

Issue #1242, opened 2023-10-09
Incorrect letter spacing when using SF Mono


cqexbesd:
### Contour Terminal version

0.4.0-master-616315cc

### Installer source

Github: source code cloned

### Operating System

Arch Linux as of approx a week ago

### Architecture

x86-64

### Other Software

SF Mono font, installed from aur as `otf-san-francisco-mono`
bsdtar, installed from `libarchive`

### Steps to reproduce

`man bsdtar`

### Actual Behavior

Spacing of letters is not correct. See screen shot. Looks OK using the same font in RoxTerm, but also using Contour with the `LiberationMono` font.

Contour:
![image_480](https://github.com/contour-terminal/contour/assets/6312149/8915272c-d2a1-4d73-8dea-1881a6c2472d)

RoxTerm:
![image_480](https://github.com/contour-terminal/contour/assets/6312149/8c51a4f4-a19e-4dfc-93d2-526fda5b2b97)

### Additional notes

@christianparpart already knows the cause and the fix...I can't explain it though!

christianparpart:
As discussed locally. This shows two problems.

1. `SF Mono` font probably does not expose the correct metrics to shift the pen to the right. It looks like it's always 0. This goes in line with the other Apple's font for emoji, where we already auto-center in case of zero-metrics.
2. the underline is too bold for small grid cell heights (IIRC, @cqexbesd is using 12pt size on a standard 96dpi screen)

uspasojevic96:
The underline issue i told you about last year as well when i was on my  imac

PR #1522, opened 2024-06-12
Allow thickness of underline to be 1

Yaraslaut:
This PR makes a small refactor if DecorationRendere.cpp and allows underline thickness to be equal to 1.
I think that this can close https://github.com/contour-terminal/contour/issues/663
And refs https://github.com/contour-terminal/contour/issues/1242

### DIFF
--- a/src/vtrasterizer/DecorationRenderer.cpp
+++ b/src/vtrasterizer/DecorationRenderer.cpp
@@ -139,16 +139,15 @@ auto DecorationRenderer::createTileData(Decorator decoration, atlas::TileLocatio
             // TODO (default to Underline for now)
             [[fallthrough]];
         case Decorator::Underline: {
-            auto const thicknessHalf = max(1u, unsigned(ceil(underlineThickness() / 2.0)));
-            auto const thickness = thicknessHalf * 2;
-            auto const y0 = max(0, (underlinePosition() - static_cast<int>(thicknessHalf)));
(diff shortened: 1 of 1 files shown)


src/vtrasterizer/DecorationRenderer.cpp
```diff
@@ -139,16 +139,15 @@ auto DecorationRenderer::createTileData(Decorator decoration, atlas::TileLocatio
...
+            auto const thickness = max(1u, unsigned(ceil(underlineThickness() / 2.0)));
+            auto const y0 = max(0, (underlinePosition() - static_cast<int>(thickness)));
             auto const height = vtbackend::Height(y0 + thickness);
             auto const imageSize = ImageSize { width, height };
             return create(imageSize, [&]() -> atlas::Buffer {
                 auto image = atlas::Buffer(imageSize.area(), 0);
                 for (unsigned y = 1; y <= thickness; ++y)
-                    for (unsigned x = 0; x < unbox(width); ++x)
-                        image[(*height - y0 - y) * *width + x] = 0xFF;
+                    for (auto x: crispy::times(unbox(width)))
```
christianparpart, line 149: actually, I think we can use `std::views::iota()` by now. what do you think? (also, `for (auto const x: ...` rather than `for (auto x: ...`.

Analog the equivalent parts below. :)

christianparpart:
we agreed to leave it like this, because std::ranges::iota requires a start parameter, which we do not need here.

christianparpart (approved):

PR #1525, opened 2024-06-20
Fixes application window icon on (KDE) Wayland

christianparpart:
Contour's app icon in the window title bar (on KDE plasma on Wayland) was showing a default W-icon rather than our own one. This fixes it.


src/contour/Config.h
```diff
@@ -313,7 +313,7 @@ struct TerminalProfile
     ConfigEntry<bool, documentation::MouseHideWhileTyping> mouseHideWhileTyping { true };
     ConfigEntry<bool, documentation::SeachModeSwitch> searchModeSwitch { true };
     ConfigEntry<vtbackend::LineOffset, documentation::CopyLastMarkRangeOffset> copyLastMarkRangeOffset { 0 };
-    ConfigEntry<std::string, documentation::WMClass> wmClass { "contour" };
+    ConfigEntry<std::string, documentation::WMClass> wmClass {};
```
Yaraslaut, line 316: You need to change it to `"org.contourterminal.Contour"` then, since we use this value to generate default config

christianparpart: thx. fixed.

PR #1537, opened 2024-06-24
[clang-tidy] Improvements

Yaraslaut:
Fix several clang-tidy checks:
* `performance-enum-size`
* `modernize-use-constraints`
* `readability-avoid-nested-conditional-operator`
* `bugprone-narrowing-conversions`
And made them error on warnings

### DIFF
--- a/src/.clang-tidy
+++ b/src/.clang-tidy
@@ -9,6 +9,7 @@ Checks: >-
   -bugprone-reserved-identifier,
   -bugprone-suspicious-include,
   -bugprone-unchecked-optional-access,
+  -bugprone-chained-comparison,
   clang-analyzer-core.*,
   clang-analyzer-cplusplus.*,
   clang-analyzer-deadcode.*,
@@ -66,12 +67,12 @@ Checks: >-
   -readability-use-anyofallof,
   misc-const-correctness,
 WarningsAsErrors: >-
+  performance-enum-size,
   readability-identifier-naming,
   misc-const-correctness,
   modernize-use-nullptr,
 UseColor: true
-HeaderFilterRegex: 'src/(contour|crispy|text_shaper|vtpty|vtparser|vtbackend|vtrasterizer)/'
-AnalyzeTemporaryDtors: false
+HeaderFilterRegex: "contour\/src\/(.*)h"
 FormatStyle:     none
 CheckOptions:
   - key:             bugprone-easily-swappable-parameters.MinimumLength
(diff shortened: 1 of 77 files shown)


src/crispy/result.h
```diff
@@ -146,13 +151,13 @@ class result
...
+    template <std::enable_if_t<!std::is_same_v<T, void>, int> = 0> //NOLINT
     [[nodiscard]] constexpr const T& operator*() const& noexcept { return std::get<value_type>(_data); }
-    template <std::enable_if_t<!std::is_same_v<T, void>, int> = 0>
+    template <std::enable_if_t<!std::is_same_v<T, void>, int> = 0> //NOLINT
     [[nodiscard]] constexpr T& operator*() & noexcept { return std::get<value_type>(_data); }
-    template <std::enable_if_t<!std::is_same_v<T, void>, int> = 0>
+    template <std::enable_if_t<!std::is_same_v<T, void>, int> = 0> //NOLINT
     [[nodiscard]] constexpr const T&& operator*() const&& noexcept { return std::move(std::get<value_type>(_data)); }
-    template <std::enable_if_t<!std::is_same_v<T, void>, int> = 0>
+    template <std::enable_if_t<!std::is_same_v<T, void>, int> = 0> //NOLINT
```
christianparpart, line 160: wasn't it the idea to get rid of SFINAE (std::enable_if)? Isn't it possible to add a `requires` modifier to these member functions, such that they're only selected for function resolving when the requires clause is met? 🤔

src/.clang-tidy
```diff
@@ -64,14 +65,18 @@ Checks: >-
...
 WarningsAsErrors: >-
+  performance-enum-size,
   readability-identifier-naming,
   misc-const-correctness,
   modernize-use-nullptr,
+  modernize-use-constraints,
 UseColor: true
-HeaderFilterRegex: 'src/(contour|crispy|text_shaper|vtpty|vtparser|vtbackend|vtrasterizer)/'
-AnalyzeTemporaryDtors: false
+HeaderFilterRegex: "contour/src"
```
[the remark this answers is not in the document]
Yaraslaut: `directory does not necessarily need to be named contour.` Oh, indeed, will change it back

src/crispy/result.h
Yaraslaut: We can add requires, but we need template instantiation mechanics for it, so here we would need to create dummy template argument

christianparpart (approved):
thank you :)

PR #1538, opened 2024-06-25
Restrict font locator configurability down to `native` and `mock`.

christianparpart:
I mainly wanted to get rid of the fontconfig dependency on Windows platform. There's no real need for it, especially since we have a DirectWrite font locator implementation. I'd rather focus on improving that one (if needed) than supporting two different one on a single platform.


src/contour/Config.cpp
```diff
@@ -1069,23 +1069,18 @@ void YAMLConfigReader::loadFromEntry(YAML::Node const& node,
...
-            return vtrasterizer::FontLocatorEngine::FontConfig;
-        if (literal == "coretext")
-            return vtrasterizer::FontLocatorEngine::CoreText;
-        if (literal == "dwrite" || literal == "directwrite")
-            return vtrasterizer::FontLocatorEngine::DWrite;
+        for (auto const& deprecated: { "fontconfig", "coretext", "dwrite", "directwrite" })
+        {
+            if (literal == deprecated)
+            {
+                errorLog()("Setting font locator to \"{}\" is deprecated. Using \"native\" instead.", literal);
```
Yaraslaut, line 1080:
```suggestion
                errorLog()("Setting font locator to \"{}\" is deprecated. Use \"native\" instead.", literal);
```
I think we need to ask user to change config file to get rid of this error

vcpkg-configuration.json
Yaraslaut, line 1: What is this file for?

src/text_shaper/font_locator_provider.cpp
```diff
@@ -18,32 +18,20 @@ font_locator_provider& font_locator_provider::get()
...
-    return *_directwrite;
-}
+    if (!_native)
+    {
+#if defined(__APPLE__)
+        _native = make_unique<coretext_locator>();
+#elif defined(_WIN32)
+        _native = make_unique<directwrite_locator>();
+#else
+        _native = make_unique<fontconfig_locator>();
```
Yaraslaut, line 30: Can we move it into  constructor?

Yaraslaut (commented):
Some small things

src/contour/contour.yml
```diff
@@ -370,9 +370,7 @@ profiles:
             # This is implicitly also responsible for font fallback
             # Possible values are:
             # - native          : automatically choose the best available on the current platform
-            # - fontconfig      : uses fontconfig to select fonts
-            # - CoreText        : uses macOS CoreText to select fonts.
-            # - DirectWrite     : selects DirectWrite engine (Windows only)
+            # - mock            : a mock locator that does not actually locate any fonts (for development testing)
```
Yaraslaut, line 373: This file only for reference, we will get rid of it in the future. Primary place for config documentation is in `ConfigDocumentation.h`

vcpkg-configuration.json
Yaraslaut: Ah, alright, see from commit message

src/text_shaper/font_locator_provider.cpp
christianparpart: Hmm.... on CI we don't call `native()` but `mock()` and thus, this would usually never get instanciated. Moving it to the constructor, would do so, howeve.r Not sure we want that (I think, as of today, it would not hurt though)

src/contour/contour.yml
christianparpart: done

Yaraslaut (approved):
thanks

christianparpart:
I just rebased on top of latest master.

Issue #1442, opened 2024-01-30
Add key bindings documentation to the web page


Yaraslaut:
We currently have information regarding key bindings available in the configuration [file](https://github.com/contour-terminal/contour/blob/master/src/contour/contour.yml). However, there is a lack of corresponding details on our website, http://contour-terminal.org/. We need to update the web page to include this information.

saliaku:
on it :+1:  please do let me know which all file are to be changed

Yaraslaut:
> on it 👍 please do let me know which all file are to be changed @Yaraslaut

for key bindings we need to create a separate file similar to [profiles](https://github.com/contour-terminal/contour/blob/master/docs/configuration/profiles.md)

saliaku:
yaa i will look into this, would appreciate if I could get some more help :bow:

christianparpart:
@saliaku maybe it helps if you read through the existing documentation (on source level). It's not much, we don't have much yet, so it shouldn't take long at all.

This should give you a good understanding on how to extend it. 

On top of that, feel free to ask questions on Discord, in `#contour-dev` or `#contour` chat. :)

Yaraslaut:
@saliaku  is there any updates? do you need any help?

PR #1543, opened 2024-06-26
Add key map documentation generation

Yaraslaut:
Closes https://github.com/contour-terminal/contour/issues/1442

### DIFF
--- a/src/contour/Actions.h
+++ b/src/contour/Actions.h
@@ -136,6 +136,166 @@ using Action = std::variant<CancelSelection,
 
 std::optional<Action> fromString(std::string const& name);
 
+namespace documentation
+{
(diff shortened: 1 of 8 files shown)


src/contour/Actions.h
```diff
@@ -136,6 +136,156 @@ using Action = std::variant<CancelSelection,
 
 std::optional<Action> fromString(std::string const& name);
 
+namespace documentation
+{
+
+    constexpr std::string_view CancelSelection { "Cancels currently active selection, if any." };
```
christianparpart, line 142:
```suggestion
    constexpr inline std::string_view CancelSelection { "Cancels currently active selection, if any." };
```

Don't you think they should all be marked `inline`, because they're in the header? 🤔

christianparpart (approved):
Very nice. Many thanks, @Yaraslaut :)

Issue #808, opened 2022-09-08
A bind-able action to disable autoscroll while the action is active


Marenz:
Implement an bindable action that can be used to temporarily disable autoscroll so one can investigate output of a process that is still outputting more lines.

more generic alternative approach: add an action that, while active/key pressed down, switches another configured option.

Yaraslaut:
I think that this can be a default behavior when switching into normal mode. 
If  process is writing something and you go to normal mode, scrolling will stop

PR #1544, opened 2024-06-26
Stop scroll when switching to normal mode

Yaraslaut:
I think this is a sane default behavior that allows you to investigate the output from the command
Closes https://github.com/contour-terminal/contour/issues/808


christianparpart (approved):
I think this might be sufficient. It was just not obvious when reading the diff. Thank you :)

christianparpart:
force-push included a rebase on top of latest master, and merge fix to metainfo.xml :)

christianparpart (approved):
thank you :)

Issue #629, opened 2022-03-26
When spawned with a directory, interpret it as the starting cwd


whisperity:
When Contour is started from a desktop integration (e.g. right click -> open with -> contour), unfortunately the integration only fires `contour [... directory path ...]`. This causes contour to crash because a directory cannot be started as a program image.

The solution would either be:

 1. Implement desktop integrations that work with the right-click menu. This might be problematic due to the absolute plethora of desktop environments to support.
 2. "Hotwire" Contour so if it is spawned with only a directory as the first argument, interpret it as "start with the default profile but spawn the configured shell **in that directory**".

For example, Konsole does not crash (at least...), but is also rather buggy: when opening, from Dolphin, Konsole in `~/foo/bar` as the "Open with..." action, it spawns the default shell in `~/foo`.

uspasojevic96:
Oh this is interesting, I can look into this one later, the whole topic of .desktop files is convoluted

christianparpart:
This could be fixed by being a little bit more smart and check the file modes to decide if it is supposed to be the CWD or shell.

christianparpart:
@uspasojevic96 Mind, I broke the master branch (RGB colors) last night, I will fix that tonight. Sorry.

PR #1545, opened 2024-06-26
Handle different arguments for contour executable

Yaraslaut:
Closes https://github.com/contour-terminal/contour/issues/629

### DIFF
--- a/src/contour/ContourGuiApp.cpp
+++ b/src/contour/ContourGuiApp.cpp
@@ -269,7 +269,27 @@ bool ContourGuiApp::loadConfig(string const& target)
         }
         else
         {
-            shell.program = flags.verbatim.front();
+            auto frontCommand = flags.verbatim.front();
+
+            // check if this is a file
+            if (fs::exists(frontCommand) && fs::is_regular_file(frontCommand))
+            {
+                // check if this is an executable file
+                if ((fs::status(frontCommand).permissions() & fs::perms::owner_exec) != fs::perms::none)
+                    shell.program = frontCommand;
+                else // find a path to file and open shell in this path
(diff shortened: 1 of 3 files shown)


src/contour/ContourGuiApp.cpp
```diff
@@ -269,7 +269,27 @@ bool ContourGuiApp::loadConfig(string const& target)
...
+                else // find a path to file and open shell in this path
+                    shell.workingDirectory = fs::path(frontCommand).parent_path();
+            }
+            else if (fs::exists(frontCommand) && fs::is_directory(frontCommand))
+            {
+                shell.workingDirectory = fs::path(frontCommand);
+            }
+            else
+            {
+                errorLog()("Do not know what to do with `{}` will use it as a program", frontCommand);
```
christianparpart, line 289:
```
   p - execlp(), execvp(), execvpe()
       These  functions  duplicate  the  actions of the shell in searching for an executable file if the specified filename does not
       contain a slash (/) character.  The file is sought in the colon-separated list of directory pathnames specified in  the  PATH
       environment  variable.   If  this  variable isn't defined, the path list defaults to a list that includes the directories re‐
       turned by confstr(_CS_PATH) (which typically returns the value "/bin:/usr/bin") and possibly also the current working  direc‐
       tory; see NOTES for further details.
```

Looks like we should allow non-fully-qualified paths. Because we use `execvp` and that's totally fine with, say, executing `ls`, resolving to `/bin/ls` iff found in $PATH.

I think what would work best is:

1. check if it's a (relative or absolute) directory path, then take this as working directory and default shell

2.) alternatively assume it's a program, give execvp (through Process) a try as usual. we'll fall-back to configured login shell otherwise.

christianparpart (approved):

PR #1548, opened 2024-06-27
add OpenBSD support

jg1uaa:
## Description

add OpenBSD support, as described in the title.

## Motivation and Context

Simply I want to use Contour-terminal on OpenBSD.
I have to test sj3 (Japanese Kana-Kanji converter) tty client with status-line capable terminal emulator, and I found contour.

## How Has This Been Tested?

Tested with OpenBSD-current/amd64 and libunicode/boxed-cpp (by private ports).

To check changes are not harm for other platforms, I also build modified code on Void Linux.
Both OpenBSD and Void Linux could not support Qt6. Only Qt5 works.
 
## Checklist:

- [ ] I have read the [**`CONTRIBUTING`**](https://github.com/contour-terminal/contour/blob/master/CONTRIBUTING.md) document in my spoken language, and understand the terms

well, it is not the time to put 'x' in all the boxes yet.
CONTRIBUTING.md is 404, early to add OpenBSD support to documents, and test method is simply "pass the compile" or not.


christianparpart:
Wow. This is awesome! Manyh thanks, @jg1uaa for your very valuable contribution. I will have a closer look ASAP. :)

Yaraslaut:
Hi @jg1uaa
As you can see from actions: this changed breaking some other platform, the reason being is that libxcb installed separately and not as a part of X11, so you need to distinguish between openbsd and all other os, you can do something like this:
```
if(${CMAKE_SYSTEM_NAME} MATCHES "OpenBSD")
    set(OPENBSD TRUE)
endif()
```
and then 
```
        if(NOT(WIN32) AND NOT(APPLE) AND NOT(OPENBSD))
            target_link_libraries(contour Qt5::X11Extras xcb)
        endif()
        if(OPENBSD)
            target_link_libraries(contour Qt5::X11Extras X11::xcb)
        endif()
```

Yaraslaut:
Also, if you want us to keep an eye on openbsd builds you can add openbsd github action using: 
https://github.com/vmactions/openbsd-vm
And we already have one for freeBSD
https://github.com/contour-terminal/contour/blob/4c71d0cce2e63a23859a1cf7b389dc9e872384a1/.github/workflows/build.yml#L288-L321

jg1uaa:
I am planning to use libunicode and boxed-cpp as OpenBSD-ports and today I posted them.
https://marc.info/?l=openbsd-ports&m=171961710727187&w=2
https://marc.info/?l=openbsd-ports&m=171961232324802&w=2

If they are added to ports (and released with new revision, OpenBSD-7.6), build.yml and install-deps.sh can be fixed.

christianparpart:
@jg1uaa May thanks.

I think there's not much left.

Would you mind adding a changelog entry to metainfo.xml? Because that's actually a good thing worth mentioning. For example at around:

https://github.com/contour-terminal/contour/blob/4c71d0cce2e63a23859a1cf7b389dc9e872384a1/metainfo.xml#L107-L112

And secondly, would you mind adding the bits for ensuring we keep building on openBSD by adding it to build.yml? It is probably best done by somebody who knows openBSD best. We really value this contribution. I think it's best preserved by making sure we retain support in the future as good as it would work today :)

jg1uaa:
libunicode and boxed-cpp is now at OpenBSD ports. Avilable for OpenBSD-current and future coming OpenBSD-7.6.

http://cvsweb.openbsd.org/ports/devel/boxed-cpp/
http://cvsweb.openbsd.org/ports/textproc/libunicode/

to test on vmactions, we have to wait release of OpenBSD-7.6, fix termbench-pro (it has <format> related problem) and prepare some stuff (build.yml, metainfo.xml, install-deps.sh).

Yaraslaut:
Thanks for the update, `termbench-pro` is optional dependency and can be disregarded for the packaging I think. We have new release coming soon so we can switch to packaged versions of `libunicode` and `boxed-cpp` later, `install-deps.sh` script will fetch them locally when needed, so i think that we can create action to build contour to merge this PR without the need to wait for next release

Yaraslaut:
@jg1uaa  I added openbsd github action and added info about openbsd support, do you want to add something to this PR? otherwise we can merge it

jg1uaa:
currently there is nothing to add this PR. thanks!

PR #1552, opened 2024-07-01
[contour] Config: Fix default foreground color loading (was not read anymore)

christianparpart:
did we have the config PR in the last release already? if not, no changelog needed :)


Yaraslaut (approved):
Thanks for fixing this :)

Issue #1448, opened 2024-02-01
yW (yank WORD) not working properly in normal mode


ferdinandyb:
### Contour Terminal version

 Contour Terminal Emulator 0.4.2.6408-prerelease

### Installer source

Github: CI actions

### Operating System

Windows 11

### Architecture

x86-64

### Steps to reproduce

Run something like `echo "hello 1236"`, enter normal mode, search for `hello` and then yank WORD with `yW`.

### Expected Behavior

I would expect "hello" to be yanked.

### Actual Behavior

"hello 1" is yanked

christianparpart:
@ferdinandyb I think there's a misunderstanding.

- `w` refers to the beginning of the next word (using word delimiters, such as `.`, `,`, `:`, space, etc..
- `W` refers to the beginning of the next word (using whitespace delimiters only)
- `e` refers to the end of the current word (delimiters as in `w`)
- `W` refers to the end of the current word (delimiters as in `W`, i.e. whitespaces only)

So pressing `yW` will yank everything from the current cursor position to the beginning of the next word whereas `yE` (and similarily `ye`) will yank from current position to the end of the current word.

p.s.: I was implementing this based on what I was reading in the vim documentation. However, trying it out right now in vim (and Contour), I noticed a different in vim itself.

vim itself behaves differently in marking the range for `yW` vs `vW`. This is a WTF for me. I may have accidentally skipped reading something in the docs that would describe this, but to me that sounds a bit weird. Not sure sure why `yW` and `vW` in vim (neovim for me) do operate on a different range, that is, `yW` excludes the right boundary whereas `vW` does indeed include it (as we do in Contour, however, for both cases).

So not sure how to process on this ticket. If vim and neovim are doing it that way, then we should adapt of course. Because it's meant to be inspired by vim. It should feel the same then.

ferdinandyb:
> @ferdinandyb I think there's a misunderstanding.
>
> - `w` refers to the beginning of the next word (using word delimiters, such as `.`, `,`, `:`, space, etc..
> - `W` refers to the beginning of the next word (using whitespace delimiters only)
> - `e` refers to the end of the current word (delimiters as in `w`)
> - `W` refers to the end of the current word (delimiters as in `W`, i.e. whitespaces only)
>
> So pressing `yW` will yank everything from the current cursor position to the beginning of the next word whereas `yE` (and similarily `ye`) will yank from current position to the end of the current word.
>
[...]

PR #1555, opened 2024-07-01
Fix selection of w/W vim motions

Yaraslaut:
Closes https://github.com/contour-terminal/contour/issues/1448

PR fixes selection for w/W vim motions

### DIFF
--- a/src/vtbackend/ViCommands.cpp
+++ b/src/vtbackend/ViCommands.cpp
@@ -320,7 +320,9 @@ void ViCommands::executeYank(ViMotion motion, unsigned count)
         }
         default: {
             auto const [from, to] = translateToCellRange(motion, count);
-            executeYank(from, to);
+            // motion is inclusive but for yank we want to exclude the last cell which is the first cell of
+            // the next word
+            executeYank(from, { to.line, to.column - 1 });
         }
         break;
     }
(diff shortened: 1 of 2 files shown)


src/vtbackend/ViCommands.cpp
```diff
@@ -1033,7 +1034,7 @@ CellLocation ViCommands::translateToCellLocation(ViMotion motion, unsigned count
                 --count;
             }
 
-            return result;
+            return { result.line, result.column - 1 };
```
christianparpart, line 1037: Hmm... While fixing it, would it make sense to add a test for it to `ViCommands_test.cpp`?

Is it possible that `result.column` could be `0` already and then subtracting `1` might be an issue?

Yaraslaut: I will try to add a test for it, and also i figured out that we need this change only when we yank, so if previous command was not yank we should not go back by one character

christianparpart: yeah, this is the thing with inclusive vs exclusive motions :)

Yaraslaut: Now it is fixed

christianparpart (dismissed):
Thank you, @Yaraslaut :)

Issue #1374, opened 2023-12-20
Assign unicode ranges to fonts


classabbyamp:
## Abstract

nerd-fonts are a common way to add special symbols in terminal UIs (shell prompts, editors, etc). There are two ways this is implemented, a patched font or a symbols-only font. I prefer to use the symbols-only font because it allows for more flexibility and I don't need to install 2GB worth of duplicate fonts, but it does mean some form of fallback mechanism is necessary. It seems that currently this does partially work:

![image](https://github.com/contour-terminal/contour/assets/5366828/e189ed3d-4422-480c-baee-0036832af1ad)

(kitty left, contour right; notice the incorrect symbol for the `>`)

## Motivation

To ensure that certain ranges of codepoints have the correct font, I propose a configuration option that allows for assigning ranges to certain fonts. I suspect this assignment might already be possible under-the-hood, given the `font.emoji` config option.

## Specification

My thought is something like:
```yaml
font:
  symbols:
    - range: U+E5FA-U+E62B
      family: "Symbols Nerd Font"
    - range: U+E700-U+E7C5,U+F000-U+F2E0
      family: "Symbols Nerd Font"
```

## Other Implementations

https://sw.kovidgoyal.net/kitty/conf/#opt-kitty.symbol_map

christianparpart:
Ranges for custom assignment of codepoints to glyphs is AFAIK only available on Kitty, right? Or do we have other TEs allowing that as well?

If I am not mistaken, on Linux at least, you could use fontconfig to achieve the same.
Googling around, I found this: https://stackoverflow.com/questions/47501411/how-to-set-font-per-unicode-range-codepoint-in-fontconfig 
I'd have to try it myself, but I think that would be a proper way to do it. Plus point: it would work on **every** application automatically then.

On the other hand, we planned to integrate "Symbols Nerd Font" (and others) for convenience into Contour natively anyways (#170).

classabbyamp:
> Ranges for custom assignment of codepoints to glyphs is AFAIK only available on Kitty, right? Or do we have other TEs allowing that as well?

not sure, I've used kitty for a while and don't know the whole landscape.

> If I am not mistaken, on Linux at least, you could use fontconfig to achieve the same.

I was going to say I already had a fontconfig conf for this but it seems [it doesn't do that](https://github.com/ryanoasis/nerd-fonts/blob/master/10-nerd-font-symbols.conf). Will try the fontconfig conf method.

0x5c:
[...]

Issue #225, opened 2021-05-29
Configurable font fallback


christianparpart:
It is said that `fontconfig` should be used for font fallback, but it seems that terminal emulators are not always using this recommended API for font fallback.

* kitty uses fontconfig only for retrieving the set of all fonts, then filters and prioritizes manually)
* wezterm seems to only support font fallback based on what the user has provided in the configs.
* What is konsole, gnome-terminal and alacritty doing?
* Not all fontconfig system configurations seem to be reliable (if any at all).

On top of using fontconfig for getting the list of fallback fonts maybe we could provide a manually provided list of font fallbacks too.

* Evaluate the usefulness of selecting font files based on codepoints / codepoint sequences (such as kitty does provide)
* Evaluate embedding Powerline glyphs (should be configurable, too)
* Evaluate embedding box drawing glyphs (should be configurable, too)

Yaraslaut:
@eproxus  Thanks for the interest
I will add this into milestone for 2024 since I wanted to disable font fallback and if memory is correct  @Utkarsh-khambra as well

Yaraslaut:
> > I will add this into milestone for 2023 since I wanted to disable font fallback
> 
> You mean rely on fontconfig only? Will this work on macOS?

Yes, only fontconfig, this will be naturally configurable, so if this does not work for macOS this option will not be available then. And for sure we will add other options for fallback. something like :
`auto` <- current behavior  
`priority` <- list of fonts that we need to check first and if not found proceed with the rest 
`list` <- user specified list of fonts that are used for fallback and not others
`none` <- do not try to find another font

Yaraslaut:
Closing this one, now configuration allows some flexibility for font fallback and builtint fonts will be implemented in https://github.com/contour-terminal/contour/issues/170

PR #1557, opened 2024-07-02
Add config option to configure font fallback

Yaraslaut:
Refs https://github.com/contour-terminal/contour/issues/225
Refs https://github.com/contour-terminal/contour/issues/1374
PR adds configuration option to disable font fallback 

Implemented following logic: fallback entry in the config has several options: `auto` `none` or a list of fonts `[ "A" , "B" ]`
`auto` will insert all fonts in fallback list
`none` will make fallback list empty 
`list` will  put in fallback list fonts in the same order as they appear in configuration given that they are found


src/text_shaper/open_shaper.cpp
```diff
@@ -677,8 +678,14 @@ void open_shaper::shape(font_key font,
         logMessage.append("Using font: key={}, path=\"{}\"\n", font, identifierOf(fontInfo.primary));
     }
 
-    if (_d->tryShapeWithFallback(
-            font, fontInfo, hbBuf, hbFont, script, presentation, codepoints, clusters, result))
+    auto shapingFunction = [&](auto&&... args) {
+        if (useFontFallback)
```
christianparpart, line 682: Hmmm 🤔 
Is this really useful, to be able to completely turn off font fallback?

Even if, we should not do something like `if (useFontFallback) ... else ...` in the rendering hot path, but rely on function pointers / virtual dispatch instead. I know this sounds like a micro optimization, but would be cleaner IMHO on top of it, too.

Ignoring that, I think we could instead promote the `mock` font locator instead, because there you can explicitly declare how many fonts (font files) to pick in order for fallback. If one would not specify any more than a single font, that would be the same as having no font fallback at all.

But none of it is actually addressing the point in #225.

We could instead work on embedding fonts (Nerd Fonts and standard emoji fonts), which would help people like @eproxus to not require font patching (nerd font), or always come with the same experience when enabling something like `use_builtin_emoji: true` (or `emoji: builtin` or not specifying `emoji:` key at all, indicating that the builtin should be used)

Yaraslaut: I updated PR, now we configure fallback on the stage of font_locator and also can specify list of fonts that use for fallback

christianparpart (approved):
Thank you, @Yaraslaut.

In the long run, we should try to also extend the documentation as we go with the code changes. This is a user-facing configuration change, which would fit nicely into our website's documentation as well :)

PR #1558, opened 2024-07-02
Improves handling of constant bell sound spawning

Yaraslaut:
Do not spawn bell sound when bell is playing, it allows to handle bell signals much faster in the case when the bell is constantly spawning


christianparpart (approved):
Thank you

PR #1567, opened 2024-07-15
Fix copy of multi-line selection

Yaraslaut:
PR fixes a bug when copying a selection of a wrapped line. 
Previously new-line was inserted without the need when copying the selection span across few grid-line, when they represent  one logical lines that is wrapped due to terminal size


christianparpart (approved):
nice catch, also thanks for the test case. :)

PR #1574, opened 2024-07-21
fix error, "" is an invalid MIME type ("" does not contain a subtype)

jg1uaa:
## Description

fix `Error in file "/usr/local/share/applications/org.contourterminal.Contour.desktop": "" is an invalid MIME type ("" does not contain a subtype)` at install


Yaraslaut:
Thanks for fixing it @jg1uaa

PR #1581, opened 2024-07-31
Add missing keymapping for numpad during config load

Yaraslaut:
Closes https://github.com/contour-terminal/contour/issues/1325


Glog78:
image

I guess 2 times Numpad_Add is not needed :) just because i was curious and thank you for the fix

Yaraslaut:
> image
> 
> I guess 2 times Numpad_Add is not needed :) just because i was curious and thank you for the fix

oh, indeed, i first checked that numpad_add is working and then added all entries :)  thanks for checking this out and good that issue is resolved now.

PR #1591, opened 2024-09-16
install-deps.sh: Flip curl/wget priority to favor curl over wget, because at least Fedora 40 Linux seems to complain on wget

christianparpart:
at least my Fedora 40 on my new Linux laptop instantly started to complain on a fresh Fedora install. I sadly don't seem to remember what exactly the error message was ;(

Also, added a commit to streamline all upload-artifact actions to v4 (we had 2 at v3 and v2 still)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -224,7 +224,7 @@ jobs:
           docker cp contour_static:/contour/build/src/contour/contour contour
           docker container rm contour_static
       - name: "Uploading static executable"
-        uses: actions/upload-artifact@v3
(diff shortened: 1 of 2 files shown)


Yaraslaut (dismissed):
I remember that we switched it few times already xD 
We can keep it this way given existing support for CPM

christianparpart:
> I remember that we switched it few times already xD

I wonder why we did not mention anywhere why we switched. I think curl is the way to go in the future anyways.

Also, added a commit to streamline all upload-artifact actions to v4 (we had 2 at v3 and v2 still)

Issue #1598, opened 2024-09-21
Error when starting contour.


Dooez:
### Contour Terminal version

0.5.0

### Installer source

something else (please specify below)

### Operating System

Arch Linux

### Architecture

x86-64

### Steps to reproduce

Install from AUR.

### Actual Behavior

contour does not launch and returns an error:
`Unhandled error caught. cannot switch from manual to automatic argument indexing`

Yaraslaut:
@ptr1337 maybe you can take a look?

whisperity:
Can reproduce on macOS Sequioa 15.0 as well:

```
whisperity@mac ~ % /Applications/contour.app/Contents/MacOS/contour
Unhandled error caught. cannot switch from manual to automatic argument indexing
```

Yaraslaut:
Ok, i can repro it, most likely this is something from fmt lib

Yaraslaut:
fmt broke there api in fmt 11 
https://godbolt.org/z/4xfMdoWa7

Yaraslaut:
Created issue on fmt side https://github.com/contour-terminal/contour/issues/1598

christianparpart:
We want to switch to std fmt soon anyways. That will bring more long term stability.

Dooez:
This behavior seems to be intended and fmt closed the issue.
From my testing i've found that modifying `ConfigDocumentation.h` format strings to include parameter numbers fixes the issue. I am not sure how to test if formatting is performed correctly. 
I can make a PR, but it is a pretty small change.

ptr1337:
We can only build contour against fmt10 currently from the AUR. We might want to adjust this.
But since contour automatically picks the installed fmt version, it would (likely) only in the chroot.

Yaraslaut:
This will be fixed soon

topazus:
@whisperity I tried to build contour from commit of https://github.com/contour-terminal/contour/commit/62b749bc03d5823d2924bc3bd1a6dd2243b72863 with libunicode of https://github.com/contour-terminal/libunicode/commit/42494f9798148c67662a29de73daaf0273b1b1a3, it built with the following error. see: https://download.copr.fedorainfracloud.org/results/topazus/test/fedora-rawhide-x86_64/08090302-contour-terminal/builder-live.log.gz

```bash
[...]

PR #1600, opened 2024-09-22
Maintenance/stdlib

christianparpart:
### CHECKLIST BEFORE MERGE

- [x] resolve review items
- [ ] have dependency git tags updates to their newly released versions

### How:

```sh
# fish shell syntax

for file in (git grep fmt/format | cut -d: -f1); sed -i -e 's,include <fmt/format.h>,include <format>,' $file; end
for file in (git grep fmt:: | cut -d: -f1); sed -i -e 's,fmt::,std::,g' $file; end
for file in (git grep std::print | cut -d: -f1); sed -i -e 's,std::print,std::cout << std::format,g' $file; end

# replace `std::print` with `std::cout << std::format`

## tidy-up
find src \( -name '*.cpp' -o -name '*.h' \) -exec clang-format -i {} \;

```

## misc notes

- check statusline's time in `%H:%S` is working correctly
- some lines wrt. `fmt::runtime()` have been rewritten
- `fmt::arg()` rewritten for `std::format()`
- Also fix timestamps in debug logs

Closes https://github.com/contour-terminal/contour/issues/1598


examples/watch-mouse-events.cpp
```diff
@@ -225,7 +225,7 @@ struct MouseTracker final: public BasicParserEvents
 
         auto const state = _decrpm.value().second;
         auto const supported = state == 1 || state == 2;
-        fmt::print("Passive mouse tracking: {}\n", supported ? "supported" : "not supported");
+        std::cout << std::format("Passive mouse tracking: {}\n", supported ? "supported" : "not supported");
```
Yaraslaut, line 228: If we switch to c++23 we can call `std::print`

scripts/install-deps.sh
```diff
@@ -110,18 +110,24 @@ fetch_and_unpack_termbenchpro()
 
 fetch_and_unpack_boxed()
 {
-    local boxed_cpp_version="1.4.2"
+    local boxed_cpp_git_sha="45e7c2e393dd1644f1ab18fb48ba59910c75aad0"
```
Yaraslaut, line 113: We also have a cmake variable for minimal boxed-cpp version `BOXED_CPP_MINIMAL_VERSION` in `cmake/ContourThirdParties.cmake` file

src/contour/Actions.h
```diff
@@ -304,7 +303,7 @@ inline auto getDocumentation()
 // {{{ fmtlib custom formatters
 #define DECLARE_ACTION_FMT(T)                                                                          \
```
Yaraslaut, line 304: we need to change name for this macro

src/contour/Config.cpp
```diff
@@ -155,6 +155,11 @@ namespace
 #endif
     }
 
+    std::string replaceCommentPlaceholder(std::string const& docString)
+    {
+        return std::regex_replace(docString, std::regex { "\\{comment\\}" }, "#");
+    }
+
```
Yaraslaut, line 162: We might want to customize it for different writers in the future, so we should make it writer type dependent, something like 
```suggestion
    template <typename Writer=YAMLConfigWriter>
    std::string replaceCommentPlaceholder(std::string const& docString)
    {
        return std::regex_replace(docString, std::regex { "\\{comment\\}" }, Writer::comment);
    }

```

or make it a function of Writer type

src/contour/Config.h
```diff
@@ -936,13 +935,15 @@ struct Writer
...
     {
-        return fmt::format("{}", v);
+        return std::format("{}", v);
     }
 
     template <typename... T>
     [[nodiscard]] std::string format(std::string_view doc, T... args)
     {
-        return fmt::format(fmt::runtime(doc), format(args)..., fmt::arg("comment", "#"));
+        std::string replacedDoc =
```
Yaraslaut, line 944: i think we need to use `replaceCommentPlaceholder` function here 
Also, why `vformat`?

src/crispy/CLI.cpp
```diff
@@ -816,9 +816,9 @@ namespace // {{{ helpers
                               help_element::HelpText);
 
                 // {{{ append default value, if any
-                // NB: It seems like fmt::format is having problems with
+                // NB: It seems like std::format is having problems with
                 //     formatting std::variant<>'s on some systems.'
-                // auto const defaultValueStr = fmt::format("{}", option.value);
+                // auto const defaultValueStr = std::format("{}", option.value);
```
Yaraslaut, line 821: Can we delete this comment?

src/vtbackend/StatusLineBuilder.cpp
```diff
@@ -305,7 +304,16 @@ struct VTSerializer
     std::string visit(StatusLineDefinitions::Clock const&)
     {
         crispy::ignore_unused(this);
-        return fmt::format("{:%H:%M}", fmt::localtime(std::time(nullptr)));
+
+        // NB: The following is printing the time in UTC, but we need it in local time.
```
Yaraslaut, line 308: Maybe add `TODO` so we have a chance to return here in the future

Yaraslaut (commented):
Few small things to consider, 
I think that we can switch in this PR to c++23 and use std::print

src/contour/Config.cpp
christianparpart: I think I'll leave potential config API/implementation improvements to you :)

src/contour/Config.h
christianparpart: I tried also without `vformat`, but then the formatting string needs to be a literal (i.e. `std::format_string<Args...>`)

Yaraslaut: Oh, I was hoping that they made it into c++23, but only in c++26 https://en.cppreference.com/w/cpp/utility/format/runtime_format

scripts/ci-install-run-deps.sh
Yaraslaut, line 1: we can use install-deps.sh to get all dependencies, why this file?

Yaraslaut (approved):
Many thanks, I think we can merge it
just one file in scripts not sure that we need it

christianparpart:
Which file? I am in the forest right now. It's hard to check on the phone :)

Yaraslaut:
> Which file? I am in the forest right now. It's hard to check on the phone :)

[scripts/ci-install-run-deps.sh](https://github.com/contour-terminal/contour/pull/1600/files/87bde6b2aa460fe0996f795a24f0e1fb5e1ae427#diff-54a0f34bae984b37737703e9a3775783e925bf718099ec8e83adf15c790face8) :)

scripts/ci-install-run-deps.sh
christianparpart: we call it twice. I'd prefer to maintain the actual dependencies to install in as few locations as ossible :)

PR #1614, opened 2024-10-02
Fix macOS specific key handling

Yaraslaut:
Fixes macOS specific handling of the input
 -  Option used as a Control modifier, to allow Option-Left[Right] jumps between words in the shell

### DIFF
--- a/src/vtbackend/InputGenerator.h
+++ b/src/vtbackend/InputGenerator.h
@@ -54,6 +54,12 @@ using Modifiers = crispy::flags<Modifier>;
 /// @returns CSI parameter for given function key modifier
 constexpr size_t makeVirtualTerminalParam(Modifiers modifier) noexcept
 {
+#if defined(__APPLE__)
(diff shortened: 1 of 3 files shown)


src/vtbackend/InputGenerator.h
```diff
@@ -54,6 +54,12 @@ using Modifiers = crispy::flags<Modifier>;
 /// @returns CSI parameter for given function key modifier
 constexpr size_t makeVirtualTerminalParam(Modifiers modifier) noexcept
 {
+#if defined(__APPLE__)
+    // Use option key as a control modifier to use
+    // Ctrl-Left[Right]Arrow for word navigation.
+    if (modifier == Modifier::Alt)
+        return 1 + Modifier::Control;
```
christianparpart, line 61: @Yaraslaut i've force-pushed my suggestion right into the PR.

The change is to use symbolic names in order to make it more intuitive to read, also with the line (1 + value) below.

PR #1616, opened 2024-10-03
Add example program to detect OS dark/light mode setting in the terminal

christianparpart:
This example .cpp serves as PoC on how other TUI/CLI programs might make use of this to one-off detect dark/light mode OS settings.


Yaraslaut (commented):
I think that entry in documentation can be also helpful

Yaraslaut (approved):
Great, thanks :)

Issue #1599, opened 2024-09-22
TERM variable in default-config is ignored


herrhotzenplotz:
When I use a default-generated config file the declared `TERM` environment variable is ignored. This manifests in that whenever one launches contour `TERM` is set to `vt100`.

- OS

	FreeBSD hades 14.1-RELEASE-p2 FreeBSD 14.1-RELEASE-p2 releng/14.1-n267685-73905162d0ab GENERIC amd64

- contour version

	0.5.0.7168

I built contour through my FreeBSD Port at [https://github.com/herrhotzenplotz/ports-overlay/tree/master/x11/contour/](https://github.com/herrhotzenplotz/ports-overlay/tree/master/x11/contour/).

Yaraslaut:
What happens when you do `export TERM=contour` inside the shell? 
Also, we are trying to set TERM to contour while starting terminal, so maybe terminfo file is missing, and `exort TERM=contour` will tell you about it

herrhotzenplotz:
> What happens when you do `export TERM=contour` inside the shell?

Of course that works. It's just annoying that everytime I use an ncurses application it starts up in black/white and scrolling in `less(1)` is dead-slow. Then you quit said application, update your environment and start over.


> Also, we are trying to set TERM to contour while starting terminal, so maybe terminfo file is missing

the terminfo is installed correctly:

```console
nico@hades:~ $ find /usr/local/share/terminfo/
/usr/local/share/terminfo/
/usr/local/share/terminfo/c
/usr/local/share/terminfo/c/contour
/usr/local/share/terminfo/c/contour-latest
nico@hades:~ $ 
```


> and `exort TERM=contour` will tell you about it

running `export TERM=...` will tell you nothing. It just sets the shell variable and then exports it to the environment. Unless this is something shell/OS-specific.

Yaraslaut:
Thanks @herrhotzenplotz, will try to figure out what is happening 

When you export TERM, shell checks if you can set it 
```
➜  ~ export TERM=asdfasdf
zsh: can't find terminal definition for asdfasdf
```

herrhotzenplotz:
You're assuming that I am using zsh. However I am using FreeBSD's `/bin/sh` (which itself is an enhanced version of `ash`).

Yaraslaut:
> You're assuming that I am using zsh. However I am using FreeBSD's `/bin/sh` (which itself is an enhanced version of `ash`).

oh, right, i forgot that this is only zsh implementation

[...]

PR #1618, opened 2024-10-03
Add few BSD specific locations during terminfo loading

Yaraslaut:
Closes https://github.com/contour-terminal/contour/issues/1599
@herrhotzenplotz  I hove this is sufficient to find terminfo files on freebsd


christianparpart (approved):
LGTM. I hope that'll help BSD though (I don't have one around right now) :)

Issue #1604, opened 2024-09-30
Exit `ViNormalMode` after yanking and map `Esc` key to exit normal mode


budimanjojo:
Currently `ViNormalMode` feels a little awkward to me. Usually we want to exit normal mode after yanking the selection (at least this is the behavior in `tmux` and `wezterm`).
And another one is (I don't know if separate issue is needed) map `Esc` key to exit vi mode. I know it's not the default behavior in `vi` but in `vi` we usually starts with Normal mode but we start in Insert mode here. And we don't enter normal mode with `Esc` key by default too like in `vi` here. `Esc` feels more natural to me and it's also the default behavior in `tmux` and `wezterm`.

Yaraslaut:
Hi @budimanjojo, thanks for the feedback 
> Usually we want to exit normal mode after yanking the selection (at least this is the behavior in tmux and wezterm).`

I am not sure that this is a good default behaviour, we can customize it,  we already have similar customization for search mode
```
# Whether or not to switch from search mode into insert on exit
search_mode_switch: false
```

> map Esc key to exit vi mode

This is something that is we can add to increase qol of users

budimanjojo:
Yeah I'm good with something similar to `search_mode_switch` option too. Thank you!

Yaraslaut:
Everything is done in two separate PR, so i am closing this

PR #1619, opened 2024-10-03
Add insert_after_yank config option

Yaraslaut:
Refs https://github.com/contour-terminal/contour/issues/1604

### DIFF
--- a/src/vtbackend/Settings.h
+++ b/src/vtbackend/Settings.h
@@ -89,6 +89,7 @@ struct Settings
     PrimaryScreen primaryScreen;
 
     bool fromSearchIntoInsertMode = true;
+    bool isInsertAfterYank = false;
 
     // TODO: we could configure also the number of lines of the host writable statusline and indicator
     // statusline.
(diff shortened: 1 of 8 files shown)


src/vtbackend/Settings.h
```diff
@@ -89,6 +89,7 @@ struct Settings
     PrimaryScreen primaryScreen;
 
     bool fromSearchIntoInsertMode = true;
+    bool inInsertAfterYank = true;
```
christianparpart, line 92: Here it defaults to `true`, but above (in `src/contour/Config.h`), the value is defaulted to `false`. I think it should both be the same (my preference would be `false`, to retain current semantics and standard vi-experience).

Yaraslaut: You are right, I was using it for testing purposes, I still need to connect config and settings

christianparpart:
Should we mention this in the documentation (website) as well?

Yaraslaut:
> Should we mention this in the documentation (website) as well?

Oh, right
We should optimize it away xD

christianparpart (approved):
Thank you

PR #1620, opened 2024-10-04
Exit normal mode with Escape

Yaraslaut:
Refs https://github.com/contour-terminal/contour/issues/1604


src/vtbackend/ViInputHandler.cpp
```diff
@@ -385,7 +386,7 @@ Handled ViInputHandler::sendKeyPressEvent(Key key, Modifiers modifiers, Keyboard
     if (modifiers.any())
         return Handled { true };
 
-    auto const keyMappings = std::array<std::pair<Key, std::string_view>, 10> { {
+    auto const keyMappings = std::vector<std::pair<Key, std::string_view>> { {
```
christianparpart, line 389: Why did you go for vector here? It is a minor, but we could avoid the unnecessary associations when sticking with array.

Yaraslaut: I switched to vector to make it a bit easier to update this list in the future, so you can just add it with one line, without the need to update number of elements, if you think that it is better to keep array here i can undo it

christianparpart (dismissed):
LGTM, thanks :)

PR #1621, opened 2024-10-04
Disable openbsd action

Yaraslaut:
Disable openbsd build 

 @jg1uaa I disable github action to build openbsd since it is not working at the moment because c++ library is to old to support std::format 
 Latest release that does not require std::format is https://github.com/contour-terminal/contour/releases/tag/v0.5.0.7168 but you need to have fmt version lower than 11

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -361,7 +361,7 @@ jobs:
   # }}}
   # {{{ OpenBSD
   openbsd:
-    if: github.ref == 'refs/heads/master' || github.head_ref == 'release'
+    if: false #github.ref == 'refs/heads/master' || github.head_ref == 'release'
     runs-on: ubuntu-latest
     name: OpenBSD 7.5
     # env:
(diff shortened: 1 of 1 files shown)


christianparpart:
Will there be any way to get notified/pinged when OpenBSD has a more up-to-date compiler packaged?

Yaraslaut:
> Will there be any way to get notified/pinged when OpenBSD has a more up-to-date compiler packaged?

I think the best i can do is to subscribe to openbsd mailing :)

jg1uaa:
Sorry for late reply, OpenBSD-7.6 (will be released at this month) and current -current branch has fmt-10.x package.
And, clang-16.0.6 on amd64 platform.

Does this problem comes from termbench-pro?

Yaraslaut:
> Sorry for late reply, OpenBSD-7.6 (will be released at this month) and current -current branch has fmt-10.x package. And, clang-16.0.6 on amd64 platform.
> 
> Does this problem comes from termbench-pro?

No, this is not termbench related, we switched to formatting library that is in the standard of c++ language (https://en.cppreference.com/w/cpp/utility/format) instead of implementation from fmt

PR #1628, opened 2024-10-06
simplify, cleanup, and improve window size/resize handling

christianparpart:
it took me a while to figure out that fraction scaling on KDE Plasma (probably due to Wayland) is problematic. However, this PR still cleans up a lot and reliably produces the expected columns x lines page size on initial window spawn.

Fractional scaling on at least Wayland (KDE?) should be fixed later on its own.

### DIFF
--- a/metainfo.xml
+++ b/metainfo.xml
@@ -110,6 +110,7 @@
           <li>Add terminal tabs (#90)</li>
           <li>Add binding to exit normal mode with `Esc` (#1604)</li>
           <li>Add config option to switch into insert mode after yank (#1604)</li>
+          <li>Improves window size/resize handling on HiDPI monitor settings (#1628)</li>
         </ul>
       </description>
     </release>
--- a/metainfo.xml
+++ b/metainfo.xml
@@ -107,6 +107,9 @@
     <release version="0.5.2" urgency="medium" type="development">
       <description>
         <ul>
+          <li>Add terminal tabs (#90)</li>
+          <li>Add binding to exit normal mode with `Esc` (#1604)</li>
+          <li>Add config option to switch into insert mode after yank (#1604)</li>
         </ul>
       </description>
     </release>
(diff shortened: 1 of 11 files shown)


metainfo.xml
```diff
@@ -107,16 +107,17 @@
     <release version="0.5.2" urgency="medium" type="development">
       <description>
         <ul>
+          <li>Add terminal tabs (#90)</li>
```
Yaraslaut, line 110: oh, my bad xD

src/contour/ContourGuiApp.cpp
```diff
@@ -408,6 +408,11 @@ int ContourGuiApp::terminalGuiAction()
 
     setupQCoreApplication();
 
+    QScreen *screen = QApplication::primaryScreen();
+    assert(screen);
+    qreal scaleFactor = QApplication::primaryScreen()->devicePixelRatio();
+    printf("scaleFactor: %f\n", scaleFactor);
```
Yaraslaut: I suppose this is only for debug

src/contour/TerminalSession.cpp
```diff
@@ -235,20 +235,15 @@ void TerminalSession::detachDisplay(display::TerminalDisplay& display)
...
 
     {
         // NB: Inform connected TTY and local Screen instance about initial cell pixel size.
-        auto const pixels = _display->cellSize() * _terminal.pageSize();
-        // auto const pixels =
-        //     ImageSize { _display->cellSize().width * boxed_cast<Width>(_terminal.pageSize().columns),
-        //                 _display->cellSize().height * boxed_cast<Height>(_terminal.pageSize().lines) };
         auto const l = scoped_lock { _terminal };
-        _terminal.resizeScreen(_terminal.pageSize(), pixels);
+        _terminal.resizeScreen(_terminal.pageSize(), _display->cellSize() * _terminal.pageSize());
```
Yaraslaut, line 246: there is `_display->pixelSize()` can we use it here?

src/contour/display/TerminalDisplay.cpp
```diff
@@ -566,9 +577,9 @@ void TerminalDisplay::onBeforeSynchronize()
 
 void TerminalDisplay::createRenderer()
 {
-    // Require(!_renderTarget);
+    Require(!_renderTarget);
+    // Require(!_renderer);
```
Yaraslaut, line 581: What is the correct check for this condition, before we were checking that renderer exist

Yaraslaut (commented):
Some small comments

src/contour/display/TerminalDisplay.cpp
christianparpart: this commenting-out was a left-over of my failed attempt to fix sizing on fractional-scaling - until I realized that Qt's devicePixelRatio() returns different values, based on time (during startup)

Yaraslaut (approved):
Thanks for fixes :)

PR #1629, opened 2024-10-06
[contour] Change default tabs position on statusbar to the left

christianparpart:
tiny one


Yaraslaut (approved):
Thanks, this is a better default

PR #1630, opened 2024-10-07
Fixes mouse wheel scrolling on High-DPI screens feeling too slow

christianparpart:
funny enough, mouse wheel pixel data depends on High DPI setting of your monitor. Well... May make sense.

### DIFF
--- a/src/contour/helper.cpp
+++ b/src/contour/helper.cpp
@@ -436,8 +436,8 @@ void sendWheelEvent(QWheelEvent* event, TerminalSession& session)
     //       it will send horizontal wheel events instead of vertical ones. We need to compensate
     //       for that here.
 
-    auto const pixelDelta =
-        (modifiers & Modifier::Alt) ? transposed(event->pixelDelta()) : event->pixelDelta();
+    auto const scaledPixelDelta = session.display()->contentScale() * event->pixelDelta();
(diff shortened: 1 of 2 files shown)


src/contour/helper.cpp
```diff
@@ -436,8 +436,10 @@ void sendWheelEvent(QWheelEvent* event, TerminalSession& session)
     //       it will send horizontal wheel events instead of vertical ones. We need to compensate
     //       for that here.
 
+    auto const dpr = session.display()->contentScale();
+
     auto const pixelDelta =
-        (modifiers & Modifier::Alt) ? transposed(event->pixelDelta()) : event->pixelDelta();
+        dpr * ((modifiers & Modifier::Alt) ? transposed(event->pixelDelta()) : event->pixelDelta());
```
Yaraslaut, line 442: Maybe something like this slightly better for readability
```
auto transform = [&](auto delta){ 
return (modifiers & Modifier::Alt) ?  transposed(delta) : delta;
};

auto const pixelDelta = dpr * transform(event->pixelDelta());
```

Yaraslaut (commented):
Small comment concerning readability

src/contour/helper.cpp
christianparpart: Not sure the right hand side is a transform but rather an identity, but I'll refactor :)

christianparpart: @Yaraslaut check again, please

Yaraslaut (approved):
Thanks :)

Issue #1632, opened 2024-10-07
Terminal tabs


Yaraslaut:
Some terminal tab customization and features that would be nice to have

* this issue is much more relevant now https://github.com/contour-terminal/contour/issues/758 and we need to customize this behavior since user might want to open new tab in the same directory 
* For every tab show status of the running process, ideally i would like to see indication whether program that i launched is finished or not, for example I `wget` big file and switched to another tab, i would like to notice when `wget` is finished

PR #1633, opened 2024-10-08
Add styling-capability for active tab in Tabs view in statusline

christianparpart:
Improve UI/UX a bit of tabs feature's statusline:

- have the active tab's color customizable
- teh tabs textual representation in the statusline must not jump around in spacing, based on what tab is active (due to ` [N] ` vs ` N `.

- refs: #1634 (would benefit from it)

### DIFF
--- a/src/contour/Config.h
+++ b/src/contour/Config.h
@@ -351,7 +351,7 @@ struct TerminalProfile
         statusDisplayPosition { vtbackend::StatusDisplayPosition::Bottom };
     ConfigEntry<std::string, documentation::IndicatorStatusLineLeft> indicatorStatusLineLeft {
         " {InputMode:Bold,Color=#FFFF00}"
-        "{Tabs:Left= │ }"
+        "{Tabs:ActiveColor=#FFFF00,Left= │ }"
         "{SearchPrompt:Left= │ }"
         "{TraceMode:Bold,Color=#FFFF00,Left= │ }"
         "{ProtectedMode:Bold,Left= │ }"
--- a/src/vtbackend/StatusLineBuilder.cpp
+++ b/src/vtbackend/StatusLineBuilder.cpp
@@ -42,6 +42,15 @@ namespace // helper functions
     }
 } // namespace
 
+std::optional<RGBColor> tryParseColorAttribute(crispy::string_interpolation const& interpolation,
+                                               std::string_view key)
+{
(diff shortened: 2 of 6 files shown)


Yaraslaut:
Refs https://github.com/contour-terminal/contour/issues/1632

src/vtbackend/StatusLineBuilder.cpp
```diff
@@ -420,9 +434,33 @@ struct VTSerializer
...
     std::string visit(StatusLineDefinitions::VTType const&) { return std::format("{}", vt.terminalId()); }
 
-    std::string visit(StatusLineDefinitions::Tabs const&)
+    std::string visit(StatusLineDefinitions::Tabs const& tabs)
     {
-        return std::format("{}", vt.guiTabInfoForStatusLine());
+        auto const tabsInfo = vt.guiTabsInfoForStatusLine();
+
+        std::string fragment;
+        for (size_t position: std::views::iota(1u, tabsInfo.tabCount + 1))
```
Yaraslaut, line 442: we can use crispy::times here

src/contour/Config.h
```diff
@@ -351,7 +351,7 @@ struct TerminalProfile
         statusDisplayPosition { vtbackend::StatusDisplayPosition::Bottom };
     ConfigEntry<std::string, documentation::IndicatorStatusLineLeft> indicatorStatusLineLeft {
         " {InputMode:Bold,Color=#FFFF00}"
-        "{Tabs:Left= │ }"
+        "{Tabs:ActiveColor=#FFFF00,Left= │ }"
```
Yaraslaut, line 354: We need to change it in the documentation as well, and add ActiveBackground info

christianparpart: I somehow didn t get the config values out of contour.yml, that is why I left it as draft and took a nap instead of finishing it.  I hope to figure out the reason today :)

christianparpart: EDIT: loading config worked all night long. I obviously edited the wrong contour.yml and wondered why it wasn't reflecting my changes :D

src/vtbackend/StatusLineBuilder.cpp
christianparpart: @Yaraslaut Yeah, I tried making more use of standard library. What do you think? `crispy::times` comes from a time where not even range-v3 was used :)

Yaraslaut (approved):
Thanks, great addition :)

Issue #1603, opened 2024-09-29
Long character and underscore is cut off when font size is set to 12


budimanjojo:
### Contour Terminal version

0.4.3

### Installer source

something else (please specify below)

### Operating System

Ubuntu Linux with Nix

### Architecture

x86-64

### Steps to reproduce

Have `profiles.main.font.size` set to `12` with `UbuntuMono Nerd Font`.

### Expected Behavior

Should display fine

### Actual Behavior

The font is cut off, characters like `g`, `q`, `p`, and `_` are displayed incorrectly. Look at the screenshot below.
![image](https://github.com/user-attachments/assets/c093f1f6-4ecc-4eb7-9a8c-2138867cd604)

### Additional notes

Setting the font to `12.1` fixed this issue. I believe this should also only happen with some fonts (mine being UbuntuMono Nerd Font). This issue doesn't happen with other terminal emulators (tested with Alacritty, Wezterm, Gnome-terminal).

Yaraslaut:
Hi, few font related bugs where fixed in 0.5 release, can you please check this bug in the latest release?

budimanjojo:
Okay I'll try to build the latest version with Nix and report back. Thank you!

budimanjojo:
Apologize for the delay. I couldn't build `contour` at all on my 16GB RAM machine, it keeps killing my entire system on 79% with 100% memory usage. So I decided to run the `contour-0.5.0.7168-fix.AppImage` from the release assets. The bug still persists though. The font is still cut off with `12` font size.

Yaraslaut:
> Apologize for the delay. I couldn't build `contour` at all on my 16GB RAM machine, it keeps killing my entire system on 79% with 100% memory usage. So I decided to run the `contour-0.5.0.7168-fix.AppImage` from the release assets. The bug still persists though. The font is still cut off with `12` font size.

I think at some point gcc was broken, somewhere around gcc 13, with gcc 14 everything is alright, and clang is usually more reliable as well, thanks for trying out, i will try to reproduce it on my system

jg1uaa:
Hello, I found this issue with [VLGothic](https://github.com/daisukesuzuki/VLGothic/releases/tag/20230918)(family: vlgothic)
![vl](https://github.com/user-attachments/assets/84e31845-af4d-4029-a6a0-b824538c7df4)
[...]

PR #1639, opened 2024-10-20
Do not crop symbol that is completely below cell

Yaraslaut:
Closes https://github.com/contour-terminal/contour/issues/1603

Some fonts want underscore character to be below grid cell completely, we can allow


christianparpart (approved):
Oh thank you

Issue #1198, opened 2023-09-11
Protect user from accidentally pasting too large input


christianparpart:
@Marenz just found out, that when pasting too big text data to Contour, it will freeze (or at least looks like this).

After a short investigation, we found out that the data to be pasted was still being pasted, it just takes very long. Most likely because the way we do it is certainly not most efficient.

## Action items

- [ ] Ask the user something like: "Are you sure you want to paste {X amount} characters?" in a Yes/No dialog and only continue pasting on Yes, abort otherwise.
- [ ] If the content is really to be pasted, make sure it's more efficiently done and NOT blocking the GUI thread

Yaraslaut:
I noticed that `BracketedPaste` decmode restrict pasting more than one screen of text and is enabled by default, so if i for example start with 
```
echo "
> 
```
I can not paste large text. I am not sure what to do with this, maybe @christianparpart  or @j4james you can help out with what is the right solution for this one

PR #1643, opened 2024-10-23
Request permission on big paste

Yaraslaut:
Closes https://github.com/contour-terminal/contour/issues/1198

### DIFF
--- a/src/contour/TerminalSession.cpp
+++ b/src/contour/TerminalSession.cpp
@@ -368,6 +368,7 @@ void TerminalSession::executeRole(GuardedRole role, bool allow, bool remember)
         case GuardedRole::ShowHostWritableStatusLine:
             executeShowHostWritableStatusLine(allow, remember);
             break;
+        case GuardedRole::BigPaste: applyPendingPaste(allow, remember); break;
     }
 }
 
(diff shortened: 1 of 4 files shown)


src/contour/TerminalSession.cpp
```diff
@@ -651,23 +653,65 @@ void TerminalSession::pasteFromClipboard(unsigned count, bool strip)
         sessionLog()("pasteFromClipboard: mime data contains {} formats.", md->formats().size());
         for (int i = 0; i < md->formats().size(); ++i)
             sessionLog()("pasteFromClipboard[{}]: {}\n", i, md->formats().at(i).toStdString());
-        string const text = strip_if(normalize_crlf(clipboard->text(QClipboard::Clipboard)), strip);
-        if (text.empty())
+
+        auto const text = clipboard->text(QClipboard::Clipboard);
+
+        // 1 MB hard limit
+        if (text.size() > 1024 * 1024)
```
christianparpart, line 660: I think this is definitely a good self-protection to have. Pasting that much into `stdin` most likely smells like either abuse, or somebody not knowing that this will freeze your session. We can see if we need to tweak the number in the future.

christianparpart (approved):
Thank you very much :)

Issue #1622, opened 2024-10-05
Generate configuration documentation


Yaraslaut:
It is often quite easy to forget about documentation entry on the web-page when adding new entry to the configuration. 
Some parts of the web-page are generated using contour executable already, for example info about supported VT sequences (https://contour-terminal.org/vt-sequence/) and key mapping section (https://contour-terminal.org/configuration/key-mapping/)
It would be great to provide automatic generation of config entries for the web-page

Automatic generation:

- [x] global config
- [x] profile config
- [x] key mapping
- we can ignore the color scheme because changes are rare

PR #1644, opened 2024-10-26
Transform camelCase into snake_case for docs

Yaraslaut:
Transform camelCase of names  inside c++ structures to snake_case that is used inside config files 
Refs https://github.com/contour-terminal/contour/issues/1622


christianparpart (approved):
Very nice. Thanks.

PR #1645, opened 2024-10-26
Position middle segment of status line properly

Yaraslaut:
Fixes positioning of middle segment in the status line

### DIFF
--- a/src/vtbackend/Terminal.cpp
+++ b/src/vtbackend/Terminal.cpp
@@ -589,10 +589,12 @@ void Terminal::updateIndicatorStatusLine()
         // Don't show the middle segment if text is too long.
         auto const middleLength = serializeToVT(*this, definitions.middle, Styling::Disabled).size();
         auto const center = pageSize().columns / ColumnCount(2) - ColumnCount(1);
+        // size of middleVT segment is less that number of elements in the string due to formatting
+        // and on average we can assume that the middle segment is half the size of the string
         _indicatorStatusScreen.moveCursorToColumn(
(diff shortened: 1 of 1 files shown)


src/vtbackend/Terminal.cpp
```diff
@@ -589,10 +589,11 @@ void Terminal::updateIndicatorStatusLine()
         // Don't show the middle segment if text is too long.
         auto const middleLength = serializeToVT(*this, definitions.middle, Styling::Disabled).size();
         auto const center = pageSize().columns / ColumnCount(2) - ColumnCount(1);
+        // size of middleVT string is double the actual size
         _indicatorStatusScreen.moveCursorToColumn(
-            ColumnOffset::cast_from(center - ColumnOffset::cast_from(middleLength / 2)));
+            ColumnOffset::cast_from(center - ColumnOffset::cast_from(middleLength / 4)));
```
christianparpart, line 594: for me to understand, what am I missing here that we div by 4 instead 2? I'd probably do the same bug again, if I'd had to, because I don't get it :)

christianparpart (approved):
I'm puzzled. But I trust you. :)

src/vtbackend/Terminal.cpp
Yaraslaut: Length of this segment is twice the actual size after rendering (i suppose because of the bold vt sequence of similar), so when position cursor to render middle segment in the middle we have to take this into account, number of cells that we need is `middleLength/2` thus initial offset is `center-middleLength/4`

Issue #1101, opened 2023-05-23
vi-like normal mode: support Jumps


christianparpart:
when pressing `%` to jump to a matching symbol and land somewhere completely unexpected (or do any other kind of long distance movement), you may want to quickly jump back.

For that, adding support for jumps is more useful than hard to implement.

- [ ] `''` ``` `` ``` (To the position before the latest jump, or where the last "m'" or "m`" command was given.)
- [ ] `<Ctrl>+O`
- [ ] `<Ctrl>+I`

### To quote vim doc

```
A "jump" is a command that normally moves the cursor several lines away.  If
you make the cursor "jump" the position of the cursor before the jump is
remembered.  You can return to that position with the "''" and "``" commands,
unless the line containing that position was changed or deleted.  The
following commands are "jump" commands: "'", "`", "G", "/", "?", "n", "N",
"%", "(", ")", "[[", "]]", "{", "}", ":s", ":tag", "L", "M", "H" and the
commands that start editing a new file.
```

```
Jumps are remembered in a jump list.  With the CTRL-O and CTRL-I command you
can go to cursor positions before older jumps, and back again.  Thus you can
move up and down the list.  There is a separate jump list for each window.
The maximum number of entries is fixed at 100.
```

There's more interesting stuff that might be worth adding, see `:help mark-motions` in vim.

### Applying this to a terminal

I think most important is to add some sort of jump list (maybe start capping it to 100, just like in vim) and then make sure that certain jumps are recorded into that jump list.

An item in a jump list would probably be simply a `CellLocation`. This means, line values are relative. negative points into the history. This also means, that when adding new lines, each recorded line location would needed to be adjusted accordingly. I wonder if that one could be optimized (or if that doesn't matter at all, perf wise).

The jump list should probably be remembered per screen buffer, because each screen buffer might most likely have different history line counts. Most likely, jump list only make sense when having a history at all.

PR #1646, opened 2024-10-26
Implement vi mode jumps

Yaraslaut:
Closes https://github.com/contour-terminal/contour/issues/1101

### DIFF
--- a/metainfo.xml
+++ b/metainfo.xml
@@ -113,6 +113,7 @@
           <li>Protect user from accidentally pasting too large input (#1198)</li>
           <li>Add binding to exit normal mode with `Esc` (#1604)</li>
           <li>Add config option to switch into insert mode after yank (#1604)</li>
+          <li>Add vi-like normal mode Jumps. `''`, `C-O` and `C-I` motions (#1101)</li>
           <li>Improves window size/resize handling on HiDPI monitor settings (#1628)</li>
           <li>Improves macOS key handling for Option+Left|Right keys to jump between words in the shell</li>
           <li>Fixes cropping of underscore character for some fonts (#1603)</li>
--- a/src/vtbackend/JumpHistory.cpp
+++ b/src/vtbackend/JumpHistory.cpp
@@ -1,0 +1,56 @@
+// SPDX-License-Identifier: Apache-2.0
+#include <vtbackend/JumpHistory.h>
+
+namespace vtbackend
+{
+
+CellLocation JumpHistory::jumpToLast(CellLocation current)
+{
+    applyOffset();
+    CellLocation last = _history.back();
+    if (last == current)
+    {
+        _history.pop_back();
+        if (_history.empty())
+        {
+            return current;
+        }
+        last = _history.back();
+    }
+    _history.pop_back();
+    _history.push_back(current);
+    _current = _history.size();
+    return last;
+}
+
+CellLocation JumpHistory::jumpToMarkBackward([[maybe_unused]] CellLocation current)
+{
+    applyOffset();
+    if (_current == 0)
+    {
+        // loop
+        _current = _history.size() - 1;
+    }
+    else
+    {
+        _current--;
+    }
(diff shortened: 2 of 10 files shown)


src/vtbackend/ViInputHandler.h
```diff
@@ -100,6 +100,9 @@ enum class ViMotion : uint8_t
     ToCharLeft,            // F {char}
     RepeatCharMove,        // ;
     RepeatCharMoveReverse, // ,
+    JumpToLastJumpPoint,   // '' or `` (jump to last jump)
+    JumpToMarkUp,          // <C-O>
+    JumpToMarkDown,        // <C-I>
```
christianparpart, line 105: I think that Up/Down is really misleading here. It should be either Out/In or Backward/Forward, which makes more sense in my opinion :)

metainfo.xml
```diff
@@ -113,6 +113,7 @@
           <li>Protect user from accidentally pasting too large input (#1198)</li>
           <li>Add binding to exit normal mode with `Esc` (#1604)</li>
           <li>Add config option to switch into insert mode after yank (#1604)</li>
+          <li>Add vi-like normal mode Jumps (#1101)</li>
```
christianparpart, line 116: we should probably mention what that means...

 `('', <C-O>, <C-I>`)

src/vtbackend/JumpHistory.cpp
```diff
@@ -0,0 +1,55 @@
+#include <vtbackend/JumpHistory.h>
```
christianparpart, line 1: license header missing

src/vtbackend/JumpHistory.h
```diff
@@ -0,0 +1,40 @@
+#pragma once
```
christianparpart, line 1: license header missing

src/vtbackend/JumpHistory.h
```diff
@@ -0,0 +1,40 @@
...
+
+#include <vtbackend/primitives.h>
+
+#include <vector>
+
+namespace vtbackend
+{
+
+struct JumpHistory
```
christianparpart, line 10: if we have private data/methods, it should be a `class` instead (ref: CppCoreGuidelines IIRC)

src/vtbackend/JumpHistory.h
```diff
@@ -0,0 +1,40 @@
...
+{
+
+    template <typename T>
+    void add(T&& cell)
+    {
+        applyOffset();
+        _history.push_back(std::forward<decltype(cell)>(cell));
```
christianparpart, line 17:
```suggestion
        _history.push_back(std::forward<T>(cell));
```

src/vtbackend/ViCommands.cpp
```diff
@@ -773,8 +782,13 @@ CellLocation ViCommands::globalCharDown(CellLocation location, char ch, unsigned
     return result;
 }
 
-CellLocation ViCommands::translateToCellLocation(ViMotion motion, unsigned count) const noexcept
+CellLocation ViCommands::translateToCellLocation(ViMotion motion, unsigned count) noexcept
```
christianparpart, line 785: requiring both functions to be mutable sounds like a bad design choice. they were meant to *only* translate. But I do think I know why you where doing it, as these functions translate AND record cell location.

So either we factor out the recording of the current cell location, or (IMHO), I'd suggest to rename those function to something e.g. like `translateToCellLocationAndRecord`. What do you think?

christianparpart (commented):
Thank you very much, @Yaraslaut. I've tested it locally, and it's amazing. I only left very few comments though :)

src/vtbackend/ViCommands.cpp
Yaraslaut: renamed to `translateToCellLocationAndRecord`

christianparpart (approved):

Issue #1649, opened 2024-11-06
Releases do not say minimum version of macOS required


peterjc:
I was curious to try contour on my main machine which is currently an old Intel Mac, which Apple no longer supports and is therefore stuck as macOS 12.7.6 "Monterey". Looking at https://github.com/contour-terminal/contour/releases it was not clear what minimum version of macOS was expected, if any. I found out by downloading and trying:

* Contour 0.5.1.7247 Darwin - requires macOS 14.0 or later
* Contour 0.4.3.6442 Darwin - requires macOS 13.0 or later
* Contour 0.4.0.6245 Darwin - requires macOS 13.0 or later
* Contour 0.3.12.262 Darwin - willing to run on macOS 12.7.6

It would be nice to add a line to the first releases where this changed, which looks like v0.4.x requiring macOS 13.0 "Ventura" or later, and v0.5.x requiring macOS 14.0 "Sonoma" or later.

Also, https://github.com/contour-terminal/contour/blob/master/README.md still says under requirements:

> **operating system**: A recent operating system (macOS 12, ...)

That should probably now say macOS 14?

PR #1650, opened 2024-11-06
Since v0.5 requires macOS 14

peterjc:
This is based on trying the recent releases, see #1649

## Description

Update minimum version of macOS in the README file requirements section.

## Motivation and Context

Avoids false hope in users with older Apple Mac that they can run the latest release.

## How Has This Been Tested?

Tested recent releases under macOS 12.7.6 "Monterey" to determine the minimum version of macOS reported as required, see #1649.

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ Click the following button to install Contour from the Flathub store.
 
 ## Requirements
 
-- **operating system**: A *recent* operating system (macOS 12, Windows 10+, an up-to-date Linux, FreeBSD or OpenBSD)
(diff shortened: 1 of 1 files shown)


whisperity:
I don't think we should say version numbers there because it will be a constant chase of updates when new OSes come out. Who sets the minimum OS version anyway? Even the change to not having Intel binaries happened because GitHub changed their CI, it was not directly a conscious decision by this project I think.

Yaraslaut:
I close this PR since readme will be updated in https://github.com/contour-terminal/contour/pull/1653

PR #1656, opened 2024-11-22
Fix typo in "operating system"

JakeWharton:
Although sometimes our OSes can truly can be an "operaging" system, I suspect "operating" was intended.


uspasojevic96:
Thanks, it was indeed a typo 🫣

PR #1662, opened 2024-12-07
Drop arch action and update readme

Yaraslaut:
Drop arch linux action  and update README for installation


christianparpart (approved):
LGTM (retriggered failed Fedora build)

PR #1682, opened 2024-12-28
Fix error with wrong terminal size during tab switch

Yaraslaut:
Fix crash with resize and tab switching, we notice this bug on systems with scaled UI (specifically macOS, but bug can appear on other systems as well)


christianparpart (dismissed):
I would like you to have a look on my second commit on top to, but I'd think it now looks all "okay" (bug is fixed) :)

Yaraslaut:
> I would like you to have a look on my second commit on top to, but I'd think it now looks all "okay" (bug is fixed) :)

looks good for me :)

Issue #1685, opened 2024-12-30
Shift+Tab no longer works


T-727:
### Contour Terminal version

0.5.1.7247

### Installer source

Github: release page

### Operating System

Windows 10.0.26100

### Architecture

x86-64

### Other Software

Tried opening `cmd`, `pwsh`, and `nvim` none of them seem to receive Shift+Tab correctly anymore

### Steps to reproduce

press Shift+Tab within any application which has Shift+Tab bindings

### Expected Behavior

Shift+Tab working like previous versions

### Actual Behavior

contour itself seems to receive Shift+Tab fine but the shell receives a regular Tab

### Additional notes

- no custom config file was used
- can only reproduce using `0.5.1.7247` due to https://github.com/contour-terminal/contour/issues/1684
- likely an effect of https://github.com/contour-terminal/contour/pull/1579

relevant `contour terminal debug *`:
previously:
```shell
[2024-12-30 08:55:21.8839874.883987] [gui.input] Character Press event received: Shift '\t'
[2024-12-30 08:55:21.8845888.884588] [vt.input] Sending Shift "\t" Press.
[2024-12-30 08:55:21.8849575.884957] [pty.output] Sending bytes: "\e[Z"
```
now:
```shell
[2024-12-30 11:20:19.550819] [gui.input] Key Press event received: Shift Tab
[2024-12-30 11:20:19.551802] [vt.input] Sending Shift "Tab" Press.
[2024-12-30 11:20:19.552115] [pty.output] Sending bytes: "\t"
```

Yaraslaut:
Hm, so it looks like we do not apply modifier correctly, since for tab we need to send "\t" but for Shift tab "\e[Z"

Issue #1578, opened 2024-07-29
Key mapping Ctrl+Shift+Tab not working.


csy19960309:
### Contour Terminal version

0.4.3

### Installer source

something else (please specify below)

### Operating System

Archlinux with kernel 6.9.10-zen1-1-zen

### Architecture

x86-64

### Other Software

tmux 3.4-9, wayland 1.23.0-1

### Steps to reproduce

Add _{ mods: [Control, Shift], key: Tab,           action: SendChars, chars: "test" }_ to config.

### Expected Behavior

It prints "test" to stdout.

### Actual Behavior

An actual _tab_ is printed. I mean, it's just like that I press the key TAB.

Yaraslaut:
Hi @csy19960309, thanks for the bug report

PR #1686, opened 2024-12-30
Fix backtab pty.output generation

Yaraslaut:
Closes https://github.com/contour-terminal/contour/issues/1685 
Solution is somewhat ugly, I am open for better one @christianparpart  :) 
but we need to preserve behavior for sending it as a key for https://github.com/contour-terminal/contour/issues/1578

### DIFF
--- a/src/vtbackend/InputGenerator.cpp
+++ b/src/vtbackend/InputGenerator.cpp
@@ -68,6 +68,7 @@ bool StandardKeyboardInputGenerator::generateChar(char32_t characterEvent,
         return true;
     }
 
+    // Backtab handling, 0x09 is Tab
     if (modifiers == Modifier::Shift && characterEvent == 0x09)
     {
         append("\033[Z"); // introduced by linux_console in 1995, adopted by xterm in 2002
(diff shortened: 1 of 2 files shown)


src/vtbackend/InputGenerator.cpp
```diff
@@ -70,6 +70,7 @@ bool StandardKeyboardInputGenerator::generateChar(char32_t characterEvent,
 
     if (modifiers == Modifier::Shift && characterEvent == 0x09)
     {
+        // Shift+Tab (backtab) should hit here
```
christianparpart, line 73: Not just "should". 0x09 is well known ASCII code for TAB. Which makes this comment kind of pointless, don't you think so? :)

christianparpart (dismissed):
LGTM, but I think the comment above isn't really adding anything :)

src/vtbackend/InputGenerator.cpp
Yaraslaut: Changed a bit, i do not recognize ASCII codes :)

PR #1689, opened 2024-12-31
[contour] Improve tab switch handling and add `SwitchToPreviousTab`

christianparpart:
Improves usability of tab switching and tab closing a bit.

Also adds new action `SwitchToPreviousTab` to bind some input mapping to, to easily switch between two tabs that are not necessarily next to each other.

I also refactored the code around it for better clarity.

### DIFF
--- a/src/contour/TerminalSessionManager.cpp
+++ b/src/contour/TerminalSessionManager.cpp
@@ -14,7 +14,6 @@
 
 #include <algorithm>
 #include <filesystem>
-#include <mutex>
 #include <string>
 
 using namespace std::string_literals;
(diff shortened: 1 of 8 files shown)


src/contour/TerminalSessionManager.cpp
```diff
@@ -84,49 +89,74 @@ TerminalSession* TerminalSessionManager::createSession()
...
+    _lastTabChange = std::chrono::steady_clock::now();
+    updateStatusLine();
+
+    if (display)
     {
-        _activeSession = _sessions[index];
+        managerLog()("Attaching display to session.");
+        auto const pixels = display->pixelSize();
+        auto const totalPageSize =
+            display->calculatePageSize() + _activeSession->terminal().statusLineHeight();
```
Yaraslaut, line 130:
```suggestion
            display->calculatePageSize() + _previousActiveSession->terminal().statusLineHeight();
```
to preserve the behavior

src/contour/TerminalSessionManager.cpp
```diff
@@ -84,49 +89,74 @@ TerminalSession* TerminalSessionManager::createSession()
...
 
 void TerminalSessionManager::addSession()
 {
-    setSession(_sessions.size());
+    activateSession(createSessionInBackground());
+}
+
+void TerminalSessionManager::switchToPreviousTab()
+{
+    managerLog()("SWITCH TO LAST TAB (current: {}, last: {})",
```
Yaraslaut, line 152:
```suggestion
    managerLog()("Switch to previous tab (current: {}, previous: {})",
```

Yaraslaut (changes requested):
Great addition, some minor things

src/contour/TerminalSessionManager.cpp
christianparpart: Not sure why this one should be correct just yet :thinking:

PR #1695, opened 2025-01-01
Further improvements to tabs handling (add MoveTabToLeft / MoveTabToRight / MoveTabTo actions)

### LINKED ISSUE #1691: Improved tab management: move tabs & tab positioning at creation
Allow moving tabs around by adding some new actions:

- [x] `MoveTabToLeft`
- [x] `MoveTabToRight`
- [x] `MoveTabTo { int position; }`
- [x] Creating a new tab should always position it right next to the current tab (feels more similar to other tabbed applications)

### LINKED ISSUE #1692: Improved tab management:  Prevent spam-spawning new tabs
Assuming the `CreateTab` action is set to keyboard mapping `Ctrl`+`T`, when the user presses this combination and does not release the buttons (maybe on purpose, maybe by accident not quick enough), then we currently keep spawning new sessions on every key repeat interval, which might be a lot, depending on the user setting (e.g. every 50ms).

While spamming input bindings should generally be possible, this is the one action where we should guard against it (this would also match how KDE's Konsole is doing it).

christianparpart:
initial support of moving tabs

@Yaraslaut I think we should give `TerminalSession` access to its owning list of tabs (`TerminalSessionManager` currently).

Maybe the name isn't the best, when thinking of tabs alone. In the long term, we might think of a tree data structure in order to implement tiling-like splits/pane views.

So whatever the DS / API will be, a leaf node (like a terminal session) should always have access to its parent node. What do you think? IMHO, this could also simplify the logic in the QML side (something I failed to revive in my head today)

### Update

Changes:

- [x] add MoveTabToLeft / MoveTabToRight, MoveTabTo actions
- [x] ensure inserting new tab happens right next to the currently active tab
- [x] fixes `CreateTab` to sometimes not spawn more than one tab (fixed implicitly)

closes #1692
closes #1691

### DIFF
--- a/metainfo.xml
+++ b/metainfo.xml
@@ -107,7 +107,9 @@
     <release version="0.6.2" urgency="medium" type="development">
       <description>
         <ul>
+          <li>Fixes `CreateTab` to sometimes spawn more than one tab (#1695)</li>
           <li>Adds `MoveTabToLeft` and `MoveTabToRight` actions to move tabs around (#1695)</li>
+          <li>Ensure inserting new tabs happens right next to the currently active tab (#1695)</li>
         </ul>
       </description>
     </release>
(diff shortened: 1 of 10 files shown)


Yaraslaut (approved):
Nice, great changes
We can switch to internal handling of actions in another PR

PR #1696, opened 2025-01-02
add missing separators

izabera:
fixes broken replies to decrqss (and possibly more, but i only noticed while testing that)

tested by calling decrqss


christianparpart (approved):
Thank you very much for your contribution, @izabera.

p.s.: It's only affecting the case you found by accident. I'm glad you could find yourself home in the source code :)

PR #1697, opened 2025-01-02
Change `colors` terminfo entry from 256 up to 32767 colors

christianparpart:
With a little bit of help, I've got `msgcat --color=test` to output true colors in the "hue/saturation" section, which looks partly like this:

image

While this one better exposes the colors supported by Contour, we still have a remaining problem. gettext (the package providing `msgcat` executable) is heavily hardcoded down to xterm :vomiting_face:), so we need to make it pretend we are exterm by symlinking `c/contour` terminfo file to `x/xterm-contour`.

We should probably talk to the gettext guys to make their heuristic less xterm-centric.

### DIFF
--- a/src/vtbackend/Capabilities_test.cpp
+++ b/src/vtbackend/Capabilities_test.cpp
@@ -27,7 +27,7 @@ TEST_CASE("Capabilities.get")
     REQUIRE(rgb == "8/8/8");
 
     auto const colors = tcap.numericCapability("colors");
-    REQUIRE(colors == 256);
+    REQUIRE(colors == std::numeric_limits<int16_t>::max());
 
     auto const bce = tcap.numericCapability("bce");
     REQUIRE(bce);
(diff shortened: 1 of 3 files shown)


src/vtbackend/Capabilities_test.cpp
```diff
@@ -27,7 +27,7 @@ TEST_CASE("Capabilities.get")
     REQUIRE(rgb == "8/8/8");
 
     auto const colors = tcap.numericCapability("colors");
-    REQUIRE(colors == 256);
+    REQUIRE(colors == 0x7FFF);
```
Yaraslaut, line 30: can we use here numerical limit, i am very bad at hex numbers :)

christianparpart: done

Yaraslaut (approved):
Thanks :)

PR #1698, opened 2025-01-02
Switch to TerminalSessionManager everywhere and fix small bug

Yaraslaut:
1) Avoid qml signal handling 

2) Handle action per key press without repeating, so you can not spawn new tabs when holding key-binding for it

### DIFF
--- a/src/contour/TerminalSession.cpp
+++ b/src/contour/TerminalSession.cpp
@@ -833,7 +833,8 @@ void TerminalSession::sendKeyEvent(Key key, Modifiers modifiers, KeyboardEventTy
         if (auto const* actions =
                 config::apply(_config.inputMappings.value().keyMappings, key, modifiers, matchModeFlags()))
         {
-            executeAllActions(*actions);
+            if (eventType == KeyboardEventType::Press)
+                executeAllActions(*actions);
             return;
         }
     }
@@ -868,7 +869,8 @@ void TerminalSession::sendCharEvent(
                 config::apply(_config.inputMappings.value().charMappings, value, modifiers, matchModeFlags());
             actions && !_terminal.inputHandler().isEditingSearch())
         {
-            executeAllActions(*actions);
+            if (eventType == KeyboardEventType::Press)
+                executeAllActions(*actions);
             return;
         }
     }
@@ -1440,7 +1442,7 @@ bool TerminalSession::operator()(actions::CreateNewTab)
 
 bool TerminalSession::operator()(actions::CloseTab)
 {
-    emit closeTab();
+    _manager->closeTab();
     return true;
 }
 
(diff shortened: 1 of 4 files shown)


src/contour/TerminalSession.cpp
```diff
@@ -833,7 +833,8 @@ void TerminalSession::sendKeyEvent(Key key, Modifiers modifiers, KeyboardEventTy
         if (auto const* actions =
                 config::apply(_config.inputMappings.value().keyMappings, key, modifiers, matchModeFlags()))
         {
-            executeAllActions(*actions);
+            if (eventType != KeyboardEventType::Repeat)
```
christianparpart, line 836: instead of adding the `if` here, change the already existing `if`-condition in line 831 to `== KeyboardEventType::Press`.

src/contour/TerminalSession.cpp
```diff
@@ -868,7 +869,9 @@ void TerminalSession::sendCharEvent(
                 config::apply(_config.inputMappings.value().charMappings, value, modifiers, matchModeFlags());
             actions && !_terminal.inputHandler().isEditingSearch())
         {
-            executeAllActions(*actions);
+            if (eventType != KeyboardEventType::Repeat)
```
christianparpart, line 872: same

src/contour/TerminalSession.cpp
Yaraslaut: Updated

Yaraslaut: Now it will make a bell sound while holding key, but i suppose this is ok

christianparpart (approved):
nice. thanks

src/contour/TerminalSession.cpp
christianparpart: > Now it will make a bell sound while holding key, but i suppose this is ok

I think I am confused. :thinking:

Yaraslaut: Previously, during `Repeat` event type, input was ignored and nothing was happening, after suggested changes we ignore this branch for `Repeat` events and send them further into the terminal, and since vt backend does not know what to do with them it notifies user that something is happening via bell sound

christianparpart: you're right. Repeat should only be sent if a Press was sent out first. So I think the solution is a little bit more complex:

If an input binding matches, it'll be considered "handled". But only Press is going to be sent.
Anything else is forwarded as usual.

But maybe we could leave the logic as-is and only check for Press event (vs Repeat) inside NewTab action. Because most actions you probably want to be spammable. (I at least checked with konsole TE, they only guard against spamming for spawning a new tab)

Yaraslaut: I update it and now we trigger action only on press events, also i think we can keep all actions triggered only on press events

christianparpart (approved):
thank you :)

Issue #1701, opened 2025-01-05
Control-6 inputs just 6


sakhnik:
### Contour Terminal version

0.5.1.7247

### Installer source

something else (please specify below)

### Operating System

Arch Linux

### Architecture

x86-64

### Other Software

NVIM v0.11.0-dev-1465+ga09c7a5d5

### Steps to reproduce

Take the script `keys.sh`:
```bash
#!/bin/bash
echo "Press Ctrl-^ and then Enter. Press Ctrl-C to exit."
while true; do
    read -rsn1 input
    if [ "$input" = "" ]; then
        echo "Enter pressed"
        continue
    fi
    printf "You pressed: %s (hex: %02X)\n" "$input" "'$input"
done
```

When ctrl-6 is pressed, just 6 (0x36) is captured. This happens in Neovim, too: I intend to switch the keymap with ctrl-^, but the character `6` is inserted instead.

### Expected Behavior

Keyboard layout switch in Neovim

### Actual Behavior

6 is inserted

PR #1702, opened 2025-01-05
Add control codes for Control+(5,6,7,8) input

Yaraslaut:
Closes https://github.com/contour-terminal/contour/issues/1701


j4james:
The Ctrl key combinations are all listed in DEC STD-070 in section 6.16 (Control Codes and Keystrokes) on page 6-170. It's also worth reading the earlier section 6.13.1 (Control Key Operations) starting on page 6-75, because that clarifies their intended use with international keyboards. I think the point of these keys is that they guarantee a way to generate control codes on keyboard layouts which wouldn't necessarily have all the punctuation characters.

christianparpart:
Ah, many thanks @j4james. :)

Issue #1710, opened 2025-01-20
Neovim escape key no longer working after insert mode in terminal.


kierun:
### Contour Terminal version

Contour Terminal Emulator 0.6.1

### Installer source

something else (please specify below)

### Operating System

Fedora 40

### Architecture

x86-64

### Other Software

```
√ ; bob list
┌────────────────────┬─────────────┐
│  Version           │  Status     │
├────────────────────┼─────────────┤
│  nightly-9e7b0bc   │  Installed  │
│  nightly-4889935   │  Installed  │
│  nightly-6719276   │  Installed  │
│  nightly           │  Installed  │
│  v0.10.3           │  Used       │
└────────────────────┴─────────────┘
```

All these versions were checked.

### Steps to reproduce

1. Check terminal: `echo $TERM`. I have tried with `xterm`, `vt100`, `screen`, `screen-256color`, and `contour`.
2. Open a file with `nvim FILE`.
3. Get into insert mode and type some text.
4. Try to get into command mode via the escape key: nothing happens.

![Image](https://github.com/user-attachments/assets/faae12aa-3a6a-46ee-b438-d723f8fb6a89)

### Expected Behavior

The escape key should work, as it 

### Actual Behavior

The escape key does not work… 

### Additional notes

- Note that `xterm-ghostty` as `TERM` works fine using ghostty. 
- Alacritty using `screen-256color` works too. 

Therefore, I think it might be a bug with contour.

whisperity:
Is it only an nvim problem, or does the problem arise with other tools? If you just start `cat` or `hexdump` normally in the shell and type things in and press ESC meanwhile, do you get a `^[` or in hex mode the <kbd>0x1b</kbd> escape code?

```
❯ hexyl
0123456789abcde^[
┌────────┬─────────────────────────┬─────────────────────────┬────────┬────────┐
│00000000│ 30 31 32 33 34 35 36 37 ┊ 38 39 61 62 63 64 65 1b │01234567┊89abcde•│
```

kierun:
> Is it only an nvim problem, or does the problem arise with other tools? If you just start `cat` or `hexdump` normally in the shell and type things in and press ESC meanwhile, do you get a `^[` or in hex mode the 0x1b escape code?

@whisperity 

This is what I get. After the `0`, I did try using the escape key several times and nothing happened. 

```bash
√ ; hexyl
1234567890
^C
```

I tried `xev` which got me this:

```bash
[...]

PR #1714, opened 2025-01-26
Fix default binding for cancel selection with escape

Yaraslaut:
Closes https://github.com/contour-terminal/contour/issues/1710


christianparpart (approved):
Looks good. Many thanks.

Issue #1707, opened 2025-01-14
Crash using Chinese IME


adre5577:
### Contour Terminal version

Contour Terminal Emulator 0.5.1

### Installer source

something else (please specify below)

### Operating System

Manjaro/Arch Linux 6.12.4-1-MANJARO

### Architecture

x86-64

### Other Software

IBus 1.5.31 with any Chinese input method

### Steps to reproduce

Type more than one character with your IME of choice.

### Expected Behavior

The characters you type appear in the terminal as normal

### Actual Behavior

The terminal crashes with the following error: src/contour/helper.cpp:411: bool contour::sendKeyEvent(QKeyEvent*, vtbackend::KeyboardEventType, TerminalSession&): Assertion `codepoints.size() == 1' failed.

PR #1715, opened 2025-01-26
Relax condition for codepoints size for IME

Yaraslaut:
Closes https://github.com/contour-terminal/contour/issues/1707
We need to relax the condition on the number of codepoints that can be inserted at once, since IME allows inserting more than one  glyph at a time.

### DIFF
--- a/src/contour/helper.cpp
+++ b/src/contour/helper.cpp
@@ -407,7 +407,7 @@ bool sendKeyEvent(QKeyEvent* event, vtbackend::KeyboardEventType eventType, Term
     if (!event->text().isEmpty())
     {
         auto const codepoints = event->text().toUcs4();
-        assert(codepoints.size() == 1);
+        assert(codepoints.size() >= 1);
 #if defined(__APPLE__)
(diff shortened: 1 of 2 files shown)


christianparpart (approved):
Oh wow. I've never thought about it. Thanks.

src/contour/helper.cpp
```diff
@@ -407,7 +407,7 @@ bool sendKeyEvent(QKeyEvent* event, vtbackend::KeyboardEventType eventType, Term
     if (!event->text().isEmpty())
     {
         auto const codepoints = event->text().toUcs4();
-        assert(codepoints.size() == 1);
+        assert(codepoints.size() >= 1);
```
christianparpart, line 410: i just noticed now, I am confused. Isn't assert() to only evaluate in debug mode? Our release builds should have that all not generated any code. :thinking:

Yaraslaut: In crispy we have `Require` , this `assert` is from `<cassert>`

Yaraslaut: So this is archlinux specific issue, since debug build is packaged there

PR #1716, opened 2025-01-26
Do not crop underflow glyphs if they fit in cell size height

Yaraslaut:
Closes https://github.com/contour-terminal/contour/issues/1603
Some fonts position characters to underflow, for example char `p` so we should not crop such characters as long as they are not exceeding cell size


christianparpart (approved):
interesting. thanks.

PR #1717, opened 2025-01-27
Small changes to validate key mapping documentation and no assert when compiling with CMAKE_BUILD_TYPE None

Yaraslaut:
Closes https://github.com/contour-terminal/contour/issues/1622
Also, remove assertions when compiling on arch linux with CMAKE_BUILD_TYPE None


christianparpart (approved):
as discussed. thanks :)

PR #1718, opened 2025-02-02
Instead of option_as_alt introduce command_as_alt and fix behaviour

### LINKED ISSUE #1607: MacOS transform special symbols into expected key press
On MacOS some symbols generated from combintation of `Opt-{}` and `Opt-Shift-{}` keys,  if we use Opt as an Alt key, we want to send to the running application not special symbol, but actual combination of pressed keys, for example `Alt-j` instead of `\Delta` 
We want to make it configurable 

Images for layout from @whisperity 
image
image

Issue is similar to https://github.com/contour-terminal/contour/issues/1588

Yaraslaut:
on my system macOS shows Sigma symbol when I press Option+W for example, same happening to other keybindings, maybe it is better to use command instead of option as an alt alternative on the macos then? @christianparpart 
Also, PR fixes option_as_alt behaviour since we need to check for __APPLE__ macro and send press event to trigger user key_mappings action

Closes #1607.

### DIFF
--- a/src/contour/Config.cpp
+++ b/src/contour/Config.cpp
@@ -398,6 +398,7 @@ void YAMLConfigReader::loadFromEntry(YAML::Node const& node, std::string const& 
         loadFromEntry(child, "insert_after_yank", where.insertAfterYank);
         loadFromEntry(child, "bell", where.bell);
         loadFromEntry(child, "wm_class", where.wmClass);
+        loadFromEntry(child, "command_as_alt", where.commandKeyAsAlt);
         loadFromEntry(child, "margins", where.margins);
         loadFromEntry(child, "terminal_id", where.terminalId);
         loadFromEntry(child, "frozen_dec_modes", where.frozenModes);
--- a/src/contour/Config.h
+++ b/src/contour/Config.h
@@ -445,7 +445,7 @@ struct TerminalProfile
     ConfigEntry<HyperlinkDecorationConfig, documentation::HyperlinkDecoration> hyperlinkDecoration {};
 
     ConfigEntry<std::string, documentation::WMClass> wmClass { CONTOUR_APP_ID };
-    ConfigEntry<bool, documentation::OptionKeyAsAlt> optionKeyAsAlt { false };
+    ConfigEntry<bool, documentation::CommandKeyAsAlt> commandKeyAsAlt { true };
 };
 
 const InputMappings defaultInputMappings {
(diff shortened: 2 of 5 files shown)


christianparpart:
I *think* I got inspired by some other app to map "option as alt". 🤔

src/contour/Config.h
```diff
@@ -445,7 +445,7 @@ struct TerminalProfile
     ConfigEntry<HyperlinkDecorationConfig, documentation::HyperlinkDecoration> hyperlinkDecoration {};
 
     ConfigEntry<std::string, documentation::WMClass> wmClass { CONTOUR_APP_ID };
-    ConfigEntry<bool, documentation::OptionKeyAsAlt> optionKeyAsAlt { false };
+    ConfigEntry<bool, documentation::CommandKeyAsAlt> commandKeyAsAlt { true };
```
christianparpart, line 448: do we really want to default to `true` here?

src/contour/contour.yml
```diff
@@ -246,7 +246,7 @@ profiles:
         # This value is ignored on other platforms.
         #
         # Default: false
-        option_as_alt: false
+        command_as_alt: false
```
christianparpart, line 249: above it says default is true, here false 🤔

src/contour/Config.cpp
```diff
@@ -398,6 +398,7 @@ void YAMLConfigReader::loadFromEntry(YAML::Node const& node, std::string const&
         loadFromEntry(child, "insert_after_yank", where.insertAfterYank);
         loadFromEntry(child, "bell", where.bell);
         loadFromEntry(child, "wm_class", where.wmClass);
+        loadFromEntry(child, "command_as_alt", where.commandKeyAsAlt);
```
christianparpart, line 401: option_as_alt was not removed?

christianparpart: ah, looks like it was not there since config refactor. oops :)

src/contour/Config.h
Yaraslaut: I think that this is better for the default since user do not need to tweak config to use Alt+something key mappings on mac

src/contour/Config.cpp
christianparpart: After discussion, we agreed to this:

```yaml

# that would be the default (any modifier can be put here)
meta_key: Meta

# alternatively, users can customize and distinguish between OSs, 
# in order to share the same config across different machines/OSs
meta_key:
    default: Meta
    macos: Control

# example use
input_mapping:
    - { mods: [Meta, Alt],      mouse: Left,        action: FollowHyperlink }
    # Meta here is translated as configured above
```

PR #1720, opened 2025-02-19
Refactor shaper

Yaraslaut:
refactor function and update include for range-v3

### DIFF
--- a/src/text_shaper/shaper.cpp
+++ b/src/text_shaper/shaper.cpp
@@ -30,29 +30,37 @@ namespace
                                      vector<uint8_t>& outputBitmap) noexcept
     {
         outputBitmap.resize(outputSize.area() * NumComponents);
-        uint8_t* d = outputBitmap.data();
-        for (size_t i = 0, sr = 0; i < *outputSize.height; i++, sr += factor)
+
+        auto index = [&](size_t i, size_t j, size_t component) {
+            return 4 * i * unbox(outputSize.width) + j * 4 + component;
+        };
+
+        for (auto const i: ::ranges::views::iota(size_t { 0 }, unbox(outputSize.height)))
         {
-            for (size_t j = 0, sc = 0; j < *outputSize.width; j++, sc += factor, d += 4)
+            for (auto const j: ::ranges::views::iota(size_t { 0 }, unbox(outputSize.width)))
             {
-                // calculate area average
+                //  calculate area average
                 std::array<unsigned int, NumComponents> components { {} };
                 unsigned int count = 0;
-                for (size_t y = sr; y < min(sr + factor, unbox<size_t>(inputSize.height)); y++)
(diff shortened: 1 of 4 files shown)


src/text_shaper/shaper.cpp
```diff
@@ -30,29 +30,37 @@ namespace
...
                 unsigned int count = 0;
-                for (size_t y = sr; y < min(sr + factor, unbox<size_t>(inputSize.height)); y++)
+                for (auto y: ::ranges::views::iota(i * factor,
+                                                   min((i + 1) * factor, unbox<size_t>(inputSize.height))))
                 {
                     uint8_t const* p =
-                        inputBitmap.data() + (y * unbox<size_t>(inputSize.width) * 4) + (sc * 4);
-                    for (auto x = sc; x < min(sc + factor, unbox<size_t>(inputSize.width)); x++, count++)
+                        inputBitmap.data() + (y * unbox<size_t>(inputSize.width) * 4) + (j * factor * 4);
+                    for ([[maybe_unused]] auto _: ::ranges::views::iota(
```
christianparpart, line 50:
```suggestion
                    for ([[maybe_unused]] auto const _: ::ranges::views::iota(
```

src/text_shaper/shaper.cpp
```diff
@@ -30,29 +30,37 @@ namespace
...
+                             j * factor, min((j + 1) * factor, unbox<size_t>(inputSize.width))))
                     {
-                        for (size_t i = 0; i < NumComponents; ++i)
-                            components[i] += *(p++);
+                        ++count;
+                        for (auto componentIndex: ::ranges::views::iota(size_t { 0 }, NumComponents))
```
christianparpart, line 54:
```suggestion
                        for (auto const componentIndex: ::ranges::views::iota(size_t { 0 }, NumComponents))
```

src/text_shaper/shaper.cpp
```diff
@@ -30,29 +30,37 @@ namespace
...
+                            components[componentIndex] += *(p++);
                     }
                 }
 
                 if (count)
                 {
-                    for (size_t i = 0; i < NumComponents; ++i)
-                        d[i] = static_cast<uint8_t>(components[i] / count);
+                    for (auto componentIndex: ::ranges::views::iota(size_t { 0 }, NumComponents))
```
christianparpart, line 61:
```suggestion
                    for (auto const componentIndex: ::ranges::views::iota(size_t { 0 }, NumComponents))
```

christianparpart (commented):
I know it's a minor and I did not catch it when we both looked at it interactively, but would be nice to keep immutable variables declared as such :)

src/text_shaper/shaper.cpp
Yaraslaut: updated

christianparpart (dismissed):
after fixed tests, LGTM :-)

PR #1723, opened 2025-03-02
[GitHub] Fix bug report template drop-down selection elements

whisperity:
ARM64 and AArch64 are the same thing. Some platforms and tools prefer using this or that name for it. Similarly, in many cases especially on Linux, different distributions and ecosystems use `amd64` instead of `x86-64` or `x86_64` to refer to what is essentially the same thing.

For example, Ubuntu uses `amd64`, `arm64`, and `i386` in the package names; see for example: https://launchpad.net/ubuntu/noble/+package/gcc-12-base.

Running the stock util command `arch` on Mac says `arm64` or `i386` (even for 64-bit Mac Intel builds: try running `arch -x86_64 zsh` to spawn an emulated version of the shell and then call `arch` inside, it will still say `i386`!), but the Linux `arch` command (and, by that logic, `uname`) reports `aarch64` and `x86_64` for the ARM vs. PC line.

### DIFF
--- a/.github/ISSUE_TEMPLATE/Bug_Report.yml
+++ b/.github/ISSUE_TEMPLATE/Bug_Report.yml
@@ -11,7 +11,7 @@ body:
 - type: input
   attributes:
     label: Contour Terminal version
-    placeholder: "0.2.3"
+    placeholder: "0.1.2.9999"
     description: |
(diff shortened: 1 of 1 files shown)


Yaraslaut (approved):
Thanks :)

PR #1731, opened 2025-03-06
Update adoption statuses for color-palette-update-notifications

cor:


christianparpart (approved):
Thank you very much, @cor for helping us staying as up to date as possible. :)

Issue #1728, opened 2025-03-02
Terminal tabs: `SwitchTabToLeft`/`SwitchTabToRight` keybind is executed not in the active window


whisperity:
### Contour Terminal version

0.6.1.7494

### Installer source

GitHub: release page

### Operating System

MacOS San Sequoia 15.3.1

### Architecture

ARM64

### Other Software

—

### Steps to reproduce

1. Start contour normally.
2. `echo 1` to write a `1` into the first shell
3. create a new tab
4. `echo 2` to write a `2` into the second shell
5. create a new tab
6. `echo 3` to write a `3` into the third shell

Now three windows will be open, with the oldest originally opened one and the newest one showing `3`, while the middle one is showing `2`.

![Image](https://github.com/user-attachments/assets/6ff76aac-272d-444e-9e23-7f018861ef3a)

With the **middle** window focused, execute the `SwitchTabToRight` keybind. Observe that the **oldest** window (which previously showed the `3` shell, just like the *newest* window) now switched to the `1` shell.

![Image](https://github.com/user-attachments/assets/a789256c-e40d-486a-91fc-735f355a33b1)

No matter which window is actually in focus, it looks like the switch keybinds are always handled by the original/main/oldest window of the process.

### Expected Behavior

The keybind should be handled by the window that is currently in focus.

Interestingly, other types of keybinds, such as `ViNormalMode` or `ScrollMarkUp` are properly handled by the "active" window.

### Actual Behavior

–

Yaraslaut:
See comment: https://github.com/contour-terminal/contour/issues/1726 i think we should not manage tabs between windows

Issue #1725, opened 2025-03-02
[0.6.1.7494 Regression] Switching between windows of the same Contour process broken (likely due to terminal tabs feature)


whisperity:
### Contour Terminal version

0.6.21.7494

### Installer source

GitHub: release page

### Operating System

MacOS San Sequoia 15.3.1

### Architecture

ARM64

### Other Software

—

### Steps to reproduce

It may be relevant, but have `spawn_new_process: false`. It is important that new terminals under the same binary (the MacOS idiosyncratic way of dealing with this concept…) has the issue. If a new top-level binary is started, the Contour icon keeps on multiplying as the system considers the terminals entirely separate — and the issue doesn't arise!

1. Open Contour and put something into the existing window in order to uniquely identify it, e.g., the shell's PID `$$`. 😄

![A single freshly started Contour instance running.](https://github.com/user-attachments/assets/56470002-e7c6-42f8-957b-5915e67e2a15)

2. Press the `NewTerminal` keybind (in my case it is <kbd>⌃ Control</kbd>–<kbd>⇧ Shift</kbd>–<kbd>N</kbd>).
3. A new Contour window opens under the purview of the same application (the Contour icon doesn't multiply on the Dock).
4. ⚠ **The existing window (now in the background) resets itself to show a new session as well.
5. Type an identifying command into the newly spawned foregrounded window which shows that the two windows show the same terminal tab.

![Two contour windows running the exact same session.](https://github.com/user-attachments/assets/cdca9544-3420-48a2-958a-d3b9b05886fd)

6. Normally, switching the focus between two windows of the same program would be possible by right clicking the icon on the Dock, and selecting a window. (Proper apps also offer a _Window_ drop-down in the menu bar.)

![Right-click on the running app menu](https://github.com/user-attachments/assets/e6a3341d-b012-454e-9ab3-afc7c9fc8924)

7. Clicking the "other" Contour here (the one on the bottom) brings the "older" window into the foreground, but they are still attached to the same session.
8. `exit`ing the newer session does force the older window to get back to "Contour 1" in our case, but the newer window **infinitely hangs** on the exit command, never to actually exit. `htop` reveals a zombie `zsh` process.

### Expected Behavior

Creating a new terminal window and session should not destroy the user's ability to interface with the old already open window.

### Actual Behavior

— See earlier in bug report. —

PR #1739, opened 2025-03-23
initial take on multiple windows support

Yaraslaut:
closes https://github.com/contour-terminal/contour/issues/1728
closes https://github.com/contour-terminal/contour/issues/1725
closes https://github.com/contour-terminal/contour/issues/1728


src/contour/TerminalSessionManager.cpp
```diff
@@ -107,52 +122,101 @@ TerminalSession* TerminalSessionManager::activateSession(TerminalSession* sessio
...
-    managerLog()("switch to previous tab (current: {}, previous: {})",
-                 getSessionIndexOf(_activeSession).value_or(-1),
-                 getSessionIndexOf(_previousActiveSession).value_or(-1));
+    return;
+    // managerLog()("switch to previous tab (current: {}, previous: {})",
+    //              getSessionIndexOf(_displayStates[activeDisplay].activeSession).value_or(-1),
+    //              getSessionIndexOf(_displayStates[activeDisplay].previousActiveSession).value_or(-1));
 
-    activateSession(_previousActiveSession);
+    // activateSession(_displayStates[activeDisplay].previousActiveSession);
```
christianparpart, line 219: Let's not drop this (useful) feature. :)

src/contour/TerminalSessionManager.cpp
```diff
@@ -267,21 +335,36 @@ void TerminalSessionManager::removeSession(TerminalSession& thatSession)
...
-            return _sessions[currentIndex + 1];
-        else if (currentIndex > 0)
-            return _sessions[currentIndex - 1];
-        else
-            return nullptr;
-    }();
-    managerLog()("Calculated next \"previous\" session index {}",
-                 getSessionIndexOf(_previousActiveSession).value_or(-1));
+    tryFindSessionForDisplayOrClose();
+    //_app.onExit(thatSession); // TODO: the logic behind that impl could probably be moved here.
```
christianparpart, line 339: > the logic behind that impl could probably be moved here

Why?

`onExit` is a hook. Ok, the function it is in currently is called "removeSession", which does not indicate, that this session is actually to be closed/terminated. So yeah, we could move it. But then we should also do it, such that we can remove this line here completely :)

src/contour/TerminalSessionManager.cpp
```diff
@@ -267,21 +335,36 @@ void TerminalSessionManager::removeSession(TerminalSession& thatSession)
...
+
+void TerminalSessionManager::tryFindSessionForDisplayOrClose()
+{
+    for (auto& session: _sessions)
+    {
+        bool saveToSwitch { true };
+        // check if session is not used by any display and then switch
+        for (auto& [display, state]: _displayStates)
+        {
+            if ((state == session) && (display != nullptr))
```
christianparpart, line 350:
```suggestion
            if (display && state == session)
```

src/contour/TerminalSessionManager.h
```diff
@@ -50,32 +52,38 @@ class TerminalSessionManager: public QAbstractListModel
     [[nodiscard]] int count() const noexcept { return static_cast<int>(_sessions.size()); }
 
     void updateColorPreference(vtbackend::ColorPreference const& preference);
-    display::TerminalDisplay* display = nullptr;
-    TerminalSession* getSession() { return _sessions[0]; }
+
+    display::TerminalDisplay* activeDisplay = nullptr;
```
christianparpart, line 56: let's not mix fields with functions. Also, I think this should be private, _and_ it is missing a private datamember prefix then (`_activeDisplay`) :)

src/contour/TerminalSessionManager.h
```diff
@@ -50,32 +52,38 @@ class TerminalSessionManager: public QAbstractListModel
...
+        return getSessionIndexOf(_displayStates[activeDisplay]).value();
     }
 
     void updateStatusLine()
     {
-        if (!_activeSession)
+        if (!_displayStates[activeDisplay])
             return;
-        _activeSession->terminal().setGuiTabInfoForStatusLine(vtbackend::TabsInfo {
+        _displayStates[activeDisplay]->terminal().setGuiTabInfoForStatusLine(vtbackend::TabsInfo {
```
christianparpart, line 86:
```suggestion
        if (auto* displayState = _displayStates[activeDisplay]; displayState->terminal().setGuiTabInfoForStatusLine(vtbackend::TabsInfo {
```

Avoid the double-lookup to `_displayStates[activeDisplay]`.

src/contour/display/TerminalDisplay.h
```diff
@@ -36,8 +36,6 @@ namespace contour::display
 
 class OpenGLRenderer;
 
-// It currently just handles one terminal inside, but ideally later it can handle
-// multiple terminals in tabbed views as well tiled.
 class TerminalDisplay: public QQuickItem
```
christianparpart, line 39: Why did you drop the comment entirely?

src/vtbackend/Terminal.cpp
```diff
@@ -1262,9 +1262,9 @@ void Terminal::verifyState()
     Require(*_currentMousePosition.column < *thePageSize.columns);
     Require(*_currentMousePosition.line < *thePageSize.lines);
 
-    Require(_hostWritableStatusLineScreen.pageSize() == _indicatorStatusScreen.pageSize());
-    Require(_hostWritableStatusLineScreen.pageSize().lines == LineCount(1));
-    Require(_hostWritableStatusLineScreen.pageSize().columns == _settings.pageSize.columns);
+    // Require(_hostWritableStatusLineScreen.pageSize() == _indicatorStatusScreen.pageSize());
+    // Require(_hostWritableStatusLineScreen.pageSize().lines == LineCount(1));
+    // Require(_hostWritableStatusLineScreen.pageSize().columns == _settings.pageSize.columns);
```
christianparpart, line 1267: Why did you comment them out? Did they fail? Still? Maybe they fail for a reason, that we should address?

src/contour/TerminalSessionManager.cpp
Yaraslaut: this comment was here before :)

src/contour/display/TerminalDisplay.h
Yaraslaut: yes, suppose should adjust it to show what is actually happening with the display now

christianparpart (approved):
Let's do this! 😄

PR #1751, opened 2025-04-25
a workaround to issue #1750

### LINKED ISSUE #1750: scaleDownExplicit does not compile with pedantic and NumComponents=3 (rgb)
scaleDownExplicit failed to compile with GCC 14.2.1 20250110 (Red Hat 14.2.1-7).
It complains about dereferencing null pointer when trying to `resize()` outputBitmap for rgb (3)

From "tests" that I did I found that _any_ odd multiplier fails the same.
I have no idea why :shrug

regevran:
- **just a patch to compile rbg**
- **added a patch to metainfo.xml**

## Description

this is not the required solution to #1750 just a workaround that works for me.
I added this PR for visibility, maybe someone else will need it
I don't think it should be merged

Fixes: #1750

## Motivation and Context
Just to overcome compilation error

## How Has This Been Tested?

- [X] Please describe how you tested your changes.
Just compiled

## Checklist:

- [-] I have read the [**`CONTRIBUTING`**](https://github.com/contour-terminal/contour/blob/master/docs/CONTRIBUTING.md) document in my spoken language, and understand the terms
- [-] I have updated (or added) the documentation accordingly.
- [-] I have added tests to cover my changes.
- [-] I have gone through all the steps, and have thoroughly read the instructions

### DIFF
--- a/src/text_shaper/shaper.cpp
+++ b/src/text_shaper/shaper.cpp
@@ -23,13 +23,12 @@ namespace text
 namespace
 {
     template <std::size_t NumComponents>
-    constexpr void scaleDownExplicit(vector<uint8_t> const& inputBitmap,
-                                     vtbackend::ImageSize inputSize,
-                                     vtbackend::ImageSize outputSize,
-                                     size_t factor,
-                                     vector<uint8_t>& outputBitmap) noexcept
+    constexpr vector<uint8_t> scaleDownExplicit(vector<uint8_t> const& inputBitmap,
+                                                vtbackend::ImageSize inputSize,
+                                                vtbackend::ImageSize outputSize,
+                                                size_t factor) noexcept
     {
-        outputBitmap.resize(outputSize.area() * NumComponents);
+        vector<uint8_t> ret(outputSize.area() * NumComponents);
 
         auto index = [&](size_t i, size_t j, size_t component) {
             return (4 * i * unbox(outputSize.width)) + (j * 4) + component;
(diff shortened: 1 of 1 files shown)


Yaraslaut:
Thanks for the fix,  last time i had to refactor shaper.cpp was because some compiler errors with pedantic compilation as well :) 
also i think that the following fix can be better: 
```
diff --git a/src/text_shaper/shaper.cpp b/src/text_shaper/shaper.cpp
index 3fc65f35..ad79445f 100644
--- a/src/text_shaper/shaper.cpp
+++ b/src/text_shaper/shaper.cpp
@@ -29,6 +29,7 @@ namespace
                                      size_t factor,
                                      vector<uint8_t>& outputBitmap) noexcept
     {
+        outputBitmap.reserve(outputSize.area() * NumComponents);
         outputBitmap.resize(outputSize.area() * NumComponents);

         auto index = [&](size_t i, size_t j, size_t component) {

```
also we do not put changes that are not relevant for the users in metainfo file, so if you want to update your PR, please do it, otherwise let me know whether i need to create another one

regevran:
> also we do not put changes that are not relevant for the users in metainfo file

I just wanted to see my commit `sha` with `--version`, and I figured out that it is either deleting `metainfo` completely or adding a record in it. Is there other way?

I would expect the build to pick the latest SHA _unless_ instructed otherwise (e.g. `-DUSE_META_INFO`)

regevran:
> let me know whether i need to create another one

I'll try to come up with something that is not the also weird looking `reserve+resize` (better than mine, though :-)).
I'll update soon if I gave up, or that I'll update the PR.

regevran:
> or that I'll update the PR

Updated - but did NOT have the time to test, and make sure that semantically everything is the same.
Overall - I am OK with whatever fix you would like.
I have no strong feelings towards any of them.
Feel free to act as you think is the best, I am in favor for your decision.

regevran:
The PR needs cleaning, remove metainfo commjt for example, squash the changes into one commit, etc.
Maybe also remove `std::move` at caller site. 
However, I left this work undone to first get a feedback on the return-instead-of-output-parameter approach.

Yaraslaut:
thanks for the PR, i will merge it

PR #1753, opened 2025-04-27
Let configure compilation with -march=native flag

Yaraslaut:
Allow configure compilation to use native flag, since libunicode can benefit from simd native implementation


christianparpart (approved):
thank you :)

possibly also have a specific release-native target in cmake presets?

PR #1755, opened 2025-05-02
support mouse mode for WSL under Windows 10

tfriedel:
## Description

I had the problem that when I used Contour with WSL2 and Tmux the mouse clicks wouldn't be sent to Tmux. See:
https://github.com/contour-terminal/contour/discussions/1562

It turned out, this problem was not an issue with Windows 11. 
The same issue existed with other ConPty based terminal emulators, like Alacritty and WezTerm.
There seems to be a newer ConPty that comes with Windows11 that doesn't have this problem. In Alacritty and WezTerm it can be resolved by using ConPty from Openconsole. 

See:
https://github.com/alacritty/alacritty/issues/4794
https://github.com/alacritty/alacritty/issues/1663

## Motivation and Context

Allows me to use the Mouse with Tmux under Win10.

## How Has This Been Tested?

I copied openconsole.exe and conpty.dll from WezTerm, used Tmux with WSL2 and the mouse worked. Some more testing could be beneficial. I also didn't test if it still works if it can't find conpty.dll.

I faced an issue with some stack overflow of a configuration object and moved it from stack to heap. Should be fine, but it's a bit unrelated and I don't understand why I faced this problem to begin with. 

## Checklist:

No I haven't done that. If someone else could help here, it would be appreciated. 


Yaraslaut:
Hi @tfriedel  thanks for the PR, 
* some ci checks are failing, can you please fix it, let me know if you need any help on this one
* can you please specify which error you had with  `crispy::overloaded` function, since your changes introduce some unnecessary indirection and i would like to know if this is something we can avoid, since idea is to have a single place to add function to the custom overload set, but after this changes, to add one more function in the set will require some boilerplate

tfriedel:
Hi @Yaraslaut 

The error I got that prompted the move of code from stack to heap was:

Run-Time Check Failure #2 - Stack around the variable 'completeOverload' was corrupted.

Previously during compilation I got this error:
S:\projects\contour\src\contour\Config.cpp(1,1): error C1128: number of sections exceeded object file format limit: compile with /bigobj [S
:\projects\contour\build\src\contour\contour.vcxproj]

Which I fixed with /bigobj. I also tried to increase the size of the stack with 
```set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /STACK:4194304")```
but this didn't help.


I'll go through the CI issues and see what I can do.

tfriedel:
@Yaraslaut 
I reverted the changes to Config.cpp and built it again in release mode "--preset windows-cl-release". Compilation worked and I didn't observe the runtime error. 
I do observe it if I follow the instructions from the README which builds in debug mode.
Not sure if there's a potential issue in the release build still. Maybe  you have some idea what's going on there?

Yaraslaut:
Thanks for update @tfriedel, I suppose this is observed only on msvc and not sure what exactly is the cause, this might be connected with the way msvc materialize templates and our usage of template arguments as a documentation for config entries, anyway now everything looks good and only info in the metafile is missing, you need to add some info for the users here: https://github.com/contour-terminal/contour/blob/118fd9c1cc3c5a2d718838b87fb20e52df299940/metainfo.xml#L107

Yaraslaut (approved):
Many thanks :)

christianparpart (approved):
Many thanks, @tfriedel, for helping us improving Contour. :)

PR #1773, opened 2025-06-21
Refactor: Unify vector declaration style in files with 'using namespace std"

pelicans5:
- Standardize vector declarations to either `vector` or `std::vector` per-file
- For each file containing 'using namespace std':
  - Analyze existing usage patterns of std types
  - Convert all vector declarations to match the prevalent style:
    * Use unqualified `vector` if majority usage omits std::
    * Use fully qualified `std::vector` if majority explicitly scopes types
- Maintains internal consistency within each file
- Preserves existing coding patterns and style preferences
- No functional changes - purely stylistic refactoring

### DIFF
--- a/src/vtbackend/Selector.cpp
+++ b/src/vtbackend/Selector.cpp
@@ -85,7 +85,7 @@ bool Selection::intersects(Rect area) const noexcept
     return false;
 }
 
-std::vector<Selection::Range> Selection::ranges() const
+vector<Selection::Range> Selection::ranges() const
 {
     auto [result, from, to] = prepare(*this);
 
--- a/src/contour/Config.cpp
+++ b/src/contour/Config.cpp
@@ -108,9 +108,9 @@ namespace
                     "Could not create directory {}. {}", path.parent_path().string(), ec.message()) };
     }
 
-    vector<fs::path> getTermInfoDirs(optional<fs::path> const& appTerminfoDir)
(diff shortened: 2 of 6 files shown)


src/vtbackend/Selector.cpp
```diff
@@ -85,7 +85,7 @@ bool Selection::intersects(Rect area) const noexcept
     return false;
 }
 
-std::vector<Selection::Range> Selection::ranges() const
+vector<Selection::Range> Selection::ranges() const
```
christianparpart, line 88: In Config.cpp you've added `std::` but here you remove it. Why?

I think it is good to never import namespaces into global scope. So it would make sense to drop `using namespace std;` in global scope and fully qualify instead - unless we'd need to qualify a lot within a function body, where it would make sense to import or import parts of a namespace within that scope. This also reduces compilation times as it reduces lookup, and keeps ambiguity low. :)

christianparpart (commented):
Hey @cy-by-side.

Many thanks for your contribution. I don't mind either way in .cpp files. I just wonder why you've picked sometimes one way, and sometimes another.

src/vtbackend/Selector.cpp
codeberg-user: For the consistency. Config.cpp is using std::vector everywhere but others are not.

Issue #1774, opened 2025-07-17
compilation failed due to null pointer dereference - II


regevran:
### Contour Terminal version

master

### Installer source

GitHub: Source code

### Operating System

Fedora Linux 42

### Architecture

x86-64 / amd64

### Steps to reproduce

`cmake --build --preset linux-release`

### Expected Behavior

Compilation succeed.

### Actual Behavior

```bash
[147/150] Building CXX object src/contour/CMakeFiles/contour.dir/ContourGuiApp.cpp.o
FAILED: src/contour/CMakeFiles/contour.dir/ContourGuiApp.cpp.o
[...]

PR #1775, opened 2025-07-17
Fixed compilation error by spliting to an if statement

regevran:
## Description

Added an if staement to support constexpr usage
Fixes: https://github.com/contour-terminal/contour/issues/1774

## Motivation and Context

Without it pedentic does not work

## How Has This Been Tested?

compiled


christianparpart:
Thank you, @regevran 😄

PR #1782, opened 2025-08-29
Add empty check to JumpHistory.

Dooez:
## Description

Add empty history check to JumpHistory


uspasojevic96:
Please add entry in changelog

PR #1783, opened 2025-09-07
WIP: Box Drawings additions

Dooez:
## Description

Update and fix box drawings (most notably `Line::Double` symbols and corners)
Add git branch drawing symbols.
Add multiple styles of braile symbols.

## Motivation and Context

Some of the default box drawings were somewhat broken  (e.g. horizontal `Line::Double`)
Git branch symbols were initially added to [kitty](https://github.com/kovidgoyal/kitty/pull/7681). At the moment there are not many apps utilizing the symbols, however i want to use it myself :)  
Braile symbols are used for drawing graphs by e.g. `btop`. Rendering them in a special way makes neat looking graphs regardless of font size.

## How Has This Been Tested?
- [x] Box symbols are added to the source code comments and are rendered as expected when viewed in contour.
- [x] Git branch symbols are added to the source code comments and are rendered as expected when viewed in contour.
- [ ] Braile was looked at in btop.

## Checklist:


src/vtrasterizer/BoxDrawingRenderer.cpp
```diff
@@ -747,8 +878,8 @@ namespace detail
 
             auto const h = pixmap.size.height.as<unsigned>() - 1;
             auto const w = pixmap.size.width.as<unsigned>();
-            for (auto const y: ranges::views::iota(0u, pixmap.size.height.as<unsigned>()))
-                for (auto const x: ranges::views::iota(0u, pixmap.size.width.as<unsigned>()))
+            for (auto const y: iota(0u, pixmap.size.height.as<unsigned>()))
+                for (auto const x: iota(0u, pixmap.size.width.as<unsigned>()))
```
Yaraslaut: I am not sure about this changes, unqualified lookup is not that trivial and if we do not rely on it better to have explicit name of the function here

src/vtrasterizer/BoxDrawingRenderer.cpp
```diff
@@ -896,8 +1027,254 @@ namespace detail
...
             return image;
         }
-        // }}}
 
+        struct Braile
+        {
+            // 0 3
+            // 1 4
+            // 2 5
+            // 6 7
```
Yaraslaut, line 1008: I suppose this means something, but as it is not very understandable

src/vtrasterizer/BoxDrawingRenderer.cpp
```diff
@@ -46,16 +46,13 @@ namespace
...
+        auto const width = cellSize.width.as<size_t>();
         auto const height = cellSize.height.as<size_t>();
+        auto const pitch = width;
 
-        for (size_t i = 0; i < cellSize.height.as<size_t>(); ++i)
-        {
-            for (size_t j = 0; j < cellSize.width.as<size_t>(); ++j)
-            {
+        for (auto i: iota(0U, height))
+            for (auto j: iota(0U, width))
```
Yaraslaut: can you please use `const auto` and explicit `ranges::views::iota`

src/vtrasterizer/BoxDrawingRenderer.cpp
```diff
@@ -345,148 +376,250 @@ namespace detail
         };
 
         // U+2500 .. U+257F (128 box drawing characters)
+        constexpr auto BoxDrawingDefinitions = std::array<Box, 0x80> {
+            /*U+2500 ─ */ Box {}.horizontal(Light),
+            /*U+2501 ━ */ Box {}.horizontal(Heavy),
```
Yaraslaut, line 370: ah, great change to have a better formatting

src/vtrasterizer/BoxDrawingRenderer.cpp
```diff
@@ -345,148 +376,250 @@ namespace detail
...
+        {
+            bool standardBox = codepoint >= 0x2500 && codepoint <= 0x257F;
+            bool gitBranchBox = codepoint >= 0xF5D0 && codepoint <= 0xF5D0 + branchDrawingDefinitions.size();
+            return standardBox || gitBranchBox;
+        }
+        constexpr auto getBoxDrawing(char32_t codepoint) -> std::optional<Box>
+        {
+            bool standardBox = codepoint >= 0x2500 && codepoint <= 0x257F;
+            bool gitBranchBox = codepoint >= 0xF5D0 && codepoint <= 0xF5D0 + branchDrawingDefinitions.size();
+            if (not(standardBox || gitBranchBox))
```
Yaraslaut: please use `!`, we do not use alternative operators

src/vtrasterizer/BoxDrawingRenderer.cpp
```diff
@@ -896,8 +1027,254 @@ namespace detail
...
+                auto const fillRect = [&](size_t x0, size_t x1, size_t y0, size_t y1, uint8_t value) {
+                    for (auto const yi: iota(y0, y1))
+                        for (auto const xi: iota(x0, x1))
+                            image[(yi * width) + xi] = value;
+                };
+
+                auto xC = width / 2;
+                auto y0 = height / 4;
+                auto y1 = height / 2;
+                auto y2 = 3 * height / 4;
```
Yaraslaut, line 1056: this parameters xC, y0, y1 and y2 is extremely easy to swap, please preserve the boxing of the types and unbox only at the point when they can not be mistaken with other values, this will make lambda much safer to use

also, if we do not change the `value` we should not expose it as a parameter

src/vtrasterizer/BoxDrawingRenderer.cpp
```diff
@@ -896,8 +1027,254 @@ namespace detail
...
+                    fillRect(xC, width, y0, y1, 0xFF);
+                if (value & (1 << 6))
+                    fillRect(0, xC, 0, y0, 0xFF);
+                if (value & (1 << 7))
+                    fillRect(xC, width, 0, y0, 0xFF);
+                return image;
+            }
+            static auto buildSquare(uint8_t value, ImageSize size, size_t th, size_t ss, bool empty)
+                -> atlas::Buffer
+            {
```
Yaraslaut, line 1078: same for this function, names of the variables must be explained to some extent and strong typing must be preserved

src/vtrasterizer/BoxDrawingRenderer.cpp
```diff
@@ -896,8 +1027,254 @@ namespace detail
...
+                if (empty or (value & (1 << 7)))
+                    fillRect(x2, x3, y0, y1, 0xFF);
+                if (empty and not(value & (1 << 7)))
+                    fillRect(x2 + th, x3 - th, y0 + th, y3 - th, 0x00);
+                return downsample(image, 1, size, size / ss);
+            }
+
+            static auto buildCircle(uint8_t value, ImageSize size, size_t th, size_t ss, bool empty)
+                -> atlas::Buffer
+            {
```
Yaraslaut, line 1147: see comments for the few functions above

src/vtrasterizer/BoxDrawingRenderer.cpp
```diff
@@ -962,37 +1345,25 @@ auto BoxDrawingRenderer::createTileData(char32_t codepoint, atlas::TileLocation
...
+            return 1;
+        return val;
+    }();
+    auto tmp = buildBoxElements(codepoint, //
+                                _gridMetrics.cellSize,
+                                _gridMetrics.underline.thickness,
+                                supersamplingFactor);
+    if (!tmp)
+        return nullopt;
+    pixels = *tmp;
```
Yaraslaut, line 1341: std::move is missing, what there a problem with this one?

src/vtrasterizer/BoxDrawingRenderer.cpp
```diff
@@ -1019,19 +1390,21 @@ bool BoxDrawingRenderer::renderable(char32_t codepoint) noexcept
...
-           || ascending(0xEE00, 0xEE05)   // progress bar (Fira Code)
-           || codepoint == 0xE0B0         // 
-           || codepoint == 0xE0B2         // 
-           || codepoint == 0xE0B4         // 
-           || codepoint == 0xE0B6         // 
-           || codepoint == 0xE0BA         // 
-           || codepoint == 0xE0BC         // 
-           || codepoint == 0xE0BE         // 
+    return ascending(0x23A1, 0x23A6)          // mathematical square brackets
+                                              // || ascending(0x2500, 0x2590)   // box drawing, block elements
```
Yaraslaut: why was it commented out?

src/vtrasterizer/BoxDrawingRenderer.cpp
```diff
@@ -1576,176 +1950,364 @@ auto boxDashedVertical(auto& dashed, ImageSize size, int lineThickness)
...
+    auto xOffset = width / 2;
+    auto lightTh = (unsigned) lineThickness;
+    auto const ss = box.diagonalval != detail::NoDiagonal      //
+                            || box.arcURval != detail::NoLine  //
+                            || box.arcULval != detail::NoLine  //
+                            || box.arcBLval != detail::NoLine  //
+                            || box.arcBRval != detail::NoLine  //
+                            || box.circleval != detail::NoLine //
+                        ? supersampling
+                        : 1U;
```
Yaraslaut, line 1981: it is not easy to follow the logic here, maybe something of the form
```
auto cons samplingFactor = [] { 
    if(std::ranges::find_if({...}, [](auto v}{ return v != detail::NoLine}) return supersampling;
    return 1U;
}();
```
can be more readable

src/vtrasterizer/BoxDrawingRenderer.cpp
```diff
@@ -1576,176 +1950,364 @@ auto boxDashedVertical(auto& dashed, ImageSize size, int lineThickness)
...
+                                .x1 = (xOffset - lightTh / 2 + 2 * lightTh) * ss,
+                                .y0 = (yOffset - lightTh / 2 - lightTh) * ss,
+                                .y1 = (yOffset - lightTh / 2 + 2 * lightTh) * ss };
+            case Heavy:
+                return Center { .x0 = (xOffset - lightTh) * ss,
+                                .x1 = (xOffset + lightTh) * ss,
+                                .y0 = (yOffset - lightTh) * ss,
+                                .y1 = (yOffset + lightTh) * ss };
+            default:
+                // dashed lines are handled explictily
```
Yaraslaut: typo

src/vtrasterizer/BoxDrawingRenderer.cpp
```diff
@@ -1576,176 +1950,364 @@ auto boxDashedVertical(auto& dashed, ImageSize size, int lineThickness)
...
+                        image[(yi * width) + xi] = value;
+                }
+        };
+        auto radius = std::round(static_cast<double>(std::min(width, height)) / 2. * 0.75);
+        fillCircle(radius, 0xFF);
+        switch (box.circleval)
+        {
+            case detail::Line::Double: {
+                auto space = 1.35 * getThickness(Line::Light);
+                if (radius < 3. * getThickness(Line::Light) + space)
```
Yaraslaut: this values must be separated into `cosntexpr double` values so that we can understand what they mean

src/vtrasterizer/Pixmap.h
```diff
@@ -74,14 +79,13 @@ enum class Inverted : uint8_t
...
     NoArc,
-    TopLeft,
-    TopRight,
-    BottomRight,
-    BottomLeft
+    BR,
+    BL,
+    UL,
+    UR,
 };
```
Yaraslaut: if we change the values, can we also add a class there so that we can catch cases where cast was happening to prevent the errors since semantic meaning  has changed

src/vtrasterizer/BoxDrawingRenderer.cpp
Dooez: I am almost sure it was already used this way in couple of places in this file so I've updated to unify.  
If C++23 migration is close I suggest `stdr` and `stdv` aliases for future `std::ranges` and `std::views` temporarily binding them to `ranges`.

src/vtrasterizer/BoxDrawingRenderer.cpp
Dooez: This is how Braille dots are numbered. I will expand the comment to make it more descriptive.

src/vtrasterizer/BoxDrawingRenderer.cpp
Dooez: I've grouped git branch characters and box characters handling and moved this logic to a separate function.

I was thinking about splitting their handling. In that case this check will return and for branch characters something like `isGitBranchDrawing()` will be used.

src/contour/Config.cpp
```diff
@@ -1462,6 +1469,107 @@ void YAMLConfigReader::loadFromEntry(YAML::Node const& node, std::string const&
...
+void YAMLConfigReader::loadFromEntry(YAML::Node const& node,
+                                     std::string const& entry,
+                                     vtrasterizer::BoxDrawingRenderer::ArcStyle& where)
+{
+    using ArcStyle = vtrasterizer::BoxDrawingRenderer::ArcStyle;
+    if (auto const child = node[entry])
+    {
+        auto const lstyle = crispy::toLower(child.as<std::string>());
+        auto constexpr static Mappings = std::array {
+            std::pair { "", ArcStyle::Round },        //
```
Yaraslaut, line 1481: hm, why do you need `//` there ?

Dooez: Almost certainly go force formatting, I use this trick quite regularly. Might be an artifact from an earlier stage if it doesn't actually affect the formatting.

src/vtrasterizer/Pixmap.cpp
```diff
@@ -92,8 +91,8 @@ Pixmap& Pixmap::halfFilledCircleLeft()
         putpixel(x, y + (h / 2));
     };
     auto const radius = crispy::point { .x = w, .y = h / 2 };
-    drawEllipseArc(putAbove, size, radius, Arc::BottomLeft);
-    drawEllipseArc(putBelow, size, radius, Arc::TopLeft);
+    drawEllipseArc(putAbove, size, radius, Arc::UL);
+    drawEllipseArc(putBelow, size, radius, Arc::BR);
```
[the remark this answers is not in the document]
christianparpart: @Yaraslaut @Dooez Copilot was actually correct here, FYI. Fixed in #1837.

PR #1796, opened 2025-10-29
Fix lambda without return value

Yaraslaut:
Fixes an error where lambda does not return a value

### DIFF
--- a/src/vtrasterizer/BoxDrawingRenderer.cpp
+++ b/src/vtrasterizer/BoxDrawingRenderer.cpp
@@ -1954,6 +1954,7 @@ auto buildBox(detail::Box box, ImageSize size, int lineThickness, size_t supersa
             case Heavy4: return false;
         }
         assert(false);
+        return false; // use std::unreachable in C++23
     };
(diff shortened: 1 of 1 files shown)


christianparpart (approved):
oopsi, how did that happen :)

src/vtrasterizer/BoxDrawingRenderer.cpp
```diff
@@ -1954,6 +1954,7 @@ auto buildBox(detail::Box box, ImageSize size, int lineThickness, size_t supersa
             case Heavy4: return false;
         }
         assert(false);
+        return false;
```
christianparpart: p.s.: actually, I think we should use:


```suggestion
        std::unreachable();
```
instead.

https://en.cppreference.com/w/cpp/utility/unreachable.html

Yaraslaut: updated

Yaraslaut: oh, i forgot that we do not compile with c++23 but with c++20

PR #1797, opened 2025-10-31
Bugfix DECRQM (Dec Request Mode) response

jquast:
Interactive terminal sequence response for DEC Private Mode Query is invalid -- it may respond with a different mode and status than the one requested.

This causes programs to fail or mismatch contour's terminal capabilities. In my API I was expecting the requested mode value in the response and, not matching it for any modes under number 27, wrongly reported that Contour does not support *any* DEC Private Modes at all.

This bug is present in all versions of Contour supporting DEC Private Mode, since 256e313d and refactored in faa66e7d, the use of:

       static_cast<DECMode>(mode)

Is invalid, because the DECMode enum has "holes". It is not contiguous.

A new `fromDECModeNum()` function is provided, refactored form of `isValidDecMode()` -- returning an `optional DecMode`, existing function `isValidDECMode()` is refactored to check whether the optional value returned by 'fromDECModeNum' tests true for `has_value()`.

Given the provided test file `test/display-modes.sh`, we can make inquiry of DEC Private modes and see that DEC modes 1 through 27 are returned with non-matching response modes mismatched by enum item order and status:

    $ ./test/display-modes.sh
    Query mode 1: response=$'\E[?2;2$y' after 13ms
    Query mode 2: response=$'\E[?3;2$y' after 11ms
    Query mode 3: response=$'\E[?4;2$y' after 10ms
    Query mode 4: response=$'\E[?5;2$y' after 11ms
    Query mode 5: response=$'\E[?9;2$y' after 11ms
    Query mode 6: response=$'\E[?1000;2$y' after 12ms
    Query mode 7: response=$'\E[?1001;2$y' after 13ms
    Query mode 8: response=$'\E[?1002;2$y' after 13ms
    Query mode 9: response=$'\E[?1003;2$y' after 12ms
    Query mode 10: response=$'\E[?1048;2$y' after 14ms

After this bugfix, this is corrected:

    $ ./test/display-modes.sh
    Query mode 1: response=$'\E[?1;2$y' after 11ms
    Query mode 2: response=$'\E[?2;2$y' after 11ms
    Query mode 3: response=$'\E[?3;2$y' after 8ms
    Query mode 4: response=$'\E[?4;2$y' after 10ms
    Query mode 5: response=$'\E[?5;2$y' after 8ms
    Query mode 6: response=$'\E[?6;2$y' after 11ms
    Query mode 7: response=$'\E[?7;1$y' after 11ms
    Query mode 8: response=$'\E[?8;0$y' after 10ms
    Query mode 9: response=$'\E[?9;2$y' after 10ms
    Query mode 10: response=$'\E[?10;2$y' after 12ms

Further, looking at other uses of `isValidDECMode()`, some were unnecessary -- that they have the `DECMode` type bound to the variable is sufficient enough to describe that mode as valid, the check is superfluous and removed, eg:

      isValidDECMode(static_cast<unsigned int>(mode)));

the call to `toDECModeNum` is also superfluous and removed.

### DIFF
--- a/.github/actions/spelling/allow/terminal-terms.txt
+++ b/.github/actions/spelling/allow/terminal-terms.txt
@@ -154,4 +154,5 @@ cuu
 dcl
 dcs
 fbterm
+stty
 zellij


src/vtbackend/primitives.h
```diff
@@ -789,53 +790,56 @@ constexpr unsigned toDECModeNum(DECMode m)
...
+        case 2026: return DECMode::BatchedRendering;
+        case 2027: return DECMode::Unicode;
+        case 2028: return DECMode::TextReflow;
+        case 2029: return DECMode::MousePassiveTracking;
+        case 2030: return DECMode::ReportGridCellSelection;
+        case 2031: return DECMode::ReportColorPaletteUpdated;
+        case 8452: return DECMode::SixelCursorNextToGraphic;
     }
-    return false;
+    return std::nullopt;
```
christianparpart:
```suggestion
        case 8452: return DECMode::SixelCursorNextToGraphic;
        default: return std::nullopt;
}
```

This is purely to avoid static analysis complain about a missing case in the switch statement. :)

src/vtbackend/Screen.cpp
```diff
@@ -1762,20 +1762,20 @@ void Screen<Cell>::requestAnsiMode(unsigned int mode)
 template <CellConcept Cell>
 void Screen<Cell>::requestDECMode(unsigned int mode)
 {
-    auto const modeResponse = [&](auto mode) -> ModeResponse {
-        if (isValidDECMode(mode))
+    auto const modeEnum = fromDECModeNum(mode);
+
+    auto const modeResponse = [&]() -> ModeResponse {
```
christianparpart:
```suggestion
    auto const modeResponse = [this, mode]() -> ModeResponse {
        auto const modeEnum = fromDECModeNum(mode);
```

`modeEnum` is exclusively needed inside the lambda, so let's move it inside, and be explicit about our captures.

christianparpart (commented):
Many thanks for the fix. Is it okay to apply the minors, for you?

src/vtbackend/primitives.h
jquast: Sorry I don't know why, when I attempt to accept this by clicking "Commit suggestion", it reads:

>  Applying suggestions on deleted lines is currently not supported. Refresh and try again.

But refreshing doesn't work. It could be a github issue, anyway you are welcome to commit directly anything you like!

christianparpart: i've force-pushed into your branch to adapt the minors and also `git rebase`'d on top of latest master, preserving your commit as-is :)

PR #1798, opened 2025-11-02
Fix mouse scrolling being too slow

christianparpart:
Mouse scrolling was still too slow, way to slow. This one actually fixes it by reworking especially the mouse wheel's angle delta calculations.


Yaraslaut (approved):
lgtm, thanks

PR #1800, opened 2025-11-16
[CMake] Bump default Qt search path to C:/Qt/6.10.0 (from 6.6.1)

christianparpart:
Jumping Qt (Cmake) module search path to look for Qt version 6.10.0 instead of 6.6.1.
I am assuming that this won't break CI. It at least helps me not updating this var on windows dev box though ;)


Yaraslaut (approved):
Great, I think that we also use this version for static build

PR #1802, opened 2025-11-16
[CMake] Set CPACK_WIX_VERSION to 4.

christianparpart:
This helps me building conveniently at least on my workstation, as CPack did not seem to find WIX toolset otherwise.


Yaraslaut (approved):
Great :)

PR #1803, opened 2025-11-21
[Vi] Fix y0 crash.

### LINKED ISSUE #1801: VI normal mode `y0` causes a crash
# Steps to reproduce

1. start Contour
2. get into normal mode
3. type `y0`

this causes a crash, interestingly.

christianparpart:
Closes #1801

### DIFF
--- a/src/vtbackend/ViCommands.cpp
+++ b/src/vtbackend/ViCommands.cpp
@@ -349,10 +349,8 @@ void ViCommands::executeYank(ViMotion motion, unsigned count)
             break;
         }
         default: {
-            auto const [from, to] = translateToCellRange(motion, count);
-            // motion is inclusive but for yank we want to exclude the last cell which is the first cell of
-            // the next word
-            executeYank(from, { .line = to.line, .column = to.column - 1 });
+            auto const [from, to] = translateToCellRange(motion, count).ordered();
+            executeYank(from, to);
         }
         break;
     }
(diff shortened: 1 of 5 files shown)


src/vtbackend/ViCommands.cpp
```diff
@@ -349,10 +349,8 @@ void ViCommands::executeYank(ViMotion motion, unsigned count)
             break;
         }
         default: {
-            auto const [from, to] = translateToCellRange(motion, count);
-            // motion is inclusive but for yank we want to exclude the last cell which is the first cell of
-            // the next word
-            executeYank(from, { .line = to.line, .column = to.column - 1 });
+            auto const [from, to] = translateToCellRange(motion, count).ordered();
+            executeYank(from, to);
```
Yaraslaut, line 353: oh, interesting :)

Yaraslaut (approved):
LGTM, thanks

src/vtbackend/ViCommands.cpp
christianparpart: @Yaraslaut ordering wasn't the problem. That's for cleanness only. The problem was the off-by-on issue in the old code (and NOT being ordered), yielding in the `0` from `y0` pointing to the column `-1`. We did not need to -1 to begin with. I've tested it in with various possibilities, and added that test here as this was the one failing

Issue #1776, opened 2025-07-27
Control+key seems to only partially work


mrdgo:
### Contour Terminal version

Contour Terminal Emulator 0.6.2-master-f6ce54bb

### Installer source

Something else (please specify below)

### Operating System

Arch Linux (installed via `paru -S contour-git`)

### Architecture

x86 / i386

### Other Software

```
NVIM v0.12.0-dev-888+g95dfb063da
Build type: RelWithDebInfo
LuaJIT 2.1.1744317938
Compilation: /usr/bin/cc -O2 -g -DRELDEBUG -flto=auto -fno-fat-lto-objects -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wvla -Wdouble-promotion -Wmissing-noreturn -Wmissing-format-attribute -Wmissing-prototypes -fsigned-char -fstack-protector-strong -Wno-conversion -fno-common -Wno-unused-result -Wimplicit-fallthrough -fdiagnostics-color=always  -DUNIT_TESTING -D_GNU_SOURCE -DINCLUDE_GENERATED_DECLARATIONS -DUTF8PROC_STATIC -I/home/runner/work/neovim/neovim/.deps/usr/include/luajit-2.1 -I/home/runner/work/neovim/neovim/.deps/usr/include -I/home/runner/work/neovim/neovim/build/src/nvim/auto -I/home/runner/work/neovim/neovim/build/include -I/home/runner/work/neovim/neovim/build/cmake.config -I/home/runner/work/neovim/neovim/src -I/usr/include

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/local/share/nvim"
```

### Steps to reproduce

1. open a file that has too many lines to fit into view `nvim long-file.ext`
2. try to use `Ctrl+e` to scroll downwards
3. nvim behaves as if you only pressed `e` and moves the cursor to the end of the next word

The same thing happens for `Ctrl+y`, `Ctrl+u`, and `Ctrl+d`

### Expected Behavior

The aforementioned key combinations should:
- `Ctrl+e` scroll down one line
- `Ctrl+y` scroll up one line
- `Ctrl+d` scroll down half a page
- `Ctrl+u` scroll up half a page
- `Ctrl+f` scroll down a full page
- `Ctrl+b` scroll up a full page

### Actual Behavior

The control key is "swallowed" and thereby:
[...]

PR #1807, opened 2025-11-25
Adapts CSIu input protocol to match how recent fish shell implementation transmits it.

christianparpart:
However, I cannot really read that behaviour from https://sw.kovidgoyal.net/kitty/keyboard-protocol/

There was no example showing `CSI = u` nor `CSI = 5 u` without first entering CSIu mode.

It seems like fish shell is tweaking CSIu mode by setting/resetting without entering/leaving.

The given spec however also does not mention flags to default to 0 neither.

--

It *MAY* be that other CLI applications may also not enter/leave and just set/reset flags, so I think it MAY be good to have in anyways.

This may also relate to #1776. @mrdgo, if it is possible at all, maybe you want to give this branch a try?

### DIFF
--- a/metainfo.xml
+++ b/metainfo.xml
@@ -120,6 +120,7 @@
           <li>Fixes double line box drawing characters</li>
           <li>Fixes DECRQM (DEC Request Mode) response (#1797)</li>
           <li>Fixes Precondition failure when rendering vertically overflowing glyphs (#1805)</li>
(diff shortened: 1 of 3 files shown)


Yaraslaut (commented):
Hmmmmmm, shell specific handling
maybe we can figure out what should be the default value

christianparpart:
> Hmmmmmm, shell specific handling maybe we can figure out what should be the default value

I've tested kitty and ghostty. Turns out they all also behave like this, even though I did not find this in the document, they've implemented it like this. So I think it's good to go.

PR #1814, opened 2025-12-12
[contour] Fixes option_as_alt configuration option not working on macOS

### LINKED ISSUE #1812: In macOS, "option_as_alt: true" does not work
### Contour Terminal version

Contour Terminal Emulator 0.6.2-master-475a270b

### Installer source

GitHub: Source code

### Operating System

macOS Tahoe v26.1

### Architecture

AArch64 / ARM64

### Other Software

_No response_

### Steps to reproduce

1. Open `contour.yml` with text editor and set `option_as_alt` to `true`
2. Open contour and try to press some key combination with option key.

### Expected Behavior

It should recognize like Alt key. (e.g. Option+C behaves like Alt+C)

### Actual Behavior

Instead of recognizing as an Alt key, it inserts special characters, which is (AFAIK) default on most macOS apps. (e.g. Option+C inserts "ç")

### Additional notes

I noticed that in the `src/contour/helper.cpp`, there's a lowercased `__apple__`. Maybe this is the cause of the bug?

https://github.com/contour-terminal/contour/blob/475a270b44935073207f3663c7af3b2773429887/src/contour/helper.cpp#L382-L390

christianparpart:
closes #1812.

### DIFF
--- a/src/contour/helper.cpp
+++ b/src/contour/helper.cpp
@@ -379,10 +379,12 @@ bool sendKeyEvent(QKeyEvent* event, vtbackend::KeyboardEventType eventType, Term
         return true;
     }
 
-#if defined(__apple__)
-    if (0x20 <= key && key < 0x80 && (modifiers.alt() && session.profile().optionKeyAsAlt.value()))
+#if defined(__APPLE__)
+    if (0x20 <= key && key < 0x80
+        && (modifiers.test(Modifier::Alt) && session.profile().optionKeyAsAlt.value()))
     {
-        auto const ch = static_cast<char32_t>(modifiers.shift() ? std::toupper(key) : std::tolower(key));
+        bool const shiftPressed = modifiers.test(Modifier::Shift) ^ modifiers.test(Modifier::CapsLock);
+        auto const ch = static_cast<char32_t>(shiftPressed ? std::toupper(key) : std::tolower(key));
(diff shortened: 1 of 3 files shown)


src/contour/helper.cpp
```diff
@@ -379,10 +379,12 @@ bool sendKeyEvent(QKeyEvent* event, vtbackend::KeyboardEventType eventType, Term
         return true;
     }
 
-#if defined(__apple__)
-    if (0x20 <= key && key < 0x80 && (modifiers.alt() && session.profile().optionKeyAsAlt.value()))
+#if defined(__APPLE__)
+    if (0x20 <= key && key < 0x80
+        && (modifiers.test(Modifier::Alt) && session.profile().optionKeyAsAlt.value()))
```
Yaraslaut, line 384: hm, is it a more reliable api for testing modifiers by qt? maybe we should use it in other places

christianparpart: `Modifier::Alt` is already set. But we're not making use of this information because Qt's event struct does not just contain the basic key code (like 0x65) but also the fully translated text, such as ç for Alt+c (Option+c). Our config flag `option_as_alt` toggles that to be passed as Alt+c rather than ç, if set to true. This translation happens *only* in that code fragment (I just checked).

Issue #624, opened 2022-03-17
[Windows Installer] install into path not containing the app version


christianparpart:
We do not (want to) support multiple installed versions anyways. So let's get rid of the version component in the program folder.

side node: auto-updater.

PR #1833, opened 2025-12-24
Fixes windows MSI installer previously not allowing upgrades

christianparpart:
Trying to fix .MSI installer, to avoid requiring to manually uninstall a prior version of Contour before installing a newer version.

refs #624


Yaraslaut (approved):
Lgtm, many thanks :)

PR #1834, opened 2025-12-24
Fix static build images

Yaraslaut:
Fixes static build 
and uses LTS version of the qt such that we do not need to worry about version bump for a while

### DIFF
--- a/.github/actions/spelling/allow/allow.txt
+++ b/.github/actions/spelling/allow/allow.txt
@@ -25,3 +25,4 @@ nupkg
 nushell
 tablegen
 zypper
+openxr


.github/static/DockerQt
```diff
@@ -24,16 +24,16 @@ RUN chmod +x llvm.sh
 RUN ./llvm.sh 20
 
 WORKDIR /
-RUN wget https://download.qt.io/official_releases/qt/6.10/$QT_VERSION/single/qt-everywhere-src-$QT_VERSION.tar.xz
+RUN wget https://download.qt.io/official_releases/qt/6.8/$QT_VERSION/single/qt-everywhere-src-$QT_VERSION.tar.xz
```
christianparpart, line 27: `6.8` can be inferred from `$QT_VERSION`, through `sed` or `awk` etc, so we do not need to touch too many places - just for changing a single version number :)

.github/static/DockerUbuntu
```diff
@@ -19,6 +19,6 @@ RUN LD_LIBRARY_PATH=/usr/local/lib cmake -S . -B build -G Ninja \
     -D CMAKE_CXX_COMPILER=clang++\
     -D CMAKE_C_COMPILER=clang \
     -D CMAKE_CXX_FLAGS="-I/contour/build/_deps/yaml-cpp-src/include" \
-    -D CMAKE_PREFIX_PATH=/usr/local/Qt-6.10.0/lib/cmake
+    -D Qt6_DIR=/usr/local/Qt-6.8.3/lib/cmake/Qt6
```
christianparpart: here too, should use `$QT_VERSION`

.github/static/readme
```diff
@@ -1,7 +1,5 @@
 To build static image we need to build qt static docker container and put it on dockerhub :
 ```
-docker buildx build --tag qt_static:20.04 --build-arg VERSION=20.04 --build-arg ARCH=x86_64 --progress=plain -f .github/static/DockerQt --load .
-docker buildx build --tag qt_static:22.04 --build-arg VERSION=22.04 --progress=plain -f .github/static/DockerQt --load .
```
christianparpart, line 4: why did these 2 lines go away?

.github/static/DockerQt
Yaraslaut: unfortunately this is not a shell environment, not sure that docker provides any support to manipulate arguments

.github/static/DockerUbuntu
Yaraslaut: added variable to this script

.github/static/readme
Yaraslaut: we do not provide static builds for this versions so i dropped them just to not promote any support for this versions

PR #1835, opened 2025-12-25
Package upstream OpenConsole/ConPTY into Contour on Windows OS

christianparpart:
The goal here is to bring in Sixel support to Contour on Windows early (but also, to be generally always fresher on Windows, when it comes to ConPTY).

However, seems like the current approach is not enough to actually get sixel (DCS) being passed through.

- [x] add cmake option (default ON) whether to embed OpenConsole+ConPTY or not


src/vtpty/CMakeLists.txt
```diff
@@ -86,3 +86,54 @@ target_include_directories(vtpty PUBLIC
...
+if(WIN32)
+    option(CONTOUR_EMBED_OPENCONSOLE "Embed OpenConsole" ON)
+    if(NOT CONTOUR_EMBED_OPENCONSOLE)
+        message(STATUS "[vtpty] OpenConsole embedding disabled")
+    else()
+        message(STATUS "[vtpty] OpenConsole embedding enabled")
+        set(OPENCONSOLE_VERSION "1.24.3504.0")
+        set(OPENCONSOLE_VERSION_ALT "1.24.251216004-preview")
+        #set(OPENCONSOLE_VERSION "1.23.13503.0")
+        #set(OPENCONSOLE_VERSION_ALT "1.23.251216003")
```
Yaraslaut: Remove it :)

src/vtpty/CMakeLists.txt
```diff
@@ -86,3 +86,54 @@ target_include_directories(vtpty PUBLIC
...
+        set(OPENCONSOLE_URL "https://github.com/microsoft/terminal/releases/download/v${OPENCONSOLE_VERSION}/Microsoft.Windows.Console.ConPTY.${OPENCONSOLE_VERSION_ALT}.nupkg")
+        set(OPENCONSOLE_DEST "${CMAKE_BINARY_DIR}/_deps/OpenConsole-${OPENCONSOLE_VERSION}")
+        set(OPENCONSOLE_ZIP "${OPENCONSOLE_DEST}/OpenConsole.zip")
+
+        set(arch "x64")
+        if(CMAKE_SYSTEM_PROCESSOR MATCHES "ARM")
+            set(arch "arm")
+        endif()
```
Yaraslaut: Hm, maybe we can create global scope variable for the arch and use it here?

Yaraslaut (dismissed):
Great, thanks, 
Few small comments

src/vtpty/CMakeLists.txt
christianparpart: i kept it intentionally, as this is how it looks like for an actual (non-preview) release. 🤔

src/vtpty/CMakeLists.txt
christianparpart: this variable contents is specific to the nuget package internal structure, it maps the syntax of CMAKE_SYSTEM_PROCESSOR limited to how the naming of the directory contents inside the nuget package is

christianparpart: i should therefore chose a better variable name to reflect that :-D

christianparpart: renamed

src/vtpty/CMakeLists.txt
christianparpart: dropped

Issue #1821, opened 2025-12-17
Sixel images appear corrupted


adre5577:
### Contour Terminal version

Contour Terminal Emulator 0.6.1

### Installer source

Something else (please specify below)

### Operating System

Manjaro Linux 6.17.12-1-MANJARO

### Architecture

x86-64 / amd64

### Steps to reproduce

Contour installed from the Manjaro repository, which pulls from Arch upstream. This bug appears only in package version 0.6.1.7494-2. I downgraded to package version 0.6.1.7494-1, and images display correctly again.

Use any program to display sixel images. For example, `cat lady-of-shalott.six`. The image displays correctly in other terminal emulators.

### Expected Behavior

The image correctly displays, as in this PNG version:

<img width="1000" height="750" alt="Image" src="https://github.com/user-attachments/assets/4f03b2b5-2921-4f3c-8207-f200358a965e" />

### Actual Behavior

The image appears corrupted, as seen below:

<img width="997" height="731" alt="Image" src="https://github.com/user-attachments/assets/4fa976af-ad82-4d75-b24b-3e3331d16fd3" />

christianparpart:
Thanks. This is odd. I've tried it.

<img width="2049" height="1277" alt="Image" src="https://github.com/user-attachments/assets/1dcc184d-fca1-4ed0-9573-55fd97a9b334" />

But then pressing <kbd>Ctrl</kbd>+<kbd>+</kbd>, to increase font size, it looks like this:

<img width="2036" height="1261" alt="Image" src="https://github.com/user-attachments/assets/6a0351d5-91da-4a60-86e6-1a1bf311f215" />

So to me this looks like a problem of grid cell height and thus, font metrics.

How you created your image output worries me however even more. Not sure you this could happen. 🤔

christianparpart:
@adre5577 I've fixed at least the issue that I was facing. I even knew about this one already. I still don't know how on earth you could create that *art* that you showed, but it's likely that my fix here is affecting you somewhat positively too.

Are you happen to be able to build from latest master somehow? If not, we're soon (very soon) releasing a new version anyways. 😄

christianparpart:
I am going to setup a Manjaro for attempting to repro this. Thank you.

PR #1846, opened 2026-01-02
Revert a commit and fix clang compilation

Yaraslaut:
Fixes https://github.com/contour-terminal/contour/issues/1821


src/crispy/StrongHash.h
```diff
@@ -189,9 +189,9 @@ inline strong_hash strong_hash::compute(void const* data, size_t n) noexcept
 
     return strong_hash { hashValue };
 #else
-    auto const* i = (char8_t const*) data;
+    auto const* i = (uint8_t const*) data;
     auto const* e = i + n;
-    auto const result = fnv<char8_t, uint64_t>()(i, e);
+    auto const result = fnv<uint8_t, uint64_t>()(i, e);
```
christianparpart, line 194: This looks specifically like a signed/unsigned issue. I wonder why we did not catch this, or, why the compiler actually liked your signed version earlier better than my initial unsigned one 🤔

christianparpart:
for failing windows build, you need to `if` out the compile flag for MSVCs

src/crispy/StrongHash.h
christianparpart: @Yaraslaut ok just checked, because i got curious about that `#else` there. There's a `STRONGHASH_USE_INTRINSICS` cmake option, defaulting to `ON`, and that's what I am using. This is why i was not seeing this behaviour. I am surprised somebody sees the need to set this to off though. 🤔

Yaraslaut: hm `STRONGHASH_USE_INTRINSICS` cmake option does not set this compiler flag :)

PR #1854, opened 2026-01-10
Fix `crispy::fnv` to build with libc++ on FreeBSD

tagattie:
## Description

This PR makes the following adjustments to `crispy::fnv`:

1. Restrict string-related overloads to char only
    - Replace `std::basic_string<T>` / `std::basic_string_view<T>` with
`std::string` / `std::string_view`
    - Guard these overloads using SFINAE so they are only enabled when `T == char`
2. Add a generic hashing overload for trivially copyable types
    - Allows hashing arbitrary trivially copyable values by hashing their byte representation
    - Explicitly excludes `std::string` and `std::string_view`, which have dedicated overloads

### Background

While building contour on FreeBSD using libc++, compilation fails in `crispy::fnv` due to the use of `std::basic_string<T>` / `std::basic_string_view<T>` with non-character template parameters.

libc++ intentionally provides `std::char_traits` specializations only for character types (e.g. `char`, `wchar_t`). As a result, instantiating `std::basic_string_view<T>` / `std::basic_string<T>` for arbitrary `T` leads to compilation errors.

### Problem Description

The original implementation allowed the following patterns:

- `std::basic_string_view<T>` where `T` is not a character type
- `std::basic_string<T>` where `T` is not a character type

This causes issues on libc++ because:

- `std::char_traits<T>` is undefined for non-character types
- these errors occur at template instantiation time, not at overload resolution time
- as a result, valid uses such as `fnv<uint32_t>` fail to compile

## How Has This Been Tested?

    - Built and tested on FreeBSD (libc++, clang) via the official FreeBSD ports framework.
    - Verified that the project builds successfully without compilation errors.
    - No functional behavior changes are introduced.

## Checklist:

    - No documentation changes are needed.
    - No functional changes intended. So no tests are added.

### DIFF
--- a/src/crispy/FNV.h
+++ b/src/crispy/FNV.h
@@ -48,14 +48,10 @@ class fnv
     /// which are handled by dedicated overloads.
     template <typename V>
     constexpr U operator()(U memory, V const& value) const noexcept
-    requires(
-        std::is_trivially_copyable_v<V> &&
-        !std::same_as<V, std::string> &&
-        !std::same_as<V, std::string_view>
-    )
+        requires(std::is_trivially_copyable_v<V> && !std::same_as<V, std::string>
+                 && !std::same_as<V, std::string_view>)
     {
-        auto const* bytes =
-            reinterpret_cast<unsigned char const*>(&value);
+        auto const* bytes = reinterpret_cast<unsigned char const*>(&value);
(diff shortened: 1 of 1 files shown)


Yaraslaut:
Hi @tagattie, instead of using `enable_if` hack, please use concepts. See example on how to do it in this case: https://godbolt.org/z/va6vrs9x8

PR #1867, opened 2026-02-02
Fix initial terminal window size at fractional DPR

### LINKED ISSUE #1858: Initial window size incorrect with Wayland fractional scaling (150%)
### Contour Terminal version

0.6.1

### Installer source

GitHub: Source code

### Operating System

Aurora OS

### Architecture

x86-64 / amd64

### Other Software

  - KDE 6.4.5
  - OS: Fedora 43 (Linux 6.17.11)                                                                             
  - Display: 3840x2160 @ 150% scaling                                                                         
  - Session: Wayland                                                                                          
  - Contour version: (run contour version to get

### Steps to reproduce

1. Loginto KDE using wayland
2. Configure a display at 3840x2160 with 150% fractional scaling
3. Set terminal_size.columns: 80 in contour.yml                                                             
4. Launch Contour on Wayland                                                                                
5. Observe the actual column count (visible when resizing, or via tput cols)   

### Expected Behavior

tput cols => 80 columns

### Actual Behavior

tput cols => 63

### Additional notes

  - With platform_plugin: xcb (XWayland): configured 80 → actual 85 columns                                   
  - Resizing the window after launch works correctly and shows accurate dimensions
  - Workaround: setting columns: 105 yields ~80 actual columns on Wayland 

  This suggests the initial window geometry calculation doesn't correctly account for fractional scaling on   
  either backend, but the error is more severe on native Wayland.

christianparpart:
Closes #1858.

- Fixes the terminal spawning at the wrong size (e.g. 127×29 instead of configured 130×30) when using fractional device pixel ratios such as DPR 1.5 (150% KDE/Wayland scaling)
- Three-layer fix addressing: rounding errors in implicit size computation, stale dimensions during `createRenderer()`, and Wayland compositor configure event reversions
- No performance impact; correction logic is time-limited to 500ms after renderer creation

**Layer 1 — `updateImplicitSize()` rounding fix:** Compute required size in actual pixels using exact integer arithmetic, then convert to virtual pixels with a single `ceil` division. This avoids compounding per-cell rounding errors at fractional DPR.

**Layer 2 — `createRenderer()` stale dimensions fix:** Call `applyFontDPI()` to catch DPR corrections between `setSession()` and first render; use `implicitWidth()`/`implicitHeight()` instead of stale `width()`/`height()` during `beforeSynchronizing`; defer `configureDisplay()` via `post()` to avoid Wayland configure events using uncommitted geometry.

**Layer 3 — Wayland configure reversion correction:** Detect compositor configure reversions in `sizeChanged()` by checking if BOTH dimensions mismatch the implicit size within 500ms of `createRenderer()`. When detected, post a corrective `window()->resize()` and return early to prevent applying the wrong terminal size.

### DIFF
--- a/src/contour/display/TerminalDisplay.cpp
+++ b/src/contour/display/TerminalDisplay.cpp
@@ -372,8 +372,9 @@ void TerminalDisplay::sizeChanged()
                      implicitWidth(),
                      implicitHeight());
         post([this]() {
-            window()->resize(static_cast<int>(std::ceil(implicitWidth())),
-                             static_cast<int>(std::ceil(implicitHeight())));
+            if (auto* currentWindow = window(); currentWindow)
+                currentWindow->resize(static_cast<int>(std::ceil(implicitWidth())),
+                                      static_cast<int>(std::ceil(implicitHeight())));
         });
         return;
     }
@@ -1125,8 +1126,8 @@ void TerminalDisplay::updateImplicitSize()
     // that would otherwise cause extra pixels and wrong column/line counts at fractional DPR.
     auto const actualRequiredSize = computeRequiredSize(scaledMargins, actualGridCellSize, totalPageSize);
 
-    auto const virtualWidth = std::ceil(static_cast<qreal>(unbox(actualRequiredSize.width)) / dpr);
-    auto const virtualHeight = std::ceil(static_cast<qreal>(unbox(actualRequiredSize.height)) / dpr);
+    auto const virtualWidth = std::ceil(static_cast<double>(unbox(actualRequiredSize.width)) / dpr);
+    auto const virtualHeight = std::ceil(static_cast<double>(unbox(actualRequiredSize.height)) / dpr);
 
     displayLog()("Implicit display size set to {}x{} (actualRequired: {}, cellSize: {}, contentScale: {}, "
                  "pageSize: {})",


src/contour/display/TerminalDisplay.cpp
```diff
@@ -1020,26 +1071,31 @@ void TerminalDisplay::updateImplicitSize()
...
-                 virtualCellSize,
+    auto const scaledMargins = applyContentScale(_session->profile().margins.value(), dpr);
+
+    // Compute the required size in actual pixels using exact integer arithmetic,
+    // then convert to virtual pixels. This avoids compounding ceil-per-cell rounding errors
+    // that would otherwise cause extra pixels and wrong column/line counts at fractional DPR.
+    auto const actualRequiredSize = computeRequiredSize(scaledMargins, actualGridCellSize, totalPageSize);
+
+    auto const virtualWidth = std::ceil(static_cast<qreal>(unbox(actualRequiredSize.width)) / dpr);
+    auto const virtualHeight = std::ceil(static_cast<qreal>(unbox(actualRequiredSize.height)) / dpr);
```
Yaraslaut: why `qreal` ? use float or doule

christianparpart: that's what Qt is using, but i'm fine in using `float` directly :)

PR #1883, opened 2026-02-17
Fix hint mode highlight offset with non-ASCII characters

### LINKED ISSUE #1870: Highlight of the hint is offset in the input prompt
For the hint something is off when the text inside the input prompt 

Image

christianparpart:
Fixes #1870

- Converts regex byte offsets to codepoint indices when computing hint overlay column positions, fixing rightward shift on lines with multi-byte UTF-8 characters before a match (#1870)
- Adds 5 test cases covering Unicode prompts, CJK chars, multiple non-ASCII segments, ASCII regression, and match-at-line-start

### DIFF
--- a/src/vtbackend/HintModeHandler.cpp
+++ b/src/vtbackend/HintModeHandler.cpp
@@ -9,6 +9,28 @@ using namespace std;
 namespace vtbackend
 {
 
+/// Converts a UTF-8 byte offset within @p text to the corresponding codepoint index.
+///
+/// In the UTF-8 strings produced by Line::toUtf8(), each grid cell emits exactly one
+/// codepoint (wide-character continuation cells emit a space). Therefore the codepoint
+/// index equals the column index for these strings.
+///
+/// @note This function counts UTF-8 leading bytes and does not perform grapheme-cluster
+///       segmentation. The returned index equals a terminal grid column only when the
+///       input was produced by Line::toUtf8(), which guarantees one codepoint per cell.
+///
+/// @param text       The UTF-8 encoded string.
+/// @param byteOffset The byte offset to convert.
+/// @return The number of codepoints before @p byteOffset.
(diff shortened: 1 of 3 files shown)


src/vtbackend/HintModeHandler.cpp
```diff
@@ -9,6 +9,23 @@ using namespace std;
...
+///
+/// @param text       The UTF-8 encoded string.
+/// @param byteOffset The byte offset to convert.
+/// @return The number of codepoints before @p byteOffset.
+static constexpr auto utf8ByteOffsetToCodepointIndex(string_view text, size_t byteOffset) noexcept -> size_t
+{
+    auto const limit = min(byteOffset, text.size());
+    // Count bytes that are NOT continuation bytes (10xxxxxx).
+    return static_cast<size_t>(ranges::count_if(views::iota(size_t { 0 }, limit), [&](auto i) {
+        return (static_cast<unsigned char>(text[i]) & 0xC0) != 0x80;
```
Yaraslaut: not a big fun of `unisgned char` can we use `char8_t` here?

src/vtbackend/HintModeHandler.cpp
```diff
@@ -43,8 +60,10 @@ void HintModeHandler::rescanLines(vector<string> const& visibleLines, PageSize p
                 if (pattern.validator && !pattern.validator(matchStr))
                     continue;
 
-                auto const startCol = ColumnOffset(static_cast<int>(match.position()));
-                auto const endCol = ColumnOffset(static_cast<int>(match.position() + match.length() - 1));
+                auto const startCol =
+                    ColumnOffset::cast_from(utf8ByteOffsetToCodepointIndex(lineText, match.position()));
+                auto const endCol = ColumnOffset::cast_from(
+                    utf8ByteOffsetToCodepointIndex(lineText, match.position() + match.length()) - 1);
```
Yaraslaut: this -1 is kinda hiding there, please move `utf8ByteOffsetToCodepointIndex(lineText, match.position() + match.length())` into separate variable

PR #1885, opened 2026-02-17
Fix Kitty keyboard protocol encoding and F3 key handling

christianparpart:
## Summary
- Fix Kitty keyboard protocol encoding to match the reference implementation, ensuring correct CSI u sequences for modifier keys and special keys
- Fix F3 search navigation keybindings to only activate in Search mode, preventing them from swallowing F3 in normal terminal usage
- Fix F3 key not reaching the PTY when a keybinding action doesn't execute (e.g., mode guard fails)


christianparpart:
> lgtm, i guess :)

"i guess" ....

PR #1900, opened 2026-02-23
Add Semantic Block Protocol (DEC Mode 2034) with SBQUERY

christianparpart:
- Introduce DEC Private Mode 2034 to enable semantic block tracking via OSC 133 shell integration sequences, storing command metadata (command line, exit code, prompt/output text) alongside line flags (`OutputStart`, `CommandEnd`) in the grid ring buffer.
- Add the SBQUERY sequence (`CSI > Ps ; Pn b`) that returns structured JSON via DCS responses, supporting three query types: last completed command (`Ps=1`), last N completed commands (`Ps=2`), and current in-progress command (`Ps=3`). Query type values use named `SBQueryType` constants to avoid magic numbers.
- Include `SemanticBlockTracker` class, full test coverage (8 test cases / 27 assertions), and a protocol specification document at `docs/vt-extensions/semantic-block-query.md`.

@Yaraslaut **Background story:** I need that for endo to be able to let the shell agent read the output of the last executed command.

### DIFF
--- a/docs/vt-extensions/semantic-block-query.md
+++ b/docs/vt-extensions/semantic-block-query.md
@@ -1,0 +1,211 @@
+# Semantic Block Query (DEC Mode 2034)
+
+This extension provides a discoverable, machine-readable query mechanism that returns structured
+JSON blocks of semantic command data from the terminal. It is designed for programmatic consumers
+such as AI agents, accessibility tools, and automation scripts.
+
+## Prerequisites
+
+This feature requires [OSC 133 shell integration](osc-133-shell-integration.md) to be active
+in the running shell. The shell must emit OSC 133 sequences (A, B, C, D) to mark prompt boundaries,
+command output regions, and command completion.
+
+## Feature Discovery
+
+### DEC Private Mode 2034
+
+```
+Enable:  CSI ? 2034 h       (DECSM)
+Disable: CSI ? 2034 l       (DECRM)
+Query:   CSI ? 2034 $ p     (DECRQM)  ->  CSI ? 2034 ; Ps $ y
+```
+
+Where `Ps` in the DECRQM response is:
+- `0` = not recognized (terminal does not support this mode)
+- `1` = set (enabled)
+- `2` = reset (disabled)
+
+When enabled:
+- The terminal tracks semantic zones from OSC 133 sequences using line flags and internal metadata.
+- The terminal generates a session token and replies with a DCS containing the token (see Authentication below).
+- The query sequence `CSI > Ps ; Pn ; T1 ; T2 ; T3 ; T4 b` becomes available.
+
+When disabled:
+- All tracked semantic block data is discarded.
+- The session token is invalidated.
+- The query sequence returns an error response.
+
+## Authentication
+
+When mode 2034 is enabled via DECSM, the terminal generates a 64-bit session token and replies with:
+
+```
+DCS > 2034 ; 1 b T1 ; T2 ; T3 ; T4 ST
+```
+
+That is: `ESC P > 2034 ; 1 b T1 ; T2 ; T3 ; T4 ESC \`
+
+The 4 integer values (`T1`..`T4`) are the session token. These same values must be included
+in every SBQUERY request as additional CSI parameters (see Query Syntax below). No conversion
+is needed — the values from the DCS reply are used directly.
+
+### Token Lifecycle
+
+1. **On DECSET 2034**: Terminal generates a fresh token and replies with the DCS token response.
+2. **On SBQUERY**: The token must be included as parameters T1..T4. Missing or invalid tokens produce error responses.
+3. **On DECRST 2034**: The token is invalidated along with all tracked data.
+4. **On re-enable**: A new token is generated; previous tokens are no longer valid.
+
+### Security Note
+
+DECSET normally does not produce a reply. Some terminal multiplexers (e.g. tmux, screen)
+may not forward the DCS token response to the application. In such environments, token-based
+authentication may not be usable. This is a known limitation.
+
+## Query Syntax — SBQUERY
+
+```
+CSI > Ps ; Pn ; T1 ; T2 ; T3 ; T4 b
+```
+
+**Mnemonic:** SBQUERY — **S**emantic **B**lock **QUERY**
+
+CSI with `>` leader, final character `b` (for "blocks").
+
+| Parameter | Meaning                                              |
+|-----------|------------------------------------------------------|
+| Ps        | Query type (see table below)                         |
+| Pn        | Count parameter (default 1)                          |
+| T1..T4    | Session token as 4 × uint16 values (from DECSET reply) |
+
+| Ps | Meaning                         | Pn                  |
+|----|-------------------------------- |---------------------|
+| 1  | Last completed command block    | ignored             |
+| 2  | Last N completed command blocks | count (default 1)   |
+| 3  | Current in-progress command     | ignored             |
+
+## Response Syntax
+
+The response is a DCS (Device Control String) with `>` leader and final character `b`.
+
+**Success:**
+
+```
+DCS > 1 b {JSON} ST
+```
+
+That is: `ESC P > 1 b {JSON} ESC \`
+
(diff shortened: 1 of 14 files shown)


docs/vt-extensions/semantic-block-query.md
```diff
@@ -0,0 +1,145 @@
...
+
+CSI with `>` leader, final character `b` (for "blocks").
+
+| Ps | Meaning                         | Pn                  |
+|----|-------------------------------- |---------------------|
+| 1  | Last completed command block    | ignored             |
+| 2  | Last N completed command blocks | count (default 1)   |
+| 3  | Current in-progress command     | ignored             |
+
+## Response Syntax
```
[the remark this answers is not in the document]
Yaraslaut: valid point though, make sure that only the output of current program is visible to the program itself, so have some sort of session identifier, and that you need to provide a secret that terminal gives you in order to fetch the commands output that only follow the secret acquisition

src/vtbackend/Screen.cpp
```diff
@@ -3458,6 +3458,328 @@ void Screen<Cell>::processSequence(Sequence const& seq)
...
+    }
+} // namespace
+
+template <CellConcept Cell>
+void Screen<Cell>::handleSemanticBlockQuery(Sequence const& seq)
+{
+    // If mode 2034 is disabled, reply with error DCS.
+    if (!_terminal->isModeEnabled(DECMode::SemanticBlockProtocol))
+    {
+        reply("\033P>0b\033\\");
```
Yaraslaut: we repeat this line several times, also better to name it, so that it is obvious what it is

src/vtbackend/Screen.cpp
```diff
@@ -3458,6 +3458,328 @@ void Screen<Cell>::processSequence(Sequence const& seq)
...
+        if (!currentBlock || currentBlock->finished)
+        {
+            reply("\033P>0b\033\\");
+            return;
+        }
+
+        // Scan grid for in-progress output: find most recent OutputStart without CommandEnd.
+        auto outputText = std::string {};
+        auto outputLineCount = 0;
+        auto foundOutputStart = false;
```
Yaraslaut: Use lambda to return `auto const [foundOutputStart, outputStartLine] = [&]() { .... }();`
instead of creating mutable variables that are modified inside the loop, and similar for `outputText` and `outputLineCount`

src/vtbackend/Screen.cpp
```diff
@@ -3458,6 +3458,328 @@ void Screen<Cell>::processSequence(Sequence const& seq)
...
+        if (foundOutputStart)
+        {
+            // Collect text from OutputStart to cursor.
+            for (auto line = outputStartLine; line <= cursorLine; ++line)
+            {
+                if (line != outputStartLine)
+                    outputText += '\n';
+                outputText += _grid.lineAt(line).toUtf8Trimmed(false, true);
+                ++outputLineCount;
+            }
```
Yaraslaut: this is `std::ranges::fold_left( _grid.begin(), std::advance(_grid.begin(), outputStartLine), []...)`
if we can not iterate over grid, we can iterate over iota (0, outputStartLine)

src/vtbackend/Screen.cpp
```diff
@@ -3458,6 +3458,328 @@ void Screen<Cell>::processSequence(Sequence const& seq)
...
+                {
+                    for (auto pl = line; pl < outputStartLine; ++pl)
+                    {
+                        if (pl != line)
+                            promptText += '\n';
+                        promptText += _grid.lineAt(pl).toUtf8Trimmed(false, true);
+                    }
+                    break;
+                }
+            }
```
Yaraslaut: no raw loops here as well, i think using algorithms shows better what is happening

src/vtbackend/Screen.cpp
```diff
@@ -3458,6 +3458,328 @@ void Screen<Cell>::processSequence(Sequence const& seq)
...
+        json += ",\"prompt\":";
+        json += jsonEscape(promptText);
+        json += ",\"output\":";
+        json += jsonEscape(outputText);
+        json += ",\"exitCode\":-1";
+        json += ",\"finished\":false";
+        json += std::format(",\"outputLineCount\":{}", outputLineCount);
+        json += "}]}";
+        reply("\033P>1b{}\033\\", json);
+        return;
```
Yaraslaut: please use std::format

src/vtbackend/Screen.cpp
```diff
@@ -3458,6 +3458,328 @@ void Screen<Cell>::processSequence(Sequence const& seq)
...
+        blocks.push_back(&*it);
+
+    if (blocks.empty())
+    {
+        reply("\033P>0b\033\\");
+        return;
+    }
+
+    // Scan grid backward to find text regions for each block.
+    // We match blocks from the tracker with line-flag boundaries in the grid.
```
Yaraslaut: this feels like another procedure, separate this into it own function

src/vtbackend/Screen.cpp
```diff
@@ -3458,6 +3458,328 @@ void Screen<Cell>::processSequence(Sequence const& seq)
...
+    auto currentPrompt = std::string {};
+    auto currentOutputLineCount = 0;
+    auto outputStartLine = cursorLine;
+
+    for (auto line = cursorLine; line >= historyTop && blockIndex < static_cast<int>(blocks.size()); --line)
+    {
+        auto const& gridLine = _grid.lineAt(line);
+        auto const flags = gridLine.flags();
+
+        switch (state)
```
Yaraslaut, line 3761: I think that the switch statement is too big, separate every switch it the lambda with clear responsability and then call them inside 
```
switch(state):
   case(searching:)
        actionOnSearchState();
        break;
.....
```

src/vtbackend/Screen.cpp
```diff
@@ -3458,6 +3458,328 @@ void Screen<Cell>::processSequence(Sequence const& seq)
...
+        json += ",\"prompt\":";
+        json += jsonEscape(blockTexts[i].prompt);
+        json += ",\"output\":";
+        json += jsonEscape(blockTexts[i].output);
+        json += std::format(",\"exitCode\":{}", block->exitCode);
+        json += ",\"finished\":true";
+        json += std::format(",\"outputLineCount\":{}", blockTexts[i].outputLineCount);
+        json += '}';
+    }
+    json += "]}";
```
Yaraslaut, line 3787: std::format and std::format_to

src/vtbackend/SemanticBlockTracker.cpp
```diff
@@ -0,0 +1,76 @@
...
+    // If there's a current block that was finished, push it to completed.
+    if (_currentBlock && _currentBlock->finished)
+    {
+        _completedBlocks.push_back(std::move(*_currentBlock));
+        if (_completedBlocks.size() > _maxBlocks)
+            _completedBlocks.pop_front();
+    }
+
+    // Start tracking a new block.
+    _currentBlock = CommandBlockInfo {};
```
Yaraslaut: slightly confused, do we drop currentBlock on the start if it is not finished?

src/vtbackend/Screen.cpp
christianparpart: oopsi thx :)

src/vtbackend/SemanticBlockTracker.cpp
christianparpart: Yes, that's intentional. If a command is interrupted (e.g. Ctrl+C before the shell emits `OSC 133;D`), the block never gets `finished = true`, so it is discarded when the next prompt starts. Only commands that ran to completion have reliable metadata (exit code, output boundaries), so promoting incomplete blocks into `completedBlocks` would produce misleading query results.

I've added a comment in the code to make this explicit:

```cpp
// Unfinished blocks (e.g. from Ctrl+C before OSC 133;D) are intentionally
// discarded — only commands that ran to completion have reliable metadata.
```

Note that the in-progress block is still queryable via `SBQUERY` with `Ps=3` (query type `InProgress`) *while* it's running — it just doesn't survive into the completed history if it never finishes.

PR #1903, opened 2026-02-26
Add HUD (Heads-Up Display) overlay screen via DEC mode 2035

christianparpart:
## Summary

- Introduces a transparent overlay screen (DEC mode 2035) that composites on top of the primary screen, enabling TUI applications to render popup dialogs, menus, and floating windows without destroying underlying content
- When enabled, a clean overlay is activated and all VT output goes to it; non-empty HUD cells overlay primary cells in an O(N) linear merge during rendering; when disabled, the overlay is discarded and the primary screen is fully intact
- Includes protocol specification (`docs/vt-extensions/hud-screen.md`), 12 E2E test cases (31 assertions), an interactive demo program (`examples/hud-screen-demo.cpp`), and release note entry

### DIFF
--- a/docs/vt-extensions/hud-screen.md
+++ b/docs/vt-extensions/hud-screen.md
@@ -1,0 +1,65 @@
+# HUD Screen
+
+Terminal applications sometimes need to display popup-like overlays (menus, tooltips, floating windows)
+on top of existing screen content. The HUD (Heads-Up Display) screen provides a transparent overlay layer
+that composites on top of the primary screen without modifying its contents.
+
+This extension is controlled via a DEC mode (`2035`).
+
+## Semantics
+
+When HUD mode is enabled:
+
+- A clean overlay screen is created with the same dimensions as the primary screen.
+- All subsequent VT output is directed to the HUD screen instead of the primary screen.
+- During rendering, the primary screen is drawn first, then HUD screen cells with content
+  are composited on top. Empty (unwritten) HUD cells are transparent, showing the primary
+  screen through.
+- The cursor position is governed by the HUD screen.
+
+When HUD mode is disabled:
+
+- The HUD overlay is discarded entirely.
+- The primary screen becomes the active screen again, fully intact.
+- No cleanup of the HUD content is needed — disabling the mode suffices.
+
+Each time the mode is enabled, the HUD screen starts in a clean (freshly reset) state.
+
+The HUD mode only operates on top of the primary screen. Enabling it while the alternate
+screen is active has no effect. If the alternate screen is activated while HUD is active,
+the HUD is automatically disabled first.
+
+## Feature detection
+
+Use `DECRQM` (`CSI ? 2035 $ p`) to detect support for the HUD screen mode.
+
+## Using the feature
+
+Use `CSI ? 2035 h` to enable the HUD screen overlay.
+
+Use `CSI ? 2035 l` to disable the HUD screen overlay.
+
+## Example
+
+```
+# Write primary screen content
+printf "Hello, World!\n"
+printf "Background text.\n"
+
+# Enable HUD and draw a popup box
+printf "\033[?2035h"
+printf "\033[5;10H┌──────────┐"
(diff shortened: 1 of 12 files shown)


docs/vt-extensions/hud-screen.md
Yaraslaut, line 1: we also have this: https://github.com/contour-terminal/vt-extensions/

src/vtbackend/Terminal.cpp
```diff
@@ -537,6 +548,39 @@ void Terminal::fillRenderBufferInternal(RenderBuffer& output, bool includeSelect
...
+                                                           InputMethodData {},
+                                                           theCursorPosition,
+                                                           includeSelection },
+                      ScrollOffset(0),
+                      HighlightSearchMatches::Yes);
+
+    // Both renders produce cells in identical position order (same page size,
+    // same cell-by-cell mode). Linear merge: overlay non-empty HUD cells.
+    auto const count = std::min(output.cells.size(), hudBuffer.cells.size());
+    for (size_t i = 0; i < count; ++i)
```
Yaraslaut: cmn, this is iota

metainfo.xml
```diff
@@ -91,19 +91,14 @@
...
   <releases>
 
-    <!-- unreleased: TODO: Don't forget to add date attribute to it. -->
-    <!--
-
     <release version="x.y.z" urgency="medium" type="development">
       <description>
         <ul>
-          <li> ... </li>
+          <li>Adds HUD (Heads-Up Display) overlay screen (DEC mode 2035) for transparent popup-like overlays on top of the primary screen without destroying underlying content</li>
```
Yaraslaut: wrong place

src/vtbackend/Terminal.h
```diff
@@ -1353,6 +1364,9 @@ class Terminal
     // {{{ Displays this terminal manages
     Screen<PrimaryScreenCell> _primaryScreen;
     Screen<AlternateScreenCell> _alternateScreen;
+    Screen<HudScreenCell> _hudScreen;
+    bool _hudActive = false;
+    ScreenBase* _preHudScreen = nullptr;
```
Yaraslaut: i think that when we have such variable that are always synchronized, better to encapsulate it and keep synchronization as a class invariant, in something like `HudState`

Yaraslaut (changes requested):
Great idea, also great that it is so easy to introduce this in the codebase :) 
I left few comments, the most important one is about the repo with vt-extensions, i think we want to keep it up-to date and maybe other people might be interested in implementing and using this extension

j4james:
> enabling TUI applications to render popup dialogs, menus, and floating windows without destroying underlying content

Note that terminals can already do this with standard ANSI/DEC sequences. You can save the foreground page to a backup buffer with [`DECCRA`](https://vt100.net/docs/vt510-rm/DECCRA.html), and then when you want to close your popup window, you just copy that content back again.

This requires the terminal to support multiple pages, but several terminals already do. I think Contour's `DA1` report actually indicates that it supports multiple pages, although I'm almost certain that's not true (would be nice if it was though).

j4james:
> I mean, when we have mutliple pages, which I interpret as just multiple grid buffers, why do we have DEC mode for switching to the (one and only) alternate screen buffer? 🤔

The alternate buffer is not a DEC mode - it's an Xterm thing that is unrelated. The concept of multiple pages was defined in the original ANSI standard. Other than the operations you've already found, you'll also need the DEC sequence [`DECPCCM`](https://vt100.net/docs/vt510-rm/DECPCCM.html) (this is important for controlling which page is visible when moving between pages), and [`DECRQDE`](https://vt100.net/docs/vt510-rm/DECRQDE.html) is also somewhat relevant.

> How many pages would we have?

Conformance to Level 4 (which Contour claims) requires supports for at least three pages. The VT420 had up to six (the exact number depended on the page size, and how many sessions were active), and the VT525 had up to nine. I think some modern terminals probably have much more.

PR #1904, opened 2026-02-27
Fix initial resizing with tiling WM

Yaraslaut:
For tiling WM, resize of the window is intentional during startup, so we need to make sure that we resize in this case, and only do not proceed with resize due to DPR corrections 
can you please check that this fix works for you @christianparpart 

Closes: https://github.com/contour-terminal/contour/issues/1908

### DIFF
--- a/cmake/presets/os-linux.json
+++ b/cmake/presets/os-linux.json
@@ -53,7 +53,8 @@
                 "rhs": "Linux"
             },
             "cacheVariables": {
-                "CMAKE_CXX_COMPILER": "clang++"
+                "CMAKE_CXX_COMPILER": "clang++",
+                "CMAKE_C_COMPILER": "clang"
             }
         },
         {
@@ -71,6 +72,7 @@
             },
             "cacheVariables": {
                 "CMAKE_CXX_COMPILER": "clang++",
+                "CMAKE_C_COMPILER": "clang",
                 "CMAKE_INSTALL_PREFIX": "/usr/local"
             }
         }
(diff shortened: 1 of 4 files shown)


cmake/presets/os-linux.json
```diff
@@ -71,6 +72,7 @@
             },
             "cacheVariables": {
                 "CMAKE_CXX_COMPILER": "clang++",
+                "CMAKE_C_COMPILER": "clang",
```
christianparpart, line 75: I do not see why you need to add `CMAKE_C_COMPILER` here and above. We only use C++ files exclusively. It would just result in a cmake warning that `CMAKE_C_COMPILER` is unused.

src/contour/display/TerminalDisplay.h
```diff
@@ -262,6 +262,8 @@ class TerminalDisplay: public QQuickItem
     TerminalSession* _session = nullptr;
     std::chrono::steady_clock::time_point _startTime;
     std::chrono::steady_clock::time_point _initialResizeDeadline {};
+    double _staleVirtualWidth {};
+    double _staleVirtualHeight {};
```
christianparpart: I'm not sure the word `stale` is the best way to reflect what this value stores. Probably `lastVirtualWidth` / `...Height` would be more self explanory.

cmake/presets/os-linux.json
Yaraslaut: this is for the wayland-protocol bluring, this is some cmake qt handling, and it injects wrong compiler commands in the c++ compiler based on the c compiler

christianparpart (approved):

PR #1905, opened 2026-03-01
Add DEC VT420 multi-page support with 16 pages

christianparpart:
Replaces #1903.

- Implements the DEC VT420 multi-page specification with 16 independent screen pages (pages 1-15 for DEC navigation, page 16 reserved as the Xterm alternate screen buffer via DECSET 47/1049)
- Adds VT sequences `NP`, `PP`, `PPA`, `PPR`, `PPB` for page navigation, `DECRQDE` for display extent reporting, `DECPCCM` (mode 64) for page-cursor coupling, cross-page `DECCRA` support, and fixes `DECXCPR` to report the correct page number
- Adds also a small demo for a standalone CLI overlay demo using DECCRA and page switching to simulate HUD style rendering.

### DIFF
--- a/src/vtbackend/Screen.cpp
+++ b/src/vtbackend/Screen.cpp
@@ -338,17 +338,22 @@ void Screen<Cell>::applyPageSizeToMainDisplay(PageSize mainDisplayPageSize)
 {
     auto cursorPosition = _cursor.position;
 
+    // Only reset margins when the page size actually changes (during a resize),
+    // not during a simple page switch where the grid size already matches.
+    auto const sizeChanged = (_grid.pageSize() != mainDisplayPageSize);
+
     // Ensure correct screen buffer size for the buffer we've just switched to.
     cursorPosition = _grid.resize(mainDisplayPageSize, cursorPosition, _cursor.wrapPending);
     cursorPosition = clampCoordinate(cursorPosition);
 
-    auto const margin = Margin {
-        .vertical = Margin::Vertical { .from = {}, .to = mainDisplayPageSize.lines.as<LineOffset>() - 1 },
-        .horizontal =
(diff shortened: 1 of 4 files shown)


christianparpart:
@j4james with this, I can happily close the PR #1903. I've ported my example program that also proves that it is indeed possible what I was wanted to achieve.

j4james:
> Pages 1-15 for DEC navigation, page 16 reserved as the Xterm alternate screen buffer via DECSET 47/1049

I like this idea. In Windows Terminal we just disable the paging functionality when the alternate screen buffer is active, because I couldn't think of a good way to integrate them, but this approach seems like it could work well.

Which reminds me that there is another mode worth looking at: `DECRPL` (Review Previous Lines). On the VT520 it was just used to enable/disable the scrollback buffer, but I think on modern terminals it makes sense to handle it similar to the alternate screen mode, i.e. when disabled it doesn't just disable the scrollback, but also disables the resize reflow behavior.

Because if we had that, then `DECSET 1049` is essentially the equivalent of:
1. `DECSC` to save the cursor state
2. `PPA 16` to move to the alternate screen buffer page
3. Reset `DECRPL` to disable the scrollback and reflow.
4. `ED 2` to clear the screen.
5. Toggle `DECPCCM` (if it isn't already set) to ensure page 16 is visible.

Technically `DECSET 1049` has some other semantic differences too, but I think the steps above probably cover most of it.

My only concern is that if other terminals adopt your approach, the alternate screen buffer page is almost certainly going to be terminal dependent. And if an app wants to switch to another page from within the alternate screen buffer, then switch back again, they'd need to know what page to use. But I suppose that's easy enough to query, or maybe they could just enable `DECSET 1049` again (if it's already set, but you're not on the correct page, it just moves you there).

christianparpart:
Thanks for the link, @j4james. BTW, I've incorporated your feedback into the PR and also added my HUD example from the other (already closed PR) over to this DEC multi page PR to demo it.

Issue #408, opened 2021-09-13
sixvid causes contour grief


hackerb9:
## Description

[Sixvid](https://github.com/hackerb9/sixvid) is an extremely simple GIF viewer for sixel terminals. For some reason Contour seems to be having trouble with it. 

## Environment

- Contour Version or Git commit hash: github main 77cb600f52f94d888cafe5282a64ad27d0494137
- Operating System (name / distribution / version): Debian GNU/Linux (bullseye)
- Contour configuration: unchanged
- TERM environment variable: unchanged
- Compiler version: g++ 10

## Steps to Reproduce

1. git clone https://github.com/hackerb9/sixvid
2. cd sixvid
3. ./sixvid -b nyantocat.gif

Note that the -b option tells sixvid to run in "benchmark" mode, so it will loop the video as fast as possible.

## Odd effects occur

1. This test requires a terminal with enough space for a 500x500 image plus two lines of text. On XTerm with a 10x20 pixel font, dragging the terminal out to be 27 rows tall is sufficient. 
a. On Contour, that doesn't work. The sixel image is scaled and the screen scrolls on every frame.
b. The scaling of the image is very peculiar. For example, I have portrait monitor (taller than wide) and the video appeared smooshed. 
<details><summary>Smooshed nyantocat</summary>

![image](https://user-images.githubusercontent.com/10174223/133026256-350987ad-a989-405b-a75e-b445f50cc905.png)
</details>

2. Trying to resize the font using Ctrl+mousewheel triggered two bugs.
  a. When I zoomed in, the terminal started using up 100% CPU (on one core) and the frame rate dropped to 1% of its previous level. Eventually Contour was so unresponsive I had to kill the process. 
  b. I tried zooming out. This time I quit sixvid before zooming just in case that had caused a problem. At first the text and image were getting smaller, but then Contour crashed trying to free an invalid pointer. 
<details><summary>(gdb) bt</summary>

```gdb
free(): invalid pointer

Thread 1 "contour" received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50      ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff649e537 in __GI_abort () at abort.c:79
[...]

PR #1922, opened 2026-03-28
Bugfix segfaults when changing font/window size

### LINKED ISSUE #1712: Crash on Windows when agressively resizing.
### Contour Terminal version

Contour 0.6.1.7494

### Installer source

GitHub: release page

### Operating System

Windows 11

### Architecture

x86-64

### Other Software

_No response_

### Steps to reproduce

Start contour, start WSL, start resizing and moving it around with Win+arrows. 

### Expected Behavior

Resizing happens.

### Actual Behavior

Contour will start "not responding" with these errors:

```
PS C:\Users\bferdinandy> [error] Ignoring TERM environment variable for child process.
[error] freetype: Failed to set LCD filter. unimplemented feature
Warning: The cached device pixel ratio value was stale on window expose.  Please file a QTBUG which explains how to reproduce. ((null):0, (null))
Warning: The cached device pixel ratio value was stale on window expose.  Please file a QTBUG which explains how to reproduce. ((null):0, (null))
Warning: The cached device pixel ratio value was stale on window expose.  Please file a QTBUG which explains how to reproduce. ((null):0, (null))
Warning: The cached device pixel ratio value was stale on window expose.  Please file a QTBUG which explains how to reproduce. ((null):0, (null))
```

In the end you need to kill it.

### Additional notes

_No response_

jquast:
Problem
-------

This problem is reproduced by using ctrl+mouse wheel, and crashes more frequently when "very busy" with the render thread, it is a race condition.  I had it crash with very casual use by adjusting the font size a few notches while using [gambaterm](https://github.com/vxgmichel/gambatte-terminal), which has 60Hz full screen redraws with the DEC synchronized output sequence, but it can be reproduced with a little more effort (resizing with *very large* point sizes) by running ``find /`` command, as captured in this recording:

https://github.com/user-attachments/assets/b93f8625-57b3-43ec-97bf-1ffab549f87e

I would say this crash is a little rare, that a majority of users should not experience it.

Analysis
--------

A font change triggers ``updateFontMetrics()`` (UI thread), which rebuilds grid metrics and replaces the texture atlas while the QSG render thread is actively using them in ``renderImpl()`` (render thread).

The resulting window resize has the same problem, with ``applyResize()`` updating page size and margins without mutex. About four different segfaults were possible due to a race condition that can get null atlas dereferences, corrupted line buffers, or assertion failures on atlas dimensions.

Additionally, another race in ``setGuiTabInfoForStatusLine()`` does an unprotected ``std::move`` of a ``TabsInfo`` struct (from UI thread) while the terminal thread copies it under ``_stateMutex``, causing heap corruption.

Solution
--------

Added ``_renderMutex`` to Renderer, held by all geometry/atlas writers and ``renderImpl()``.  The ``TabsInfo`` fix just acquires ``_stateMutex`` on the writer side to match the reader.

(Probably) closes #1712 and #408

## Checklist

no documentation or tests were added or changed, this would be very difficult to create a test for, it would require preparing and running all of these threads with real Qt / OpenGL contexts, I'm afraid I can't do that.

### DIFF
--- a/metainfo.xml
+++ b/metainfo.xml
@@ -160,6 +160,7 @@
           <li>Fixes build failure on Alpine Linux (musl libc) due to missing close_range() function (#1879)</li>
           <li>Fixes third-party dependency files (headers, libraries) leaking into install tree (#1788)</li>
           <li>Fixes size of the window when using tiling WM (#1908)</li>
+          <li>Fixes crash when rapidly changing font size due to data races between UI and render threads (#1922)</li>
         </ul>
       </description>
     </release>
--- a/src/contour/helper.cpp
+++ b/src/contour/helper.cpp
@@ -781,14 +781,12 @@ void applyResize(vtbackend::ImageSize newPixelSize,
     vtbackend::Terminal& terminal = session.terminal();
     vtbackend::ImageSize const cellSize = renderer.gridMetrics().cellSize;
 
-    if (renderer.hasRenderTarget())
-        renderer.renderTarget().setRenderSize(newPixelSize);
-    renderer.setPageSize(newPageSize);
-    renderer.setMargin(computeMargin(
+    auto const newMargin = computeMargin(
         renderer.gridMetrics().cellSize,
         newPageSize,
         newPixelSize,
-        applyContentScale(session.profile().margins.value(), session.display()->contentScale())));
+        applyContentScale(session.profile().margins.value(), session.display()->contentScale()));
+    renderer.applyResize(newPixelSize, newPageSize, newMargin);
 
     if (oldPageSize.lines != newPageSize.lines)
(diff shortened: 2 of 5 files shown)


jquast:
It's been a bit of a burden to carry a "patched contour", especially to keep them "stashed" when trying to submit fixes, so I have gone without, and contour crashes almost every time I resize, especially if I use the trackpad's scrolling gesture for rapid event, I can't use contour without this patch, I'm surprised that I'm the only one who experiences periodic crashes on window and font resize, I'm sure it's because of writing programs that keep the render thread "more busy", sixel, unicode and color SGR sequences, etc.

jquast:
The time for me to communicate about this patch has ended, you are still welcome to the code but I will not be accepting feedback, best wishes

PR #1929, opened 2026-04-10
Migrate grid storage from templated AoS cells to SoA

christianparpart:
- Replace the templated `Line<Cell>`/`Grid<Cell>`/`Screen<Cell>` hierarchy with concrete, non-templated classes backed by `LineSoA` — a Structure-of-Arrays layout with 64-byte aligned per-column arrays (codepoints, widths, SGR, hyperlinks, grapheme cluster data) grouped by access frequency for SIMD auto-vectorization.
- Add `CellProxy` (lightweight single-cell accessor over LineSoA), `SoAClusterWriter` (bulk ASCII write path with redundant-fill skipping, full UTF-8+grapheme segmentation path), and `fillAttrs` tracking that lets `resetLine`/`writeAsciiToSoA` skip 8 of 11 `std::fill_n` calls when attributes haven't changed — reducing per-line memory writes from ~4000 bytes to ~800 bytes (5× bandwidth reduction for cat-via-fastpipe workloads).
- Delete the entire `cell/` directory (CompactCell, SimpleCell, CellConcept, CellConfig) and fix two SoA-migration regressions: grapheme cluster state is now reconstructed from the preceding cell instead of reusing the parser's already-advanced state (fixes ZWJ emoji sequences), and the ASCII bulk-write path now checks `WideCharContinuation` before overwriting (fixes wide-char right-half destruction).


src/vtbackend/CellProxy.h
```diff
@@ -0,0 +1,234 @@
...
+            return _line->codepoints[_col];
+
+        if (_line->clusterSize[_col] <= 1 || i >= _line->clusterSize[_col])
+            return 0;
+
+        auto const poolStart = _line->clusterPoolIndex[_col];
+        return _line->clusterPool[poolStart + i - 1];
+    }
+
+    [[nodiscard]] char32_t operator[](size_t i) const noexcept { return codepoint(i); }
```
Yaraslaut: bad api, convention is that `operator[]` is mutating operator

src/vtbackend/CellProxy.h
```diff
@@ -0,0 +1,234 @@
...
+{
+
+/// Transient proxy into SoA arrays that provides the same interface as CompactCell.
+///
+/// CellProxy is a lightweight value type (pointer + index) that reads and writes
+/// into LineSoA arrays. It must NEVER be stored across operations that could
+/// reallocate the underlying arrays (e.g., line resize, scroll).
+///
+/// Size: 2 pointers = 16 bytes on 64-bit. Cheap to copy and pass by value.
+class CellProxy
```
Yaraslaut: documentation for the member functions is missing

src/vtbackend/VTWriter.cpp
```diff
@@ -185,45 +186,28 @@ void VTWriter::setBackgroundColor(Color color)
...
+            sgrAdd(GraphicsRendition::Normal);
+
+        setForegroundColor(cell.foregroundColor());
+        setBackgroundColor(cell.backgroundColor());
+        // TODO: other flags (such as underline), hyperlinks, image fragments.
+
+        if (!cell.codepointCount())
+            write(' ');
+        else
+            write(cell.toUtf8());
```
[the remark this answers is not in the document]
christianparpart: -> Make `CellProxy` work with `LineSoA const*`!

Issue #100, opened 2020-04-28
Image Rendering


christianparpart:
The idea is, to not mess around with the main/alt textual buffer for displaying graphics, but also support switching to a graphics buffer that purely supports rendering graphics (such as images).

The app could use the already existing `CSI 14 t` to retrieve pixel dimensions, programmatically feature-test if graphics backbuffer is supported, then switch to it, and then upload the to-be-displayed image in at least the following way(s).

# Checklist

* [ ] `CSI 14 t` must be supported (get screen dimension in pixels)
* [ ] terminal emulator identification via VT sequence? (so apps can test that it is supported).
* [ ] inline image rendering in main/alt buffer (should still be supported)
* [ ] graphics backbuffer enabled image rendering.
* [ ] inline image rendering via sixels? (this is the old xterm way?)

# References
* [Neofetch](https://github.com/dylanaraps/neofetch) is a nice tool to test inline images
* https://github.com/dylanaraps/neofetch/wiki/Images-in-the-terminal

data-man:
@christianparpart 
Some terminal image viewers supports iTerm2 output.
If you don't mind I want to add this feature to Contour. :)

christianparpart:
> @christianparpart Some terminal image viewers supports iTerm2 output. If you don't mind I want to add this feature to Contour. :)

While I'd really love to have progress on image rendering in especially Contour, I do not think that adding iTerm2's image protocol is the right move here - just because some apps have support for it.

If I wouldn't be overcharged right now, I'd continue working on https://github.com/contour-terminal/terminal-good-image-protocol and push for it (will do that as soon as possible). But I am actually really happy to see that you have interest in coding parts of Contour, too. Maybe we can also pick another topic that would not bore you?

data-man:
> I do not think that adding iTerm2's image protocol is the right move here

That's a pity.

> Maybe we can also pick another topic that would not bore you?

I'd be interested in working on libunicode. :)

timsofteng:
Hello. Any updates about it?

christianparpart:
Hey. I keep you posted. We currently focus on moving/upgrading the frontend from Qt5 to Qt6/QML. I certainly want to have the GIP PR merged this year, too.

Sorry if it takes all a little long, but we are limited on resources. If you see fit where you can help us (anywhere anyhow), if you want, feel free to join us on Discord. 

Take care,
Christian.

christianparpart:
[...]

PR #1931, opened 2026-04-12
Close VT level 5 conformance gap

christianparpart:
Closes the VT level 5 conformance gap flagged by j4james in #100. Contour now honestly implements all DA1 extensions it claims via level 65.

## Changes

- **DA1 honest reporting**: Enable SelectiveErase (6), SoftCharacterSet (7), UserDefinedKeys (8), NRCS (9), StatusDisplay (11), TechnicalCharacters (15), Windowing (18), HorizontalScrolling (21), TextMacros (32), AnsiTextLocator (29) in DA1. Widen DeviceAttributes to uint32_t. Change GIP extension from 60 to 90.
- **DECSCL** (Set Conformance Level): Runtime conformance level changes via `CSI Ps1;Ps2"p`, C1 transmission mode tracking, soft reset on level change.
- **Text Macros** (DECDMAC/DECINVM): Define/invoke macros with plain text or hex encoding, recursion guard, deferred execution to avoid re-entrant parsing.
- **User-Defined Keys** (DECUDK): Reprogram F6–F20 via DCS with hex-encoded strings, lock mechanism, wired into function key input path.
- **DRCS** (DECDLD): Full pipeline from DCS parsing through GPU rendering — glyph bitmaps are converted to RGBA ImageFragments and rendered via the existing ImageRenderer (same path as Sixel/GIP).
- **NRCS**: All 11 national replacement character sets wired into SCS dispatch for G0–G3, including generic SCS fallback for two-byte DRCS designators.
- **DEC Locator** (DECELR/DECLRP/DECSLE/DECRQLP): DEC mouse protocol with one-shot mode, configurable event filtering, coordinate unit selection.
- **DEC Technical Character Set**: Full 94-position Unicode mapping (Greek letters, math symbols, bracket pieces, arrows).
- **VT parser fix**: DCS_Param→DCS_Intermediate transition now correctly collects intermediate bytes (was silently dropping them).
- **ScreenBase removal**: Merge into Screen, eliminating unnecessary virtual dispatch after de-templating.

Relates to #100


j4james:
I'm still not sure you understand how the conformance level and extension numbers work in the device attributes report. The purpose of the extensions is to report _optional_ features. Selective Erase, User Defined Keys, Status Display, Technical Characters, and Text Macros are are all _required_ features at level 5, so they're not reported by default on a level 5 device. They're only reported at levels where the feature is optional.

For example, this is what the default `DA1` report is on a VT520:
```
?65;1;2;7;9;12;18;19;21;23;24;42;44;45;46c
```

Note that there is no 6, 8, 11, 15, or 32 in that extension list, because the 65 already tells you that those features are supported.

It's also worth noting that `DECSCL` doesn't alter the conformance level reported by `DA1` - that value always indicates the _maximum_ conformance that the device supports, rather than the currently selected conformance. The only effect it has on `DA1` is that it changes what _extensions_ are reported. In some cases it'll drop extensions (because they're disabled at a lower level), and in some case it will report _more_ extensions (because they were required at level 5, but are optional at the lower level).

christianparpart:
Interesting. Many thanks for actually looking into my PR. Sorry for not fully reading the spec, that's just an overdose (given my eye condition :D). So thanks for heading me to the right direction.

> are not required to be explicitly reported

image

To me, this reads like it's implicit, but not required to not be listed. So debatable. I will of course drop it regardless, as it's *closer* to how VT's used to work. 😄

I'll also adapt the rest as mentioned. Many thanks reviewing and pointing out.

j4james:
I didn't want to get into implicit extensions, because they're a special case which only applies to a few extensions (11, 14, and 17), and they're only applicable when altering your conformance level with `DECSCL`.

For example, the Status Display extension is _required_ for level 3 and above, but _optional_ for level 2 and below. So a level 3 terminal does _not_ report extension 11, while a level 2 terminal _does_ report the extension (assuming it supports that functionality).

Under the normal rules for `DECSCL`, a level 3 terminal that has set its conformance to 2 would usually be expected to report extension 11, exactly like a level 2 terminal does. It doesn't do that, though, because extension 11 is considered _implicit_ at level 2.

Compare that with the UDK extension, which is _not_ implicit at any levels. It's _required_ for level 5, and _optional_ for level 4 and below. So a level 5 terminal does _not_ report that extension, but a level 4 terminal _does_. And a level 5 terminal that has set its conformance to 4 will also report the extension, because it's _not_ implicit.

That said, it's not going to hurt if you are reporting extensions at levels where they aren't technically required. The only thing I ask is that you don't report extensions that you don't actually support, and don't report a conformance level if you don't meet all the requirements.

christianparpart:
Thank you very much, @j4james, for your guidance and especially patience with me. 😄

PR #1944, opened 2026-05-08
Update std::simd feature test.

Dooez:
## Description

Update std::simd macro test from checking include to checking macro. With the release of gcc16 the header is present, but it requires C++26.
The gcc doesn't actually define the `__cpp_lib_simd` macro at the moment, however this should not cause any issues since the project is on C++23 for the time being.

## Motivation and Context

With the release of gcc16 contour fails to build if gcc16 is the system's default (e.g. on Arch) because the header is present, but it's contents are disabled because of the C++ version.

## How Has This Been Tested?

I've built contour locally.

## Checklist

### DIFF
--- a/metainfo.xml
+++ b/metainfo.xml
@@ -123,6 +123,7 @@
           <li>Fixes VT parser: DCS_Param to DCS_Intermediate transition now correctly collects intermediate bytes</li>
           <li>Removes ScreenBase class — merges into Screen, eliminating unnecessary virtual dispatch</li>
           <li>Fixes loading of optional fields in the config file (#1940)</li>
+          <li>Fix std::simd feature test</li>
         </ul>
       </description>
     </release>
--- a/src/vtbackend/Image.cpp
+++ b/src/vtbackend/Image.cpp
@@ -10,7 +10,7 @@
 #include <utility>
 
 // clang-format off
-#if __has_include(<simd>)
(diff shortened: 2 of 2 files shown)


Yaraslaut:
Hi, i also discovered it, I think the issue is not the c++ standard that we use, but the breaking change in the api, see 
https://github.com/contour-terminal/libunicode/issues/130

so what we actually need is to support multiple versions of the headers for the simd :(  from different versions of the `stdlibc++` since it might take a while and we will encounter the same issues with other libraries

Dooez:
While the api changes between the experimental and final versions are indeed present, libstdc++ checks for 
```cpp
#ifdef __glibcxx_simd
```
which itself depends on 
```cpp
# if (__cplusplus >  202302L) && ...
```
So if i'm not mistaken even if the api is updated, std::simd won't be available on C++23.
In my opinion hiding `#include <simd>` behind feature flag is a good enough fix for the time being, and slightly better than outright disabling it.

Yaraslaut:
so, when compiling with gcc16, we will use older header then? `experimental/simd` ?

Dooez:
I haven't checked the exact behavior, but I expect this is what happens.
Note: This also affects clang if gcc16 library is present since if it uses `libstdc++`

christianparpart (dismissed):
Thank you very much, @Dooez

PR #1945, opened 2026-05-09
Fix OOB write crash in SixelParser

jquast:
## Description

This code crashes contour of any window size (aspect ratio):

```python
for ar in range(100, 500):
    print(f'\x1bPq"{ar};1#2' + '~' * 2000 + '\x1b\\', end='', flush=True)
```

The supplied test SixelParser.aspect_ratio_overflow demonstrates out of bounds crash, and the solution is using bounds checks with min().

I also checked "cat snake.six" with varying sizes to ensure correct/same before and after has same working sixel behavior.

https://github.com/csdvrx/sixel-testsuite/blob/master/snake.six

## Related Issues

None

## Motivation and Context

I found heap overflows of the exact same kind in other terminals. This specific crash is unlikely in common sixel use, only misuse by crafted sequences.

## How Has This Been Tested?

with the provided test, and sample python code

## Checklist


christianparpart (approved):
Thank you very much, @jquast :)

PR #1946, opened 2026-05-10
Preserve scrollback for top anchored scroll regions

Sightem:
## Description

Preserve terminal history when a full width scroll region starts at the top of the page. Previously, scrolling a partial vertical region did not append rows that left the top of the region into scrollback. This change reuses the existing full page scroll path to preserve those rows, then restores rows below the requested scroll region because they are outside the scroll operation

## Related Issues

N/A

## Motivation and Context

Applications can use top anchored scroll regions to move completed output into terminal history while keeping a live area below it. in which case rows leaving the top of the scroll region should remain available in scrollback

## How Has This Been Tested?

Built and ran the existing vtbackend_test suite successfully. I am not aware if/how this specific behavior is unit testable

## Checklist

- [x] Breaking changes (if any) are documented (no breaking changes)

### DIFF
--- a/metainfo.xml
+++ b/metainfo.xml
@@ -127,6 +127,7 @@
           <li>Fixes sixel crash in sixel parser caused by aspect ratio miscalculation (#1945)</li>
           <li>Fixes numeric capability reporting by XTGETTCAP</li>
           <li>Fixes window snapping back to the configured terminal_size after a screen/DPR change on tiling Wayland compositors (#1941), by initialising the window size from the terminal's implicit size only once on creation instead of using a permanent QML binding</li>
+          <li>Fixes scrollback preservation for top anchored full width scroll regions</li>
         </ul>
       </description>
     </release>
--- a/src/vtbackend/Grid.cpp
+++ b/src/vtbackend/Grid.cpp
@@ -345,6 +345,29 @@ LineCount Grid::scrollUp(LineCount n, GraphicsAttributes defaultAttributes, Marg
 
         auto const marginHeight = LineCount(margin.vertical.length());
         auto const n2 = std::min(n, marginHeight);
+        if (*n2 == 0)
+            return LineCount(0);
+
+        if (*margin.vertical.from == 0)
+        {
+            auto const historyLinesAdded = scrollUp(n2, defaultAttributes);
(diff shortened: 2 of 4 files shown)


christianparpart:
@Sightem I apologize for the delay, realife.... no comment. However, many thanks for your contribution. I am trying to digg into it. For that matter (and to solve the merge conflict), i had to rebase your branch on top of latest master. So you may want to `git remote update` on your local master branch and check, you may then want to `git reset --hard origin/master` (most likely???) while being in your local master branch, because you sadly worked on master. Sorry for the inconveniences.

On the code logic part. The PR generally looks good, but I want to take extra care with regards to VT sequence semantics. We have a lot of checks covered, but I want to be extra sure :)

NB: what VT client application did you have in mind when doing this PR ? I'd like to locally test this, too.

Sightem:
haha no worries at all, you're good! I am very used to rebasing :p thank you for rebasing and taking a look. 

I only encountered this in codex cli from oai, I imagine the various other "agent harnesses" have the same problem. codex cli as a full screen terminal ui that keeps a prompt/status area pinned while streaming output through a top anchored scroll region. the behavior is that the scroll region starts at row 0 but does not cover the full screen height. in that case, lines scrolled out of the top of the region should still enter normal scrollback, while the protected lines below the region should remain visible. before this, contour moved the region contents inside the page but did not preserve those top lines in history

christianparpart:
Thanks so much for this, @Sightem — and for confirming the Codex CLI use case, that helped a lot.

I hope you don't mind, but I pushed (again...) three small commits on top of yours before merging:

- Tests reproducing a side effect I found: a scroll inside a top-anchored partial region was also notifying onBufferScrolled, which shifted the viewport / Normal-mode cursor / selection even though the live area below the region hadn't moved.
- A fix: only call onBufferScrolled for a full-page scroll (the // TODO that was already sitting there). Your scrollback-feeding behavior is fully preserved.
- A tiny loop-style cleanup (std::views::iota) to match our conventions.

Really appreciate the contribution — merging now. :smile:

christianparpart (approved):

Sightem:
awesome!

PR #1948, opened 2026-05-12
Bugfix XTGETTCAP reporting of numeric capabilities

jquast:
## Description

Contour is the only terminal of the 15 terminals tested for numeric capabilities using XTGETTCAP
that reports values as raw bytes, for example, "Co" is encoded as "\x01\0", but should be 256:

    contour: Co: "\x01\0"

This change allows contour to match all other terminals tested:

    AbsoluteTelnetSSH: Co: '256'
    foot: Co: '256'
    ghostty: Co: '256'
    gnometerminal: Co: '256'
    iterm2: Co: '256'
    kitty: Co: '256'
    lxterminal: Co: '256'
    mlterm: Co: '256'
    rio: Co: '256'
    terminator: Co: '256'
    termit: Co: '256'
    wezterm: Co: '256'
    xfce4terminal: Co: '256'
    xterm: Co: '256'

## Related Issues

None

## Motivation and Context

I wrote my own terminal library, 'blessed', and testing utility, 'ucs-detect', where this problem
has caused me grief, rather than write strange workarounds, I wish to solve it at its source.

It is important to have correct transfer of XTGETTCAP, to allow TUI programs to have "curses-free"
access to capabilities, as described https://codeberg.org/dnkl/foot#xtgettcap

> Foot supports this, and extends it even further, to also include boolean capabilities. This means
> foot's entire terminfo can be queried via XTGETTCAP.

## How Has This Been Tested?

with the given automatic tests and manual testing

## Checklist

### DIFF
--- a/metainfo.xml
+++ b/metainfo.xml
@@ -125,6 +125,7 @@
           <li>Fixes loading of optional fields in the config file (#1940)</li>
           <li>Fixes std::simd feature test for GCC 16 compiler</li>
           <li>Fixes sixel crash in sixel parser caused by aspect ratio miscalculation (#1945)</li>
+          <li>Fixes numeric capability reporting by XTGETTCAP (#XXX)</li>
         </ul>
       </description>
     </release>
--- a/src/vtbackend/Screen.cpp
+++ b/src/vtbackend/Screen.cpp
@@ -2617,12 +2617,7 @@ void Screen::requestCapability(std::string_view name)
     if (booleanCapability(name))
         reply("\033P1+r{}\033\\", toHexString(name));
     else if (auto const value = numericCapability(name); value != Database::Npos)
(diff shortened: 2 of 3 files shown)


jquast:
See https://ucs-detect.readthedocs.io/sw_results/contour.html#contourxtgettcap where the automatic reporting shows `\x01\x00` for colors or any other numeric cap

jquast:
The time for me to communicate about this patch has ended, you are still welcome to the code but I will not be accepting feedback, best wishes

Yaraslaut:
Sorry it took so long to react to this PR, i need to fix my notifications, since i only saw this in discord, and i use it quite rare nowadays, I will merge this one, and look at the other fix with the race conditions in the evening

Issue #1908, opened 2026-03-12
Initial Window size in Hyprland incorrect


mrdgo:
### Contour Terminal version

0.6.3-master-138e0ac4

### Installer source

GitHub: Source code

### Operating System

Arch Linux

### Architecture

x86-64 / amd64

### Shell

other

### GPU & Renderer

`contour info vt` didn't output anything about GPU

### Other Software

Hyprland 0.54.2

### Steps to reproduce

1. Use a tiling window layout in hyprland 
2. Open contour on an empty workspace (doesn't matter if via keybind, launcher or cli)

### Expected Behavior

The terminal window fills the whole screen (except waybar).

### Actual Behavior

The window either opens with the default size of 80x22 or the dimensions specified in the config `contour.yml`. I then change focus to a different window and back to contour - THEN countour adjusts its dimensions to respect the whole space that the compositor assigned.

### Configuration excerpt

```yaml
#Determines the initial terminal size in  characters
        terminal_size:
            columns: 160
            lines: 50
```

This config apparently takes priority until I refocus contour.

PR #1950, opened 2026-05-27
Fix window snapping back to implicit size after screen/DPR change

### LINKED ISSUE #1941: Initial window size incorrect with river (notion-river) tiling WM — input bar off-screen
## Description

When using a tiling window manager (river/notion-river), the contour terminal window initially has the wrong size for the cell it is assigned to by the compositor. The window seems to get progressively smaller as the WM adjusts its size.

For example, after being triggered, the dumps show a progressive shrinking:
- First dump: 153x52
- Second dump: 124x33
- Third dump: 94x25

The last size (94x25) is so small that the input prompt line (cursor at line 24/25) is barely visible or the input bar is pushed off-screen.

I can only see the input bar again after resizing the terminal manually (e.g., switching to another window and back).

## Steps to Reproduce

1. Use river (notion-river) tiling window manager on Wayland
2. Open contour in a tiled workspace
3. The terminal opens with an incorrect initial size — too small for the allocated space
4. The cursor/prompt ends up at the very bottom (line 24 of 25), making input difficult or impossible to see

## Expected Behavior

The terminal should immediately adopt the size assigned by the tiling WM, not require a focus switch to resize correctly.

## Actual Behavior

The window opens too small, and the resize only happens after switching focus away and back.

## Additional Context

- **Contour version**: 0.7.0-master-a38ae102 (commit a38ae102, built from source/package)
- **Window Manager**: river (notion-river) on Wayland
- **Shell**: zsh (with custom prompt showing on last line)
- **Screenshots & dump attached**

## Related Issues / PRs

- #1908 / PR #1904: "Fix initial resizing with tiling WM" — this PR attempted to fix a similar issue for Hyprland but the fix does not seem to cover the river case. The fix changed the detection logic from comparing against `implicitWidth/Height` to comparing against saved `_lastVirtualWidth/Height` values. On river, the initial window dimensions might not match the saved stale dimensions either, or the timing differs.
- #1922 (open): "Bugfix segfaults when changing font/window size" — possibly related

## Investigation Notes

PR #1904 saved the pre-DPR-correction window dimensions (`_lastVirtualWidth/_lastVirtualHeight`) in `createRenderer()` and then in `sizeChanged()` checks if the compositor reverted to those stale dimensions. If the compositor (river) assigns a new tile size *after* `createRenderer()` but *before* the stale-dimension revert, the check might not trigger, leaving the window at the WM-assigned tile size rather than correcting to the implicit size. Alternatively, river might not trigger the revert pattern that Hyprland does, meaning the fix doesn't apply at all.

## Attachments

- `screenshot.png` — screenshot at time of failure (94x25, very small)
- `texture-atlas-rgba.png` — texture atlas
- `screen-state-dump.vt` — full screen state dump

Marenz:
## Description

The `ApplicationWindow` in `src/contour/ui/main.qml` had permanent QML bindings:

```qml
width: vtui.implicitWidth
height: vtui.implicitHeight
```

These bindings re-evaluate every time `implicitWidth` / `implicitHeight` change. After the window is placed by the compositor and contour's `onScreenChanged()` handler runs, `applyFontDPI()` calls `updateImplicitSize()` which sets the implicit size back to the configured `terminal_size` (e.g. 800x525 virtual px for 80x25 at DPR 1). The QML binding then snaps the window back to that size, overriding the WM-assigned tile geometry.

This is what was happening on river (notion-river) — `propose_dimensions(1436, 1183)` was honoured for the first surface commit, then the screen-change handler fired and the window committed again at 800x525.

## Related Issues

Fixes #1941.

Note: #1908 / PR #1904 attempted to fix the same class of bug for Hyprland by adding a guard against "stale" Wayland configure reversions in `sizeChanged()`. That guard is independent of this fix and is not affected — the actual size revert in the river case happens via the QML binding, not via the stale-configure path.

## Motivation and Context

Tiling WMs (river, hyprland, sway, ...) assign window geometry explicitly. The terminal must respect that. With a permanent `width: implicitWidth` binding, any code path that recomputes `implicitWidth` (DPR change, font change, screen change) overrides the WM-assigned size, leaving the terminal stuck at the configured `terminal_size` until the user manually resizes the tile.

Floating WMs also benefit: a font-size change no longer auto-resizes the window. This matches the behaviour of every other terminal (foot, alacritty, kitty, gnome-terminal, xterm).

## How the fix works

Replace the bindings with a one-shot `Component.onCompleted` initialisation:

```qml
Component.onCompleted: {
    width = vtui.implicitWidth
    height = vtui.implicitHeight
}
```

Direct assignment in JavaScript breaks the QML binding immediately, so subsequent changes to `implicitWidth` no longer propagate to the window. The window still opens at the configured `terminal_size` on first show.

## How Has This Been Tested?

- Built on Linux (openSUSE Tumbleweed, GCC 15, Qt 6, Wayland/river).
- Verified the original bug reproduces on master (38e543a7): contour spawned in a river tile commits at the WM-assigned size, then a second commit drops it to 800x525.
- Verified the fix: contour spawned in the same tile commits at the WM-assigned size and stays there. Subsequent re-tiling, focus changes and font-size changes (Ctrl+= / Ctrl+−) all leave the WM-assigned geometry intact.
- Floating-WM smoke test pending — submitting as draft so a maintainer can sanity-check on KDE/Gnome before merging.

## Checklist

- [x] I have read the **CONTRIBUTING** guide
- [x] I have updated the documentation accordingly (metainfo.xml release notes)
- [ ] I have added tests to cover my changes (QML window-sizing behaviour is not currently covered by the test suite; happy to add one if you can point me at the right place)
- [x] Breaking changes (if any) are documented — minor UX change: font-size changes no longer auto-resize the window, matching all other terminals

### DIFF
--- a/metainfo.xml
+++ b/metainfo.xml
@@ -125,6 +125,7 @@
           <li>Fixes loading of optional fields in the config file (#1940)</li>
           <li>Fixes std::simd feature test for GCC 16 compiler</li>
           <li>Fixes sixel crash in sixel parser caused by aspect ratio miscalculation (#1945)</li>
+          <li>Fixes window snapping back to the configured terminal_size after a screen/DPR change on tiling Wayland compositors (#1941), by initialising the window size from the terminal's implicit size only once on creation instead of using a permanent QML binding</li>
         </ul>
       </description>
     </release>
--- a/src/contour/ui/main.qml
+++ b/src/contour/ui/main.qml
@@ -24,8 +24,17 @@ ApplicationWindow
     // title: "%1 - Contour".arg(vtui.session.title)
     title: vtui.title
 
-    width: vtui.implicitWidth
-    height: vtui.implicitHeight
+    // Initialise the window size from the terminal's implicit (configured)
+    // size once on creation. Do not use a binding (`width: vtui.implicitWidth`)
+    // because that would re-evaluate every time implicitWidth changes — e.g.
+    // when updateImplicitSize() runs after a screen/DPR change — and snap the
+    // window back to the implicit size, overriding any WM-assigned geometry.
+    // Tiling WMs that resize the window to a tile, and users who resize
+    // manually, expect the window to keep its assigned size.
+    Component.onCompleted: {
+        width = vtui.implicitWidth
+        height = vtui.implicitHeight
+    }
 
     Terminal {
         id: vtui


Yaraslaut (approved):
Oh, thanks a lot for figure it this out, i completely missed the idea that this can be something with the qml when i was trying to fix hyprland behavior :)

PR #1955, opened 2026-06-15
Fix terminal grid not syncing to tile size on tiling WMs

Marenz:
On tiling Wayland WMs (reproduced on river), a freshly-spawned terminal is mis-sized until a manual resize, due to two related faults:

1. The stale-geometry correction in `sizeChanged()` (added in #1904) misfires on a legitimate WM-assigned tile size, force-resizing the window down to the implicit/configured size (e.g. 80x25). It now only triggers when the implicit size actually changed since `createRenderer()` (i.e. a real DPR correction).

2. `onBeforeSynchronize()` sets the render viewport from the live window size, but the grid/page size is computed separately on the GUI thread; at initial map the grid can be left larger than the viewport (e.g. 131x67 cells in a ~98x50 viewport), so bottom rows render off-screen. It now posts a `sizeChanged()` to reconcile when the page size diverges; `applyResize()` no-ops when already in sync (verified no resize loop while idle).


christianparpart (dismissed):
Thank's, @Marenz :-)

PR #1959, opened 2026-06-27
refactoring: PtyHandles in UnixPty moved from public interface

Vheath:
PtyHandles was in public interface, but used only in one .cpp file and should better be defined only there

## Description

what: Moved PtyHandles struct containing two fields from public interface of UnixPty.
why: It polluted UnixPty public interface.

## Motivation and Context

Well just makes the codebase better

## How Has This Been Tested?

Does not need to be tested

## Checklist

### DIFF
--- a/src/vtpty/UnixPty.cpp
+++ b/src/vtpty/UnixPty.cpp
@@ -58,7 +58,13 @@ namespace vtpty
 
 namespace
 {
-    UnixPty::PtyHandles createUnixPty(PageSize const& windowSize, optional<ImageSize> pixels)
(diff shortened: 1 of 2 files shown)


christianparpart:
Interesting find. Many thanks. How come? What was your motivation here? Do you need these somewhere?

Thanks

Vheath:
I'm writing my terminal and was just exploring the source code and thought that it would be a little bit better than before :)

Thanks for the incredibly high-quality codebase!

christianparpart (approved):

PR #1964, opened 2026-07-01
Native GUI tabs, split panes, and an RHI-rendered display

christianparpart:
Reworks Contour's frontend: the terminal now renders through Qt RHI, and tabs and split panes are first-class GUI citizens in a custom title bar, with window sizing that cooperates with the window manager instead of fighting it.

### Adds
- **GUI tabs** in a custom window title bar, replacing the status-line tab list (the `{Tabs}` status-line directive still works for existing configs). Double-click to rename; right-click for a color-swatch picker.
- **Split panes** — split vertically (`Ctrl+Shift+E`) or horizontally (`Ctrl+Shift+O`), close a pane (`Ctrl+Shift+X`), move focus (`Alt+Shift+Arrow`).
- **Pane management actions** — swap the active pane with a neighbor (`Ctrl+Alt+Arrow`), resize it (`Ctrl+Shift+Alt+Arrow`), flip a split's orientation (`Ctrl+Shift+Backslash`), and move a pane elsewhere in the split tree (action available, unbound by default).
- **Tab dragging** — reorder a tab in its strip, drag it into another window, or tear it out into a new window, with a live drop-position indicator showing where it will land. The running terminal survives the move (same process).
- **`dim_unfocused`** profile option — dims an unfocused pane (or all panes of an unfocused window) toward its background; `0.0` (default) disables it.

### Changes
- The terminal renderer moves to **Qt RHI**.
- The window uses a **custom client-side-decorated title bar** (tab strip + window controls) by default on all platforms. Set `show_title_bar: true` to keep the native OS decoration — the tab strip stays, its controls defer to the OS frame.
- **Font zoom** now keeps the window and pane size fixed and adjusts columns/lines (and the child PTY size) instead of resizing the window.

### tested on

- [x] Linux (OpenGL, Vulkan, Software)
- [x] Windows (Direct3D, Vulkan, Software)
- [x] macOS (OpenGL, Metal, Software)


Yaraslaut:
I tested for a bit, and did not notice any issues, thanks a lot, the only small point is that in the active tab, when i hover mouse to close the tab, the highlight is white and then i do not see a mark there

PR #1966, opened 2026-07-02
fix(contour): avoid GL_INVALID_OPERATION when applying text outline color off the render thread

Yaraslaut:
## Summary
- `OpenGLRenderer::setTextOutline()` called into the shader (`bind()` / `setUniformValue()`) unconditionally once the renderer had been initialized, assuming an OpenGL context was always current on the calling thread.
- `Renderer::applyStagedReconfigDuringSetup()` is documented to run font/DPI reconfiguration synchronously from the GUI thread during initial display setup and for the minimized/occluded case, not just from the render thread that normally owns the current context.
- Applying it in that situation made `setUniformValue(_textOutlineColorLocation, ...)` fail with `GL_INVALID_OPERATION` (error 1282), logged as an OpenGL error.
- `setTextOutline()` now checks `QOpenGLContext::currentContext()` and, when absent, only updates the CPU-side color and marks it dirty. The pending color is uploaded lazily from `execute()`, which always runs on the render thread with the context current while the text shader is bound.

## Test plan
- [x] `contour.exe` builds cleanly with the fix
- [x] `vtrasterizer_test.exe` — all 624 assertions / 36 test cases pass


christianparpart:
how come? hmm...
@Yaraslaut how much sense does this make when we're about to land my Qt RHI port (including tabs/splits, #1964)?

christianparpart:
Closing as #1964 will implicitly have this addressed.

PR #1978, opened 2026-07-09
fix(contour): don't inherit a remote SSH cwd into a new local tab (Windows crash)

Yaraslaut:
## Problem

Creating a **new tab** (or split) while the current tab is running an **SSH session** crashes the entire application on Windows. There is no useful log output — only:

```
Warning: Qt has caught an exception thrown from an event handler. Throwing
exceptions from an event handler is not supported in Qt.
```

## Root cause

Confirmed under `cdb` by catching the C++ throw. Bottom-up the crashing stack is:

```
QQuickAbstractButtonPrivate::trigger              (mouse click on the "+" new-tab button)
 -> WindowController::createSessionInBackground
 -> SessionModel::createTab -> activeTabChanged -> rebuildActiveTabPaneProxies
 -> QML `session:` binding write -> TerminalDisplay::setSession
 -> TerminalSession::start
 -> vtpty::Process::start   ==> _CxxThrowException   (std::runtime_error, Process_win32.cpp:251)
```

A new tab inherits its working directory from the current pane via `TerminalSession::workingDirectory()`. On Windows that value comes only from **OSC 7**, which an SSH session reports as the **remote host's** path (e.g. `file://remotehost/home/user` → `/home/user`). That directory does not exist on the local machine, so when it is passed to the new local shell's `CreateProcess()` as `lpCurrentDirectory`, `CreateProcess()` fails (both the primary command line and the login-shell fallback) and `Process::start()` throws `std::runtime_error("Could not create process…")`.

Because the throw is reached from a QML property binding write **inside a Qt event handler**, Qt cannot let it propagate and aborts the process (`FAST_FAIL_FATAL_APP_EXIT`).

This is the same failure class the existing OSC-7 URL fix addressed one layer down (a raw `file://` URL crashing `CreateProcess`), just with a *remote* path this time.

## Fix

`workingDirectory()` now returns the extracted OSC-7 path only when it **exists as a directory on the local machine**; otherwise it falls through to the existing `"."` sentinel (which `CreateProcess` treats as "inherit the parent's cwd"). This fixes every spawn path that inherits a cwd — new tab and split pane — in one place, and the non-Windows behaviour is unchanged.

## Testing

Added a regression test (`[contour][session][cwd]`) that drives a remote/nonexistent OSC-7 cwd and asserts the `"."` fallback, plus a local-path case.

- Without the fix the test **fails**, `workingDirectory()` returning the nonexistent remote path (`"/this/path/does/not/exist/anywhere"`) — the exact value that crashes `CreateProcess`.
- With the fix it returns `"."`.
- Full `[session]` suite: **All tests passed (310 assertions in 56 test cases)**.


christianparpart (dismissed):
thank you, @Yaraslaut

PR #1983, opened 2026-07-11
ci: bump static Qt image to 6.11.1 to restore the static build

Yaraslaut:
### Problem

The "Static build" CI job has been failing on master since #1959 (`630947b4`, RHI-rendered display) with:

```
-- Could NOT find Qt6GuiPrivate (missing: Qt6GuiPrivate_DIR)
CMake Error at src/contour/CMakeLists.txt:30 (find_package):
  Expected Config file at
  "/opt/qt/lib/cmake/Qt6GuiPrivate/Qt6GuiPrivateConfig.cmake" does NOT exist
```

Root cause: contour now requires the `Qt6GuiPrivate` CMake package (for `rhi/qrhi.h`), but separate CMake packages for private Qt modules only exist since **Qt 6.9.0** (`qt_internal_write_basic_module_package(... PRIVATE)` in qtbase's `QtModuleHelpers.cmake`). The `yaraslaut/static_qt:24.04` image is built with Qt 6.8.3, which structurally cannot provide that package — no image rebuild at 6.8.x could fix it.

### Why Qt 6.11.1

Qt 6.8 is the newest LTS but predates private-module packages; 6.9 is EOL (April 2026). 6.11.1 is the current release series and matches what distros ship today (e.g. Arch), so contour is already exercised against it. The next LTS (6.12) can be picked up later with a one-line `QT_VERSION` bump.

### Changes

- `DockerQt`: bump `QT_VERSION` 6.8.3 → 6.11.1, and derive the download URL's `qt/6.x/` path segment from `QT_VERSION` instead of hardcoding `6.8`
- `DockerUbuntu`: consume the new versioned image tag `yaraslaut/static_qt:24.04-qt6.11.1` (already built and pushed to Docker Hub, digest `sha256:000f3daa…`), so future Qt bumps don't silently replace the tag CI builds against
- `readme`: document the tag scheme and the Qt >= 6.9 requirement
- `metainfo.xml`: changelog entry

### Verification

Reproduced the CI job locally with the exact `docker buildx build -f .github/static/DockerUbuntu` invocation:
- against the old image: fails identically to CI
- against the pushed Qt 6.11.1 image: configure passes (`Qt6GuiPrivate` found), full build and static link succeed, and the extracted binary runs: `Contour Terminal Emulator 0.7.0-static-build-qt6.9-417b3675 (pre-amend commit hash; identical tree)`

Note: the `staticBuild` job is gated to master (`if: github.ref == 'refs/heads/master' || github.head_ref == 'release'`), so PR checks won't run it — it will first run on CI after this merges.


christianparpart:
Many thanks for that one, man, @Yaraslaut!

christianparpart:
hold on, why not Qt 6.11.1 then? It's statically linked anyways?

Yaraslaut:
so 6.9.3 was the first version to fix the issue, will bump to 6.11.1

christianparpart (approved):
Lovely. Thanks a lot

PR #1992, opened 2026-07-14
Fix Enter, Tab and Backspace not reaching apps under Win32 Input Mode

Yaraslaut:
## Summary
- `keyToUnicodeChar()` in `InputGenerator.cpp` left the `UnicodeChar` field at 0 for `Key::Enter`, `Key::Tab`, and `Key::Backspace`, so ConPTY delivered a Win32 key record with no character for these keys.
- Applications that key off the character field rather than the virtual-key code (e.g. PSReadLine-based confirmation prompts) silently ignored the keystroke — pressing Enter appeared to do nothing, even though the escape sequence was correctly transmitted.
- This is the same class of bug as the earlier Escape/numpad-key fix already in `metainfo.xml`; Enter/Tab/Backspace were simply missed by that fix.
- Fixed by returning `0x0D`/`0x09`/`0x08` respectively, matching a real Windows `KEY_EVENT_RECORD`.
- Updated 3 existing tests that had asserted `UC=0` for these keys (one test's own comment already contradicted its assertion).
- Added a `metainfo.xml` release note entry following this project's convention.

## Test plan
- [x] `cmake --build . --target vtbackend_test --config Release` builds cleanly
- [x] Full `vtbackend_test` suite passes: 743 test cases, 7165 assertions


christianparpart (dismissed):
i am surprised i did not notice that .... 🤔

christianparpart:
@Yaraslaut I've dropped (and cleaned) the metainfo.xml (changelog entry). This addition was introduced after the last public stable release, so no need to add a fix-mention in the changelog.

PR #2004, opened 2026-07-18
Add some kitty protocols and improve Unicode support

christianparpart:
Contour measured character width one codepoint at a time, so a ZWJ sequence, a skin-tone modifier or an Indic conjunct was sized from its base alone and the rest of the cluster was unaccounted for. This branch measures the whole grapheme cluster instead, and implements the graphics and text-sizing protocols applications increasingly expect.

## References
- https://github.com/jquast/ucs-detect/pull/53

## Changes

- Measures width over the whole grapheme cluster, so ZWJ sequences, skin-tone modifiers and Indic conjuncts occupy the columns they actually need
- Makes cluster measurement switchable: it is on by default, and resetting DEC mode 2027 returns width measurement to the classic per-codepoint `wcwidth()` model, where a cluster is sized by its first codepoint and nothing arriving later revises it
- Implements the kitty graphics protocol (APC G): transmit, display, place and delete, in PNG or raw RGB/RGBA, including chunked transmission
- Implements the kitty text sizing protocol (OSC 66): text several cells tall and wide, with fractional scales and alignment inside the block. The renderer composites each block once and cuts it into cell-sized atlas tiles
- Implements the kitty pointer shape (OSC 22) and clipboard (OSC 5522) protocols, and the iTerm2 OSC 1337 capabilities report and inline images
- Implements in-band window resize notifications (DEC mode 2048), and reports `TERM_PROGRAM` / `TERM_PROGRAM_VERSION`
- Resolves a missing glyph by asking which installed font actually covers it, rather than giving up at the end of the fallback chain
- Requires libunicode 0.9.2, which measures a cluster the way `wcwidth()` does and confines a variation selector's effect to the bases Unicode defines a sequence for
- Fixes two defects that predate the branch: a glyph taller than its cell overwrote a neighbour in the texture atlas, and the startup audio probe could hang or crash for profiles that cannot ring an audible bell


j4james:
> DEC mode 2027 selects whether a codepoint arriving after the first may still revise the cluster's width

This sounds like you're now allowing VS15 to shrink the cluster width when mode 2027 is active, but that's exactly what the spec says you _SHOULD NOT_ do.

See here:
https://github.com/contour-terminal/terminal-unicode-core/blob/64f53851ceab9a3cf08db4939bcaef75a0899573/spec/terminal-unicode-core.tex#L186-L188

We discussed this at length before. You can't make a wide glyph narrower after it has been output, because that requires you to alter state that is past the point of being changed (like unwrapping a line that has already wrapped, or unscrolling content that has already disappeared offscreen). Terminals that attempt to do this just end up with unpredictable behaviour.

And the ucs-detect tests are just wrong. If you're trying to "pass" them, you're only going to make your terminal worse in my opinion.

christianparpart:
@j4james  You are absolutely right! What just came into me! I was so into improving conformance scores compared to other TEs, that I lost track. Thanks for reminding me!

PR #2007, opened 2026-07-19
Fix key bindings silently not firing, and report the ones that are dropped

### LINKED ISSUE #1987: `ClearHistoryAndReset` action doesn't work by pressing a binded key
### Contour Terminal version

Contour Terminal Emulator 0.7.0-master-7d0894b3

### Installer source

GitHub: Source code

### Operating System

Devuan/Ceres

### Architecture

x86-64 / amd64

### Shell

None

### GPU & Renderer

_No response_

### Other Software

_No response_

### Steps to reproduce

1. Add to
```yaml
- { mods: [Shift,Alt,Ctrl]    , key: 'Q'                     , action: ClearHistoryAndReset }
```
2. Start Contour
3. Press `Ctrl-Alt-Shift-Q`

### Expected Behavior

Screen should be cleared.

### Actual Behavior

Nothing.

### Configuration excerpt

```yaml

```

### Additional notes

_No response_

christianparpart:
Binding an action to a chord could do nothing at all, with no way to find out why:

```yaml
- { mods: [Shift,Alt,Ctrl], key: 'Q', action: ClearHistoryAndReset }
```

`parseModifierKey` accepted only `Control`, never `Ctrl`, so `parseModifier` bailed and the whole row was discarded by an `if (action && mods && mode)` that had no `else` — no diagnostic anywhere. The action still worked from the command palette, because that path calls `executeAction()` directly and never consults the binding tables, which is what made the report read as "the action is broken" rather than "the binding never existed".

`Ctrl` is not an odd thing to have written: it is the spelling Contour itself renders. `ShortcutModifierTable` maps `Modifier::Control` to `"Ctrl"`, so the palette and the context menus teach a chord the config parser then refused.

Chasing that turned up a second, independent defect with the identical symptom, and one consequence of it that goes beyond configuration: a `key: 'q'` binding also parsed cleanly and could never fire, because which case a letter reaches a binding in is decided by the input route rather than by the user. On macOS the option-as-Alt route derives the case from Shift XOR CapsLock and preempts the Ctrl branch, which left the shipped `Ctrl+Alt+V/S/K/J/O` bindings dead unless CapsLock happened to be on — a latched lock key deciding which shortcut fires, exactly what `sendCharEvent`'s own comment says must not happen.

## Changes

- The `mods:` vocabulary is now a table. `Ctrl` joins `Control`, and `Hyper` becomes bindable at all — it renders in the shortcut column but had no config spelling, so Contour could write a `contour.yml` it could not read back. `Meta` keeps meaning `Super`, pinned by its own test, because re-pointing it would silently move existing bindings onto a key most keyboards do not have.
- A `static_assert` in each direction holds the three modifier tables together: every chord modifier must parse back to itself, and every name the UI renders must be a spelling the parser accepts. The missing half of that is what let this issue happen.
- `bypass_mouse_protocol_modifier` and `mouse_block_selection_modifier` share the same parser and were ignoring `Ctrl` for the same reason; they are fixed with it.
- A row that fails to parse now names itself and the field that killed it, and a non-list `input_mapping` says so instead of leaving the user with a keyboard that has nothing bound.
- Character bindings are folded when parsed and the delivered codepoint is folded when looked up, so the case a binding was written in no longer matters. Neither half works alone — folding only at parse breaks `key: 'a'` bindings that work today. Only the lookup is folded; what the application receives is untouched. Deliberately US-ASCII, not `unicode::simple_uppercase`, since folding above 0x7F would merge bindings that are distinct today.
- Documents that defining `input_mapping` replaces the built-in bindings rather than adding to them — intended behaviour that was previously only true, not written down.

Fixes #1987


data-man:
Thank you, works for me! 🤝

PR #2012, opened 2026-07-21
[EXPERIMENTAL] Add daemon mode

christianparpart:
Contour has always multiplexed — tabs and split panes, in one process. This adds a **daemon mode** that moves
that ownership out of the window: a Qt-free daemon owns the sessions and their layout, and a thin-client GUI
connects to it, over Contour's own native protocol or tmux's.

tmux compatibility works in **all three directions** — as a control-mode server for tmux-aware tooling, as a
client of real tmux, and as an imsg server for the stock tmux binary itself.

Sessions outlive the window that shows them: `contour client` starts a daemon if none is running, mirrors its sessions as tabs, and reattaches after the window closes.

Design record: [`docs/internals/vthost.md`](docs/internals/vthost.md). User guide: [`docs/persistent-sessions.md`](docs/persistent-sessions.md).

## Highlights

- **One session, two taps**: raw PTY bytes feed tmux control mode (`%output`, byte-exact); per-line cell deltas with stable row ids feed the native protocol (wezterm's model).
- **tmux control-mode server** pinned to tmux 3.7b semantics, conformance-tested against the real binary (guards, octal escaping, dual queues, pause/continue budgets, layout checksums).
- **tmux client**: gateway + client model drive a real `tmux -C` (oracle-tested over forkpty), with capture-pane history replay.
- **Binary imsg endpoint**: the stock tmux binary attaches (`tmux -S <sock>-tmux -C attach-session`) — identify handshake with SCM_RIGHTS fd passing, then the existing control-mode engine over the passed descriptors.
- **GUI client**: `contour client` mirrors daemon sessions as tabs (ScreenMirror re-serializes deltas to VT, so the session's own parser emulates and scrollback/selection/search work natively); `contour client --tmux` mirrors a real tmux server (windows → tabs, panes → splits).
- **Optional remote access**: `--listen-tcp` serves the native protocol over TLS with a required preshared token; `--connect-tcp`/`--tls-ca` on the client.
- **Windows**: AF_UNIX via `afunix.h`, ConPTY daemon sessions — CI-validated only.

## Security

The TCP transport was hardened before shipping:

- `--listen-tcp` **requires** a token — an empty expected token is right on a unix socket, where file permissions are the gate, and wrong on a network port.
- `--token-file` on both verbs, because `--token` puts the secret in `argv` where `/proc/<pid>/cmdline` and `ps` expose it.
- Token comparison is constant-time (`CRYPTO_memcmp`); `std::string::==` leaks how much of a guess was right.
- With `--tls-ca`, the certificate must be **issued for the host connected to**, not merely signed by that CA.
- Wire decoding range-checks every integral field instead of narrowing, so an out-of-range value is a protocol error rather than a plausible small one; `MaxGridExtent` bounds announced *and* proposed grids, since the client sizes a repaint buffer from the announced one.
- An exception escaping a connection handler is reported with the connection's identity. The daemon still survives it — one bad peer must not take it down — but it no longer does so silently.

Without `--tls-ca` the client encrypts without verifying which daemon answered (trust on first use, with the token as the authenticator). That is documented in the user guide rather than left implicit.

## Performance

vtbackend retrofit (per-line dirty bits, stable row ids), measured clang-release with Callgrind against the pre-retrofit baseline:

- total Ir **+0.30 %** (acceptance budget was < 0.5 %)
- `writeTextToSoA` and `Screen::writeText` byte-identical
- wall-clock parity: ~103.7 → ~104.5 MB/s aggregate, per-test within noise

An idle or daemon-less terminal pays nothing: revisions are stamped lazily, only when a consumer queries.

## Testing

Four new test targets — `coro_test`, `net_test`, `vthost_test`, `vtworkspace_test` — plus `ChannelPty_test` and three controller suites in `contour_gui_test`. All 19 ctest targets pass under ASan/UBSan, clang-tidy clean under `WarningsAsErrors: '*'`.

Two CI gaps are closed as part of this PR, and both found real bugs on their first run:

- **tmux is now installed on Linux.** Four oracle tests — the ones that validate our wire format against the real binary — had never executed on any runner. A follow-up step *asserts* they ran, because a Catch2 skip prints "skipped" and still exits 0, so a green suite could not distinguish "tmux agreed with us" from "nothing ran".
- **macOS now runs tests through `ctest`** instead of a hard-coded four-binary list that predated these modules. That first run found a test asserting Linux kernel behaviour as portable (`MSG_CTRUNC` on an overflowed SCM_RIGHTS buffer) and, separately, that macOS strips `Alt` before the key-binding lookup so `mods: [Alt]` bindings cannot fire there — a pre-existing defect, skipped with the reason recorded rather than asserted into the contract.

The imsg oracle still self-skips: it needs the rewritten framing from tmux ≥ 3.6 and Ubuntu 24.04 ships 3.4.

## Risk

**Medium**, and concentrated away from existing users:

- Everything is **opt-in**. Without `contour daemon` or `contour client`, no new code path runs; the local GUI behaves exactly as before.
- The shared surface is the vtbackend retrofit — dirty bits, stable ids, generation resync — which is where a regression would reach ordinary users. It is measured (above), covered by `vtbackend_test`, and the ring-buffer invariants are asserted directly.
- `src/net` carries the security weight: AF_UNIX hardening mirrors tmux (0700 directory, `lstat` not `stat`, owner check, world-bit refusal, anti-hijack probe), and every decoder has an explicit bound.
- Windows is CI-validated only; no maintainer runs it interactively.
- Known gaps, all documented: no headless/TUI client (`contour client` opens a window), no scrollback backfill older than the attach, no session survival across a daemon restart, and no images through tmux-mode replay (a tmux `capture-pane` limitation, not ours).


Dooez:
This is incredible

data-man:
When is the 0.7.0 release scheduled, August 14? 😉

christianparpart:
> When is the 0.7.0 release scheduled, August 14? 😉

@data-man  no schedule yet except that I want it ASAP. If that feature here will make it, is by far from promised. I care more about getting 0.7.0 out than having daemon mode in 0.7.0. I want @Yaraslaut and me first have used master branch for some more days before cutting the 0.7.0 release.
I also should try again on attempting to get macOS builds working again (same old certificate issues, macOS/Apple is giving me hard times, they really don't like OSS developers, it seems, or I'm just not made for understanding macOS...).

In the meantime, I keep on working on daemon mode - regardless of having this one landed before or after the 0.7.0 release :)

PR #2021, opened 2026-07-26
Fix status line template round-trip; group and search the settings page

Yaraslaut:
The settings page was silently corrupting the profiles it edited, and its ~60 new
profile fields landed as one flat column of sixty cards. This fixes the data loss
and gives the page a structure you can navigate.

## The data loss

`serializeStatusLineSegment()` emitted `{InputMode,Bold,Color=#ffff00}` — a comma
where the template grammar needs a colon. `crispy::parse_interpolation` splits a
placeholder at its **first** colon and splits what follows on commas, so with no
colon the whole attribute list became the placeholder *name*, matched nothing, and
was echoed back as literal text. Every styled placeholder was destroyed by one
round trip, which is what the settings page does on every open and save.

Four more fidelity bugs alongside it:

| What | Effect |
|---|---|
| `Text` escaped `{` as `{{`, but the parser documents that it has no brace escaping | the text gained a brace on every save |
| `parseIndicatorSegment()` exposed 4 of 13 cell flags | a profile using `Underline`/`Inverse`/`CrossedOut` lost it on save |
| `serializeIndicatorSegment()` rebuilt `Tabs { styles, {}, {}, {} }` | opening and saving dropped the `ActiveColor` the shipped default sets |
| `history_max_lines` reported `Infinite` as `0` and read `0` back as `LineCount(0)` | unlimited scrollback became *no* scrollback on first touch |
| a bare `{Text}` hit `attributes.at("text")` | `std::out_of_range` thrown at user-authored config |

## Why it was possible

The placeholder vocabulary was hand-written in **six** places — the parser's name
chain, thirteen identical serializer `visit()` overloads, seventeen visitor
overloads in the GUI bridge, a seventeen-branch name chain back, plus two lists in
QML. Adding a placeholder meant six edits, and the copies had already drifted.

Everything now reads one `StatusLineDefinitions::ItemTraits` table (one
specialization per `Item` alternative: template name, label, preview sample),
following the table-driven idiom `CellFlags.h` already establishes and explains.
The cell flags likewise share `CellFlagNames`, and `Decorator` — which carried its
nine tokens twice, the second time as a bare array indexed by the enumerator —
gets one `DecoratorNames`.

`enumField()` was the same shape of hazard: a hand-written option list beside a
chain of string comparisons, with the getter deriving its token from `std::format`
— a third spelling, since the formatters render values for people (`"host
writable"`, `"Logical"`) rather than as config tokens. A `Decorator` field would
have shown nothing selected. It now takes the token↔value table and derives all
three from it, which also retires three empty `catch (...) {}` blocks and brings
the descriptor list back under the cognitive-complexity limit clang-tidy enforces.

## The page

- One collapsible `SettingsSection` per field group, with a search box that filters
  on label, help text or YAML key and reopens what it matched. Grouping is data on
  the C++ descriptors, which let the three `status_line_indicator_*` descriptors go
  and removed the `key.indexOf("status_line_indicator_")` hack from QML.
- A real indicator editor: preview strip, chips per placeholder, and a dialog
  exposing everything the grammar accepts — all 13 flags, both colours, side
  adornments, `Tabs`' active colours and separator — plus a raw-template escape
  hatch per segment.
- Also fixed in that editor: the chip's fill `MouseArea` was declared *after* the
  remove button and stacked over it, so the button could not be clicked; chips
  ignored `editable`; a segment summed the heights of two items sitting side by
  side; and the style popup mutated a plain JS object, whose fields notify nothing,
  so the colour swatches never updated.

Also: the parser now reads `Tabs`' `Separator`. `serializeToVT()` already consumed
`Tabs::separator` (falling back to `"|"`), but the parser hardcoded `nullopt` — the
field was write-only dead state no config could reach.

## Verification

- Build clean under `-Werror`; clang-tidy clean on both changed `.cpp` files (run
  manually — the preset has it off).
- 10365 assertions in 651 test cases pass. +19 cases: round-trip over the shipped
  default indicator segments, all 13 flags, `Tabs` attributes, both side
  adornments, and a loop asserting every enum field's current value appears in its
  own options list.
- 99% line coverage (104/105) on the changed vtbackend code, measured with the
  `clang-coverage` preset. Measuring found two real gaps — the `Right=` adornment
  and `{Command}` without a program — which are now tested.
- No hot-path impact: parse and serialize run on config load and on settings edits.

**Pre-existing, not from this branch:** `ctest` fails `contour_gui_test` on
LeakSanitizer. Verified on a clean `master` checkout, which fails the same way. No
leak trace references a file this branch touches.

The root cause is Qt's `QTextToSpeech`, which leaks a fixed ~418 bytes in 2
allocations per instance — reproducible in fifteen lines with no contour code, and
perfectly linear in the number of instances. `TerminalSession` constructs one
eagerly per session (`TerminalSession.h:792`), so a full test run builds ~147 of
them and that single Qt bug accounts for ~77 KB of the ~79 KB reported. CI never
sees it because the asan jobs set `ASAN_OPTIONS: detect_leaks=0` job-wide while
`add_test(contour_gui_test ...)` sets no `LSAN_OPTIONS` of its own — so the failure
is local-only. Note also that the reported total varies run to run (79002 … 102063
bytes) because Catch2 randomises test order. Tracked separately from this PR.

## Two notes for anyone extending the QML

An inline `component Foo: ...` is **not** usable as a `Repeater` delegate here — it
instantiates nothing, silently, which rendered every segment empty. Hence the
separate component files.

`Repeater` delegates are not created at all in the QML test harness — even
`Repeater { model: 3; delegate: Rectangle {} }` yields none, reporting the right
`count` beside a null `itemAt(0)`, with or without a `QQuickWindow` and with an
incubation controller installed. So the tests drive the signals the delegates would
have invoked; see `IncubationDriver` in `SettingsPageQml_test.cpp`.


christianparpart (approved):
thank you, @Yaraslaut

PR #2029, opened 2026-07-27
Fix crash when a pane closes: resizing an already-closed ConPTY

Yaraslaut:
## Summary

Closing a pane crashed contour on Windows with an access violation reading `0xffffffffffffffff`:

```
conpty!_ResizePseudoConsole+0xb [inlined in conpty!ConptyResizePseudoConsole+0x3a]:
    mov rcx,qword ptr [rcx]   ds:ffffffff`ffffffff=????????????????
```

Found by running `contour.exe` under CDB and hitting it on an ordinary shell exit.

**Cause.** `ConPty::close()` invalidates `_master` but leaves `_slave` alive. `ConPty::resizeScreen()` guarded on `if (!_slave)`, so after a close the guard did not fire and `INVALID_HANDLE_VALUE` was passed to `ResizePseudoConsole`, which dereferences it. That guard was written to separate *not started yet* from *running*; the third state — *started, then closed* — fell through to the running branch.

**How it is reached.** Closing a pane makes QML re-run the `Loader` binding for the surviving panes, which re-enters `setSession()` → `attachDisplay()` → `Terminal::resizeScreen()` while `onClosed()` is still unwinding — i.e. on an already-closed PTY:

```
conpty!ConptyResizePseudoConsole
vtpty::ConPty::resizeScreen              src/vtpty/ConPty.cpp
vtbackend::Terminal::resizeScreen
contour::TerminalSession::attachDisplay  src/contour/TerminalSession.cpp
contour::display::TerminalDisplay::setSession
  … QML Loader re-instantiating the pane …
contour::TerminalSessionManager::paneClosed
contour::TerminalSession::onClosed
```

**Fix**, at both ends of that path:

- `ConPty::resizeScreen()` now tests the master handle through the existing `isClosed()` predicate rather than `_slave`. That covers the pre-start *and* post-close cases in one branch, and matches `UnixPty`, whose `_masterFd < 0` check already lands in the record-the-size branch after a close. `UnixPty` was therefore never affected — `ConPty` was the outlier.
- `TerminalSession::attachDisplay()` skips the resize when the session is already closed, so a dead session stops pushing geometry at a gone device on every platform. Kept as well as the `vtpty` guard because `attachDisplay()` is not the only caller of `resizeScreen()`.

## Testing

New `src/vtpty/ConPty_test.cpp` (Windows-only, wired into `vtpty_test`):

- `ConPty.resizeScreen.beforeStart` — the pre-start branch still records the size.
- `ConPty.resizeScreen.afterClose` — the regression. Start, resize live, close, resize again.
- `ConPty.resizeScreen.afterRepeatedClose` — double close, then a resize.

New GUI test in `DisplayRendering_test.cpp` covering the whole path: a real display re-attached to a real closed session must push no resize down to the device. `BlockingMockPty` gains a `resizesAfterClose()` counter so the assertion is that the resize is *absent*, not merely harmless.

Both tests were confirmed to fail without the fix and pass with it:

- Unit test without the fix: `SIGSEGV`, exit `0xC0000005` — the production crash, reproduced in the runner.
- GUI test without the fix: `CHECK( h.pty->resizesAfterClose() == resizesBeforeReattach )` → `1 == 0`.

Full suite on `clangcl-release`, build clean under `-Werror`:

```
crispy_test, text_shaper_test, vtpty_test, vtparser_test,
vtbackend_test, vtconformance_test, vtmux_test, vtrasterizer_test .. Passed
contour_gui_test ... 656/657
```

The one `contour_gui_test` failure (`display: the deferred screenshot readback delivers a real image`) is **pre-existing and unrelated** — a temp-PNG file lock. Verified by stashing this fix and reproducing the identical failure on a clean tree.

**Open issue — not yet attributed.** A run of the suite via `ctest` ended in a segfault in `contour_gui_test` that does *not* occur when the same binary is run standalone (clean there, apart from the screenshot failure above). I have not isolated it, so I cannot yet say whether it is related to this change. Flagging it rather than assuming it is unrelated; this is the main reason the PR is a draft.

## Manual test checklist

- [ ] Attribute or rule out the ctest-only `contour_gui_test` segfault above.
- [ ] On Windows, split a pane and close it — confirm no crash (the original repro).
- [ ] Exit the shell in the last pane and confirm clean teardown.
- [ ] Sanity-check on Linux that pane close/split still behaves (`UnixPty` path unchanged, but `attachDisplay()` is shared).

### DIFF
--- a/src/contour/TerminalSession.cpp
+++ b/src/contour/TerminalSession.cpp
@@ -373,6 +373,12 @@ void TerminalSession::attachDisplay(display::TerminalDisplay& newDisplay)
     // cursorPositionChanged() early-return, freezing IME rectangle tracking and the a11y caret.
     _cursorMovedPostPending.clear(std::memory_order_release);
 
+    // A closed session has nothing to resize: its PTY is gone. Closing a pane makes QML re-run the
+    // Loader binding for the surviving panes, which re-enters setSession() -> attachDisplay() while
+    // onClosed() is still unwinding, so this path really is reached after the PTY has been closed.
+    // Skipping the resize keeps a dead session from pushing a geometry change down to it; ConPty
+    // additionally refuses the resize on its own side, since attachDisplay() is not the only caller.
+    if (!isClosed())
     {
         // NB: Inform connected TTY and local Screen instance about initial cell pixel size.
         auto const l = scoped_lock { _terminal };
--- a/src/contour/test/DisplayRendering_test.cpp
+++ b/src/contour/test/DisplayRendering_test.cpp
@@ -1412,6 +1412,50 @@ TEST_CASE("display: a session re-bound onto a resized display adopts the live gr
     second.reset();
 }
 
+TEST_CASE("display: attaching a display to an already-closed session does not resize its dead PTY",
+          "[display][session][close]")
+{
+    // Regression pin for the ConPTY teardown crash: closing a pane makes QML re-run the Loader
+    // binding for the surviving panes, which re-enters TerminalDisplay::setSession() ->
+    // TerminalSession::attachDisplay() while onClosed() is still unwinding. attachDisplay()
+    // unconditionally called Terminal::resizeScreen(), so the resize reached a PTY that had already
+    // been closed. On Windows that is fatal rather than merely pointless: ConPty::close()
+    // invalidates the HPCON but leaves the slave alive, so the resize passed INVALID_HANDLE_VALUE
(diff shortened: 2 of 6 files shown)


christianparpart:
@Yaraslaut thanks for finding this.
I have rebased your branch on top of latest master and fixed CI (ideally).

christianparpart:
p.s.: Please install `typos` locally, to perform spell-checking locally.

christianparpart (approved):
Thanks, @Yaraslaut

PR #2038, opened 2026-07-31
fix(build): make GuiPrivate conditional on Qt >= 6.10

wineee:
## Description

Qt 6.8/6.9 packaging does not expose GuiPrivate as a separate find_package component. Gate all GuiPrivate usage behind a version check so the build succeeds on older Qt without it.

## Checklist

- [ ] I have updated the documentation accordingly
- [ ] Breaking changes (if any) are documented


christianparpart (approved):
Thank you very much, @wineee

PR #2041, opened 2026-08-04
Fix mangled text rendering in split panes

### LINKED ISSUE #2040: Some render issues
### Contour Terminal version

0.7.0-master-b20087bc

### Installer source

GitHub: Source code

### Operating System

Windows

### Architecture

x86-64 / amd64

### Shell

None

### GPU & Renderer

_No response_

### Other Software

_No response_

### Steps to reproduce

When we have multiple panes and resize window, the text in the left panel is getting weirdly rendered:

Image

Image

### Expected Behavior

Font rendering is the same on left and right 

### Actual Behavior

the left panel font rendering is broken

### Configuration excerpt

```yaml

```

### Additional notes

_No response_

christianparpart:
With two split panes open, resizing the window left one pane's text visibly damaged: stems losing their bright column for a scanline or two while neighbouring glyphs in the same row and the same frame were untouched (#2040). The reporter's description — "attenuated, not deleted", the glyph's mask intact but its weight wrong — turned out to be exactly right, and is what finally identified the mechanism.

## Root cause

`PaneNode.qml` sized a split child as `SplitView.view.width * ratio`, **unrounded**. The first pane therefore got a fractional width and the second a fractional **origin**.

That origin reaches the renderer by two paths that round **differently**:

- the vertex transform composes it *unrounded* — `TerminalRenderNode::prepare()` → `composeItemToClip()`
- the scissor *rounds* it — `qRound(originScene.x() * dpr)`, in the same function

Geometry and clip then disagree by up to half a pixel. Because the glyph atlas is sampled with `QRhiSampler::Nearest`, the sample points land on texel *boundaries* rather than texel centres, and a stem drops its bright column for a scanline or two. Which glyphs are hit depends on where each one falls relative to that half-pixel disagreement — which is why the damage never looked like a whole-pane transform error, and why two of three *identical* `l` glyphs could be damaged while the middle one was perfect.

**The fix** snaps the split child so the pane boundary lands on a whole **device** pixel. Snapping in logical coordinates is not enough at a fractional DPR: the hardware pixel is the quantity both paths agree about. The rule is extracted to `contour::geometry::snapPaneExtentToDevicePixels()` so it is unit-testable without QML; the QML binding computes it inline because a property binding cannot call a free C++ function, and the header says so.

## Verification

A two-pane split was dragged continuously across the 986–1010px band while a background runspace captured the **composited** window. Contour's own readback cannot show this class of defect: `scheduleScreenshot()` replays draws into an item-sized *offscreen* target where `offscreen=true` skips the window-space node clip entirely — 546 in-process frames were byte-identical while the artifact was plainly visible on screen.

Grouping captures by separator position (so pane geometry is fixed within each group):

| | before | after |
|---|---|---|
| Corrupted frames | ~50% at **every** separator position | **0 / 1166** |
| Per-stem ink | `[2043, 2656, 2043]` — two of three identical `l` glyphs losing 23% | uniform |

## Also in this PR

Independent defects found while investigating. **None of them is the artifact** — it reproduced with all of them already applied — but each is real:

- **Four discarded DirectWrite `HRESULT`s** (`rasterize()`). Two are worse than bad pixels: `CreateGlyphRunAnalysis()` leaves the `ComPtr` **null** and the next line dereferences it, in both the monochrome and colour paths. `GetAlphaTextureBounds()` leaves a zero `RECT` that yields a 0×0 bitmap indistinguishable from an empty glyph. `MoveNext()` leaves `haveRun` **uninitialized**, so the enumeration loop branches on indeterminate memory. Plus `factory2` and `glyphRunEnumerator` were raw interfaces **leaked once per glyph rasterization**.
- **GUI debug logging was silently discarded.** `contour terminal debug <tags>` produced *no* output for *any* tag: `logstore::sink::console()` is constructed **disabled**, and `configure()` only enables categories. `log`/`log-file` avoid this via `installLogging()` → `scoped_output`, but `terminal` declared only `debug`, so that path was unreachable. On Windows shell redirection is not a workaround either — `tryAttachConsole()` `freopen`s stderr onto `CONOUT$`, replacing a `2>` redirect before the first line is written. `terminal` now accepts `log-file`.
- **Four rendering invariant probes** (`vt.rasterizer.tilebounds`, `gui.display.geometry`, `gui.display.sampling`, `gui.display.clip`), off by default and silent unless an invariant breaks. Each was validated by feeding it a deliberate fault and confirming it fires — a probe that reads clean is only evidence if it is known to be *capable* of reporting.
- **`devicePixelScaleError()`** plus six unit tests, pinning the 1:1 device-pixel mapping the Nearest-sampled atlas depends on.

## Hypotheses investigated and disproven

Recorded so they are not re-tread; each was excluded with a self-tested probe, not deprioritised:

- **`yOverflow`'s dead crop** in `TextRenderer.cpp` — computed and only ever logged. Acting on it **breaks** rendering: `position.y` is already baseline-relative (FreeType `bitmap_top`), so an ordinary `A` measures `yOverflow = 7`. The expression itself is wrong, which is likely why it was never wired up.
- **`text::scale()`'s box filter** — ink is conserved to within one unit, including the edge-overrun case.
- **Atlas tile bleed, item→clip scale, quad/texel sampling ratio, scissor coverage** — all measured clean across a full repro by probes verified to fire.
- **The glyph rasters** — ASCII glyphs are direct-mapped, rasterized *once* at startup and never re-rasterized during a resize.

Note that a **fractional pane origin was investigated and disproven earlier on this branch** on the argument that a pure translation cannot drop a texel under Nearest sampling. That argument is correct *in isolation* and is now pinned as a unit test. What it misses is that the origin is consumed twice with different rounding — the translation is not applied consistently, so the two paths disagree. The earlier conclusion was right about the mechanism it examined and wrong about the conclusion drawn from it.

## Worth flagging for review

- The QML binding duplicates `snapPaneExtentToDevicePixels()`'s arithmetic inline, because a QML property binding cannot call a free C++ function and exposing it would mean a `QObject` singleton for one expression. The header names itself as the specification; the two must stay in step by hand.
- `contour_gui_test` has one **pre-existing** failure on this machine (`SessionInput_test`, a keyboard-modifier encoding mismatch). It predates this branch and touches no rendering code.
- The PR's own `#2040` readback test **skips** where no software OpenGL context is available, which is the case on the reporter's platform — so that safety net was not running for either of us.

Fixes #2040


christianparpart:
@Yaraslaut  can you confirm this is fixing it in your environment, too?

Yaraslaut:
It is better, but small artifact still remains:
image

christianparpart:
Thanks — that screenshot was very useful. I analyzed it pixel by pixel, and it narrows things down a lot, though not all the way.

**The good news:** placement and scale are now correct. Cell pitch is a uniform 10px, every glyph sits on the same baseline, and the `r` and `n` stems hold a centre of mass of 25.68 / 55.68 — identical to two decimal places on every single scanline. (The stray ink at the right edge is just the next word’s `a`, cropped — I checked, it is byte-identical to the `a` in `personal`.)

**The remaining defect is confined to the `l`.** Its stem drops from 308 to 93 units of ink for exactly two scanlines and then recovers. Ink is not conserved, so this is genuine corruption rather than a slanted outline.

That is the puzzling part: `r` and `n` are pixel-perfect **in the same pane, same row, same frame** as the broken `l`. That rules out everything those three glyphs share — the item→clip transform, the quad geometry, the `Nearest` sampler, and the atlas configuration.

I did chase one hypothesis and want to record that it is **wrong**, so nobody re-treads it: a fractional pane origin (`PaneNode.qml` sizes a split child as `view.width * ratio`, unrounded). It cannot be the cause. A glyph quad spans exactly as many device pixels as its tile has texels, so under Nearest the sample points step one texel per pixel *whatever* the offset — a pure translation cannot drop a texel. I verified this both in simulation and by rendering a 1px stem at origins 4.0 and 4.5 and at DPR 1.5 with snapping deliberately disabled: identical, intact output. The PR body now states that precondition explicitly instead of the bare claim.

Two unrelated defects I did find while verifying are pushed here:
- the page fit scaled its margins by the *content scale* instead of the device-pixel ratio (wrong under KDE’s forced font DPI; dates to 2022, so it has shipped);
- `TerminalRenderNode::prepare()` read `QWindow::size()` live on the render thread with the GUI thread unblocked, so mid-resize it could derive the wrong ratio for a frame.

**To get to the bottom of the `l`, could you help with three things?**
1. A **lossless 1:1 crop** (PNG, no scaling) of the artifact — the current one is small enough that I am partly inferring subpixel weights.
2. Which **font** and size, and whether subpixel/ClearType antialiasing is on.
3. Is it **transient or sticky** — does forcing a redraw (scroll the line away and back, or resize once more) clear it, or does it persist?

My current suspicion is a per-glyph path rather than anything frame-wide: `restrictToTileSize()` and `sliceTileData()` both `break` mid-loop on a bounds failure and leave the remaining rows zero-filled while the bitmap still claims full height, so a partially-written tile gets uploaded rather than rejected. Two lost scanlines is exactly that shape. But I have not been able to reproduce it locally, so I do not want to guess at a fix.

christianparpart:
One more defect found while investigating, now pushed as `8a829465`.

`atlas::sliced()` — the iteration a wide bitmap is cut along — was wrong for any bitmap whose width is not a whole number of tiles. It advanced `endX` by a full `tileWidth` per step and stopped at `width + (width % tileWidth)`, which is neither the width nor a slice boundary:

- **width 20, tile 8** → sentinel 24, so the last slice spans `[16,24)` over a 20-wide bitmap. `sliceTileData()` then reads past the row; its `SoftRequire` breaks out **mid-copy**, leaving the remaining rows zero-filled while the tile still claims full height. A partially-written tile gets uploaded rather than rejected — i.e. *missing scanlines in whatever that tile draws*.
- **width 17, tile 8** → sentinel 18, which the sequence 0, 8, 16, 24… never equals. The iterators compare `beginX`, so **the iteration does not terminate**.

`endX` is now clamped to the bitmap width, which makes `beginX` land exactly on the width and lets the sentinel simply *be* the width. A zero tile width yields the empty range instead of spinning.

Two caveats, so this is not oversold:

1. **It is latent, not live.** The only present caller is `CursorRenderer`, which slices a bitmap of `cellSize.width * columnWidth` along `cellSize.width` — always an exact multiple, so neither branch is reachable today. I am *not* claiming this explains the remaining artifact.
2. It had **no test coverage at all**, which is how both bugs survived. It has now — and I verified the tests genuinely fail against the old code (3 failures, including the non-terminating case hitting the runaway cap) rather than merely passing against the new.

Worth noting given `SoftRequire` **aborts in debug builds but only logs in release**: if a path like this ever does trip in the wild, a release user sees silent corruption while our debug tests would have crashed — which is a plausible shape for "reproduces for the reporter, never in CI".

The questions in my previous comment still stand — a lossless 1:1 crop, the font, and whether the artifact is transient or sticky would each help a lot.

christianparpart:
@Yaraslaut please try again.

christianparpart:
## Status on the remaining `l` artifact

I have not fixed it, and I want to set out precisely what is and is not established, so the next person does not repeat my dead ends.

### Established from the screenshot

Summing each scanline of the stem across its **full** stroke width (not just the brightest column), background-subtracted:

```
l stem            r stem (control)
row 5  316        row  9  335
row 6  316        row 10  334
row 7  316        row 11  334
row 8  313        row 12  334
row 9   96  <--   row 13  334
row10   96  <--
row11  223
row12  316
```

Two scanlines lose ~70% of their ink and then recover exactly. The control glyph is flat to within one unit. So **ink is not conserved** — this is real corruption, not a slanted stroke redistributing weight between columns (I checked that reading and it does not survive the numbers).

The ink *mask* stays continuous, though: rows 9-10 still have ink at the same two columns, just dimmer. Nothing is structurally missing. Whatever this is, it attenuates rather than deletes.

### Ruled out, with reasons

Each of these was investigated and **excluded**, not merely deprioritised:

- **Fractional pane origin.** A quad spans exactly as many device pixels as its tile has texels, so under `Nearest` the sample points step one texel per pixel *whatever* the offset. Verified by simulation and by rendering a 1px stem at origins 4.0 / 4.5 and at DPR 1.5 with snapping deliberately off — identical, intact output. I built the fix, revert-checked it, and reverted it. The claim already in this PR body was correct.
- **Anything shared by the glyphs in that run.** `r` and `n` hold a centre of mass of 25.68 / 55.68 to two decimals on every scanline, in the same pane and frame as the broken `l`. That excludes the item→clip transform, the quad geometry, the sampler and the atlas configuration.
- **Direct-mapped vs LRU path asymmetry.** `l`, `r` and `n` are all ASCII 0x21-0x7E, so all three take the direct-mapped path.
- **Stale direct-mapped tiles across an atlas reconfigure.** `configureTextureAtlas()` constructs a brand-new `TextureAtlas`, so `_directMapping` starts zero-filled and every glyph re-rasterizes.
- **`createSlicedRasterizedGlyph` and `restrictToTileSize` copy loops.** Both clamp correctly (`min(xOffset + tileWidth, bitmapSize.width)`; `targetPitch < sourcePitch` guarded up front), so their `break`-leaves-rows-zero-filled paths are unreachable.

### Two real defects found along the way, both pushed

- `atlas::sliced()` — out-of-bounds final slice, and **non-terminating** for some widths (`c6f40fa9`). Latent: the only caller always slices an exact multiple.
- The page fit scaled its margins by the content scale instead of the device ratio (`ae8696aa`), and `prepare()` read `QWindow::size()` off the render thread (`47771b7a`).

Neither explains the `l`. I am not claiming otherwise.

### What would actually move this forward

@Yaraslaut — three things, in rough order of usefulness:

1. A **lossless 1:1 PNG crop** (no scaling). The current one is 97x23 and I am partly inferring subpixel weights from it.
2. The **font family and size**, and whether subpixel/ClearType AA is on.
3. Is it **transient or sticky**? Does scrolling the line away and back, or one more resize, clear it — or does it persist?

Also useful if easy: does it reproduce with `CONTOUR_LOG=vt.renderer.rasterizer` set? A `Cutting off oversized ... tile` line, or a `Precondition failed:` from `SoftRequire`, would be decisive.

Worth flagging for the maintainers regardless: **`SoftRequire` aborts in debug but only logs in release** (`crispy/assert.h:134-148`). Any tripped bound is silent corruption for a release user and a hard crash for us — a plausible explanation for "reproduces for the reporter, never in CI", and arguably a policy worth revisiting.

christianparpart:
## A candidate that finally matches the measurement — pushed as `6c67a007`

@Yaraslaut this one is worth your attention, because it is Windows-only and it fits the numbers.

### The measurement, restated

Per-scanline ink of the damaged stem, summed across its **full** stroke width, background-subtracted:

```
l stem            r stem (control, same pane/row/frame)
row 5  316        row  9  335
row 6  316        row 10  334
row 7  316        row 11  334
row 8  313        row 12  334
row 9   96  <--   row 13  334
row10   96  <--
row11  223
row12  316
```

Two properties, and both matter:

1. **Ink is not conserved** — 96 against a steady 316. So this is not a stroke leaning between columns.
2. **The mask is intact** — rows 9-10 still have ink at the same two columns, just dimmer. Nothing is structurally missing.

"Attenuated, not deleted" is a narrow signature. It rules out dropped texels, dropped rows, and misplacement — all of which delete.

### What produces exactly that

`directwrite_shaper.cpp`, `renderGlyphRunToBitmap()`:

```cpp
auto hr = _glyphAnalysis->CreateAlphaTexture(
    DWRITE_TEXTURE_CLEARTYPE_3x1, &_textureBounds, tmp.data(), tmp.size());
// hr assigned, never read
```

`hr` was assigned once and never checked. On failure DirectWrite leaves the destination buffer untouched — and `tmp` is zero-initialized — so a swallowed failure did **not** skip the glyph. The copy loop below carried those zeros out as if they were coverage.

In an RGB subpixel bitmap a zero run is not a hole. It is **attenuated ink**: shape and mask preserved, weight lost across the affected span. That is property (1) and property (2) simultaneously.

`renderGlyphRunToBitmap()` now returns whether it produced the texture, and `rasterize()` maps failure to `std::nullopt` so the glyph is skipped rather than drawn faded. It also logs to `errorLog`, which is **enabled by default** — so if this is your bug, a plain release build will now print `directwrite: CreateAlphaTexture failed for a NxM glyph run.` when it happens.

### Being honest about the confidence

I cannot run DirectWrite, and neither can our CI shapers on Linux or macOS — which is itself consistent with a defect that reproduces for you and never for us. So this is **a candidate, not a confirmed fix**. A discarded HRESULT on the pixel path is a real defect either way, which is why it is pushed regardless.

**What would settle it:** build this branch and check whether the artifact still occurs, and whether that `directwrite:` line appears in the log when it does.

- If the artifact is gone → this was it.
- If the line appears but the glyph still looks wrong → we are in the right function, wrong branch.
- If neither → my diagnosis is wrong and the lossless 1:1 crop plus font name from my earlier comment become the way forward.

Any of those three outcomes is useful, so please do not worry about "wasting" the result.

Yaraslaut:
the issue still remains, here is the screenshot:
image


here is the log from the font shaper
'''
[2026-08-06 14:58:43.766620] [font.textshaping] Shaping using font key: 0, text: "x64-windows-dbg"
[2026-08-06 14:58:43.766639] [font.textshaping] Shaping codepoints (Text): 0:U+78 1:U+36 2:U+34 3:U+2d 4:U+77 5:U+69 6:U+6e 7:U+64 8:U+6f 9:U+77 10:U+73 11:U+2d 12:U+64 13:U+62 14:U+67
        Using font: key=0, path="C:\WINDOWS\FONTS\MONASPICEKRNERDFONT-REGULAR.OTF"
[2026-08-06 14:58:43.766726] [vt.rasterizer] Shaped codepoints (Text): x64-windows-dbg (U+0078/0 U+0036/1 U+0034/2 U+002D/3 U+0077/4 U+0069/5 U+006E/6 U+0064/7 U+006F/8 U+0077/9 U+0073/10 U+002D/11 U+0064/12 U+0062/13 U+0067/14)
        with metrics: (89+0+0|9+0) (23+0+0|9+0) (21+0+0|9+0) (14+0+0|9+0) (11197+0+0|9+0) (11176+0+0|9+0) (79+0+0|9+0) (69+0+0|9+0) (80+0+0|9+0) (88+0+0|9+0) (84+0+0|9+0) (14+0+0|9+0) (69+0+0|9+0) (67+0+0|9+0) (72+0+0|9+0
'''

also, this corruption is happenin in between the cell size resize, it looks like there is 1 ticket between resize happening, and this corruption is happening in this state, something like the following resizes

'''
[2026-08-06 15:01:17.640552] [gui.display] Applying resize 670x600/670x600 pixels (margins 0x+0y) and 67x30 -> 67x30 cells.
[2026-08-06 15:01:17.640578] [gui.display] No resize necessary. New size is same as old size of 67x30.
[2026-08-06 15:01:17.641172] [gui.display] Setting render target size to 671x600.
[2026-08-06 15:01:17.641253] [gui.display] Setting render target size to 677x600.
[2026-08-06 15:01:17.641316] [gui.display] Setting render target size to 676x600.
[2026-08-06 15:01:17.641459] [gui.display] Setting render target size to 670x600.
[2026-08-06 15:01:30.595205] [gui.display] Display size changed to 677x600 virtual (677x600 device).
[2026-08-06 15:01:30.595291] [gui.display] Applying resize 670x600/677x600 pixels (margins 0x+0y) and 67x30 -> 67x30 cells.
'''

Yaraslaut:
@christianparpart  i pushed my commits in the branch with the fix

christianparpart:
@Yaraslaut i've been applying some `/code-review` on this branch. Would you mind please re-verifying that it did not accidentally broke you ?

Yaraslaut:
@christianparpart tested, everything is working

christianparpart:
@Yaraslaut one last time, please. :-)

Yaraslaut:
@christianparpart  tested, everything works

Issue #2043, opened 2026-08-05
[Feature proposal] Non-default Cocoa GUI for macOS


barracuda156:
@christianparpart Hi Christian!

Will you be willing to accept a PR implementing Cocoa front-end as a [non-default] alternative to Qt6?

I have it working on my PowerPC Mac now:

<img width="1037" height="999" alt="Image" src="https://github.com/user-attachments/assets/2233b3b7-5553-4097-911c-b6e9783af38a" />

christianparpart:
Hey @barracuda156. This looks amazing.
I think it's definitely not worthless at least creating the PR. It is actually hard to say if it makes sense to accept the PR (or as-is) though. I am personally not against it, but it is certainly a question of maintainability.
I have no PowerPC. I'm only having an Apple M1 macOS here. We should at least also enable CI to automatically test the build and run our unit/e2e tests on it.

Maybe, with that PR living as an open PR for a while, we can figure out how to approach this.

barracuda156:
@christianparpart Let me see if it works on Apple Silicon on my end (on my desktop MacMini it does not build, since Catalina does not support C++23 in `libc++`; ironically, I know). I obviously bothered with this primarily to have a modern terminal on powerpc, but it still makes some sense on modern systems, allowing to avoid Qt6 (which is just huge, even when it builds fine).

christianparpart:
@barracuda156 how do you get a modern C++ toolchain on such an old computer? What toolchain are you using anyways? I'd prefer to stick to C++23 and upwards rather than going back in time (as far as C++ language/library standards are concerned)

barracuda156:
@christianparpart C++23 is intact for everything in your original code and whatever C++ sources I add. There is one exception for _added_ ObjC++ sources (two .mm files): they are compiled with C++98 with a custom build command, due to https://github.com/iains/gcc-14-branch/issues/25 (remains relevant for gcc-16).

I use my toolchain, but in effect it is https://github.com/iains/gcc-16-branch and https://github.com/iains/darwin-xtools written into MacPorts infrastructure. CMake is standard (I keep my port, but upstream MacPorts one works fine).

barracuda156:
P. S. To be specific, this is all toolchain:

https://github.com/macos-powerpc/powerpc-ports/blob/084dbf1790f2ebd936fd7145f4a643aa409cf235/lang/gcc16/Portfile
[...]

PR #2044, opened 2026-08-05
Stop a GUI-less build requiring Qt Quick and a macOS app bundle

### LINKED ISSUE #2042: Regression: Qt6 pulled in outside of GUI again
### Contour Terminal version

eb4e5896b89ad8c989a86798d92f83c45ada1564

### Installer source

GitHub: Source code

### Operating System

macOS

### Architecture

Other (please specify below)

### Shell

None

### GPU & Renderer

_No response_

### Other Software

_No response_

### Steps to reproduce

See commit https://github.com/contour-terminal/contour/commit/eb4e5896b89ad8c989a86798d92f83c45ada1564

### Expected Behavior

Qt6 should only be required where it is actually needed.

### Actual Behavior

https://github.com/contour-terminal/contour/commit/eb4e5896b89ad8c989a86798d92f83c45ada1564 has put Qt6 requirement outside of GUI build.

### Configuration excerpt

```yaml

```

### Additional notes

_No response_

christianparpart:
Reported against `eb4e5896`, which was not the cause: that commit only moved a `find_package(Qt6 COMPONENTS Core REQUIRED)` above the `CONTOUR_FRONTEND_GUI` gate so `Qt6_VERSION` was set before the GuiPrivate version check. The `else()` branch already asked for `Core`, so a GUI-less build had always required Qt.

This makes `-D CONTOUR_FRONTEND_GUI=OFF` need **no Qt at all**, so `contour daemon` and the CLI verbs can be built on a platform Qt6 was never ported to — which is what the reporter is doing on macOS 10.6 / PowerPC (see #2043).

## Changes

**`crispy::environment` becomes an injectable interface.** It was a namespace of free functions over an immutable snapshot — an ambient global reached without an abstraction. It is now a class with two implementations: `snapshot_environment` (today's semantics, which all 25 existing call sites keep) and `live_environment`, which reads afresh because config loading expands `${VAR}` against the environment as it stands at that moment, while the PTY threads are live. `defaultEnvironment()`/`defaultLiveEnvironment()` are the documented escape hatch, exactly as `net::defaultSteadyClock()` is.

**The shell-integration scripts are embedded as C++ data.** `contour generate integration` read them out of Qt's resource system, which is the single reason a CLI-only build linked Qt. They stay ordinary files under `shell-integration/` — they are what users source, and they are still installed to `share/contour/shell-integration/`; only the in-binary copy is generated, by a CMake script so nothing has to be built for the host first. `CONTOUR_SHELL_INTEGRATION_SHELLS` now drives the data, the dispatch and the help text alike, which also fixes a help string that had claimed "Supported shells: fish, zsh, tcsh" ever since bash support landed.

**The last three Qt uses go.** `main.cpp`'s message handler exists only to tidy up what Qt prints, and is compiled out with it; `Config.cpp` uses `crispy::live_environment` instead of `qgetenv()` and libunicode instead of `QString::toUcs4()`; `ContextMenuModel` — the QML-facing adapter over the Qt-free decision layer in `ContextMenu.h` — moves to the GUI-only source lists.

**The build stops asking.** `find_package(Qt6)`, `qt_add_resources`, AUTOMOC, AUTORCC, `qt_add_executable`, `QT_DISABLE_DEPRECATED_BEFORE`, the `Qt6::Core` link and the top-level `include(FindQt)` are all gated. Windows gets the treatment macOS got in `fdc80c52`: `windeployqt` and `DeployQt.cmake`'s `find_package(Qt6Core REQUIRED)` are GUI-only, so a GUI-less Windows build installs the plain `bin/contour.exe`.

## How it is locked

Both headless CI jobs configure with `-D CMAKE_DISABLE_FIND_PACKAGE_Qt6=ON`, which makes every Qt lookup fail. They install Qt first on purpose: the assertion worth making is not that Qt is absent from the runner but that the build never reaches for it. The Linux job additionally greps `ldd`, since a transitive `Qt6::Core` would satisfy the configure-time gate and still land in the binary.

`contour_e2e_shell_integration` is new and registered for **both** frontend configurations: it drives the real CLI verb and compares its output against the files in `shell-integration/`, so a generator that dropped a byte or mangled a line ending is found by CI rather than by a user whose shell broke.

Verified locally: Windows and Linux headless builds configure, build, install and run with Qt fully unavailable (`ldd` shows no `libQt*`, and the binary runs with Qt off `PATH`); the GUI build is unchanged at 748 cases / 11214 assertions, +4 cases and +35 assertions being the new tests.

Fixes #2042


barracuda156:
@christianparpart Thank you very much!

PR #2049, opened 2026-08-07
sixel: clamp raster aspect ratio to stop a wraparound OOB write

prithveerarya345:
While reading through the Sixel parser I noticed `setRaster()` takes the
`Pan`/`Pad` values straight off the wire and turns them into `_aspectRatio`
with no upper bound. That value later drives `y = y0 + bit*_aspectRatio` in
`render()`/`bandRows()`, and the bounds check right after it
(`y + _aspectRatio > canvasHeight`) is 32-bit arithmetic, so a large enough
`_aspectRatio` makes that addition wrap around to something small while `y`
itself stays a huge (but perfectly valid) `unsigned`. The check passes, and
`y` goes straight into a pointer offset in `paintBit()` -- a write far past
the actual pixel buffer.

Concretely: `ESC P q "2147483904;1 A ESC \` sets `Pan=2147483904, Pad=1`
(aspect ratio ~2^31), then a single sixel character with bit 1 set walks
straight off the end of `_buffer`.

Fix is two parts:
- Clamp `_aspectRatio` to 255 in `setRaster()` -- real encoders use single-digit
  ratios, so this doesn't change behavior for any legitimate stream.
- Do the `y`/`lastRowExclusive` bounds-check arithmetic in 64 bits in both
  `render()` and `bandRows()`, so the same overflow class can't reopen if a
  future caller of `setRaster()` skips the clamp.

Added a regression test (`SixelParser.raster_aspect_ratio_wraparound_is_clamped`)
covering the exact PoC above.

Full disclosure: I used Claude to help audit this code and work through the fix
with me, not just my own eyes on it. Verified the overflow and the resulting
diff by hand and built/tested it locally under ASan before opening this.

### DIFF
--- a/src/vtbackend/SixelParser.cpp
+++ b/src/vtbackend/SixelParser.cpp
@@ -416,8 +416,15 @@ void SixelImageBuilder::newline()
 void SixelImageBuilder::setRaster(unsigned int pan, unsigned int pad, optional<ImageSize> imageSize)
 {
     if (pad != 0)
-        _aspectRatio =
-            max(1u, static_cast<unsigned int>(std::ceil(static_cast<float>(pan) / static_cast<float>(pad))));
+        // Clamped to keep every y0 + bit*_aspectRatio computation below (render(), bandRows()) well
+        // inside unsigned range: an attacker-supplied Pan close to UINT_MAX with Pad=1 would otherwise
+        // let (y + _aspectRatio) wrap around to a small value in the bounds check while y itself stays
+        // huge, producing a heap write far past _buffer. Real encoders never need more than single-digit
+        // ratios; 255 leaves generous headroom without reopening the overflow.
+        _aspectRatio = clamp(
+            max(1u, static_cast<unsigned int>(std::ceil(static_cast<float>(pan) / static_cast<float>(pad)))),
+            1u,
+            255u);
     _sixelBandHeight = 6 * _aspectRatio;
     if (imageSize)
     {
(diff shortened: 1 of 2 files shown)


christianparpart:
Oh nice. Many thanks for your contribution, @prithveerarya345.
I'll digg into it as soon as possible. I'll include this PR also into the imminent next release. :)

j4james:
Just FYI, I think the official maximum for the aspect ratio is the screen height (in pixels) divided by 6, because anything larger than that would mean a sixel row couldn't fit on screen. And on a real terminal, it will trigger a scroll as soon as the the active row extends beyond the bottom of the screen, so a sixel row that doesn't fit would be impossible to render.

And note that when margins are involved, it'll be the height of the margin area divided by 6.

christianparpart:
@prithveerarya345 FYI: I've just rebased your PR on top of latest origin/master. That should also help getting it green.

christianparpart (approved):
Thank you very much, @prithveerarya345 😄

prithveerarya345:
Thanks for the follow-up, sorry you had to write it.

@j4james was right and I should have acted on it before merge — canvasHeight/6
is the real bound. My 255 was too loose for 800x600 and too tight for anything
over 1530px. I only ever thought of it as an overflow guard, never as what a
display can actually show.

The float cast is the worse one. Clamping the result of
`(unsigned)ceil((float)pan/(float)pad)` leaves the UB sitting before the clamp,
so Pan=UINT_MAX still went two ways by arch. My test picked 2147483904, which is
float-exact and under UINT_MAX, so it missed that case completely. Integer ceil
with STATIC_CHECK over the domain is the right fix.

Also walked straight past the 32-bit height multiply and the double-counted
ratio, in the same two functions I was editing. I'll write the changelog entry
myself next time.

christianparpart:
No worries, @prithveerarya345. I also commit and push bugs / bad behaviours, and then regret it. It is though important to just react as soon as kind of possible, when one finds out that there was something broken pushed (depending on the severity, of course) 😄

PR #2056, opened 2026-08-14
Restore the window drop shadow, and give menus one too

christianparpart:
Contour's window is frameless by default so its own tab bar can serve as the title bar. A frameless window gets no decoration from the system — and on KDE Plasma the drop shadow *is* the decoration, so the shadow disappeared when `show_title_bar` started defaulting to `false`.

It was never KDE-specific. The same `Qt::FramelessWindowHint` was applied on every OS, so Windows also lost its DWM shadow, its Windows 11 rounded corners and Aero snap, and macOS kept a shadow but squared off its corners and let it go stale after a resize.

The window now publishes a shadow of its own to KWin on X11 and Wayland, sized and shaped like Breeze's so it matches the rest of a Plasma desktop. The compositor composites it *outside* the window surface, which is why this route was taken rather than drawing into a client-side margin: the window's geometry is grid-snapped through WM base size and size increments, and insetting the content to paint our own shadow would have perturbed all of it. On Windows and macOS the window simply stops being frameless, since that hint is precisely what costs it the system shadow.

Details that are easy to get wrong were checked against KWin's and Breeze's sources rather than inferred, and two of them would otherwise have been silently wrong: X11 tiles must be uploaded non-premultiplied, because KWin reads a pixmap back as `QImage::Format_ARGB32`; and the shadow protocol is version 2 with a `destroy` request, not version 1 with `release` like the blur protocol beside it.

## Changes

- The frameless window publishes a drop shadow to KWin — `_KDE_NET_WM_SHADOW` on X11, `org_kde_kwin_shadow` on Wayland. `window_shadow` selects `None`, `Small`, `Medium`, `Large` or `VeryLarge`, using Breeze's own names and parameters. Documented as KWin-only: no other compositor or window manager offers a way for an undecorated window to publish one.
- Windows keeps its frame and zeroes it in `WM_NCCALCSIZE`, so DWM still draws the shadow and the rounded corners while the tab bar still covers the whole window. Input is deliberately untouched — answering `WM_NCHITTEST` to also recover Aero snap would mean teaching the hit test which QML items are interactive, which is a rewrite of the window's input path rather than a shadow fix.
- macOS keeps a decorated `NSWindow` whose content fills the title bar, so AppKit continues to draw the shadow, the rounded corners, full-screen, window tabbing and double-click-to-zoom. **The system's own traffic lights are used now instead of Contour's drawn copies**, which is the one user-visible change here; the tab strip reserves their corner, measured from the real buttons rather than guessed from chrome tokens.
- One table decides all of this per OS, so the Windows and macOS behaviour is pinned by tests that run on Linux CI — including one asserting the Linux path is unchanged.
- Menus and popups gain a drawn shadow. Every menu is an in-scene popup (a native platform menu cannot host the sub-menus built at runtime), so there is no OS surface for a compositor to shadow. It is drawn with plain rectangles rather than a `MultiEffect` shader: the terminal paints over the QML scene graph, and a popup rendering through an offscreen target does not reliably survive that.
- Three bugs found on the way: background blur could never be turned off again on X11 once enabled (the disable path deleted an atom nothing sets); `DWMWA_USE_IMMERSIVE_DARK_MODE` was forced on regardless of the `theme` setting, putting a dark border on a light-themed window; and every X11 property access leaked one xcb reply.

The Linux half is verified against a live Plasma 6 Wayland session. **The Windows and macOS adapters are not** — there is no such hardware here, and the offscreen suite cannot drive an `HWND` or an `NSWindow`. Everything decidable in them is in the pure policy table and covered by tests; the adapters themselves want eyes on real hardware before merge. The open question on Windows is whether `DwmExtendFrameIntoClientArea` produces a shadow at all on Qt's translucent D3D11 window, and whether it fights the acrylic backdrop set when `background_blur` is on. On macOS it is worth confirming that exactly one set of traffic lights appears and that the minimum-size behaviour survives going from borderless to decorated.


christianparpart:
Rebased onto current master (`2de56977`).

Master has moved a long way today — the Qt 6.11.2 pin, the docs build, FreeBSD, and a Ninja link-job pool all landed — so this branch needed replaying on top of it. All 19 commits reapplied; **one conflict**, in `.github/workflows/build.yml`.

That conflict was this branch's own `ci: fix the Windows and macOS builds, and pin Linux Qt to 6.11.1`. It had diagnosed the same broken Qt 6.11.2 release a day early and pinned Linux to 6.11.1. Master now pins **every** platform to 6.11.1 and retries the install past a bad mirror, so master's version is a strict superset and that hunk is redundant.

Resolved by taking master's side, which leaves `build.yml` **byte-identical to master** — this PR now changes no workflow at all. The commit keeps its three real source fixes (`CocoaWindowFrame.mm`, `Dwmapi.hpp`, `Win32WindowFrame.cpp`) and is retitled `contour: fix the Windows and macOS builds`, with the now-false Qt paragraph replaced by a note that master owns the pin. Everything else replayed untouched: 63 files, +4374/−198.

Verified locally before pushing: builds clean, and `contour_gui_test` passes — **752 passed, 1 skipped, 10140 assertions** (the skip is the display test that wants a real display). The 14 `[shadow]` cases pass on their own too, 441 assertions.

Not merging this, per instruction — just getting it green.

PR #2060, opened 2026-08-16
ci: pin Qt to 6.11.1 (corrupt AARCH64 archive), and repair the static build image

christianparpart:
Two independent CI failures. The first is why **every arm64 job is red right now**, on every branch.

## 1. Qt 6.11.2 ships a corrupt Linux AARCH64 archive

`6.11.*` started resolving to **6.11.2** today, and its Linux AARCH64 `qttools` archive is corrupt where it is published:

```
INFO    : Downloading qttools...
INFO    : Finished installation of qtdeclarative-Linux-Ubuntu_24_04-GCC-Linux-Ubuntu_24_04-AARCH64.7z
WARNING : Caught BadZipFile, terminating installer workers
ERROR   : Bad offset for central directory
zipfile.BadZipFile: Bad offset for central directory
```

The evidence that it is the archive and not a flaky mirror:

| observation | |
|---|---|
| yesterday's arm64 runs, resolving to **6.11.1** | green |
| today's **x86_64** runs, resolving to 6.11.2 | green |
| today's **arm64** runs, resolving to 6.11.2 | fail, every one |
| two attempts in one job, served by two *different* mirrors (`mirror.accum.se`, `qt.mirror.constant.com`) | both got the same truncated file |

So: pin to **6.11.1**, on every platform rather than arm64 alone so one Qt is what all of them are measured against. Raise it once a later 6.11 publishes an AARCH64 archive that unpacks.

Each of the four `install-qt-action` call sites also **attempts the install twice** now. That is deliberately *not* the fix for the above and does not help there — it is worth having for the case it does cover, a single mirror serving a truncated file, which is a verdict on the mirror rather than on the commit under test. Only a repeated failure fails the job. Inlined at each site rather than factored into a composite action: `continue-on-error` inside a composite action has [documented edge cases](https://github.com/actions/runner/issues/2418) when the step also takes inputs — exactly this shape.

## 2. Static build image: no OpenSSL, and GCC 13

The **Static build** job has been red for weeks, on two counts that surface one after the other.

**Configure fails first** — `find_package(OpenSSL REQUIRED)` in `src/net/CMakeLists.txt` backs the TLS decorator behind net's `ITlsContext` seam, and the image had no OpenSSL development package:

```
Could NOT find OpenSSL ... (missing: OPENSSL_CRYPTO_LIBRARY OPENSSL_INCLUDE_DIR)
  src/net/CMakeLists.txt:69 (find_package)
```

The image installs its packages by hand instead of through `scripts/install-deps.sh`, whose Ubuntu list has carried `libssl-dev` all along — the two lists have to be kept in step, and were not.

**Behind it waits the older failure**: Ubuntu 24.04's default `g++` is 13.3, whose libstdc++ has no `std::ranges::to` (C++23, arrived in GCC 14):

```
/contour/src/vtbackend/Color.cpp:159:44: error: 'to' is not a member of 'std::ranges'
```

Every other Linux job already builds with GCC 14 or Clang; this image was the last one on 13.3. Installing `g++-14` does not make it the default, so the compiler is named at configure time.

## Verification

- `actionlint` passes on the workflow set (`scripts/check-workflows.py`).
- The retry shape is the ordinary `id` + `continue-on-error` + `if: outcome == 'failure'` pair, and it was **observed working** in this PR's own first run — it ran the second attempt, which is how the corrupt-archive diagnosis was confirmed rather than assumed.
- **The static image fixes cannot be exercised before merging**: that job carries `if: github.ref == 'refs/heads/master'`, so pull requests skip it, and no Docker daemon was available locally to dry-run the image. Both changes are derived from the exact errors in the master run logs. If a third failure waits behind these two, it will only show up once this is on master.

## Merge order

Everything with an arm64 job is blocked on the pin, so **this wants to go in first**; #2055 and #2059 then need a rebase or merge from master to pick it up.

## Risk

Low. The pin returns arm64 to the Qt every job used yesterday. The retry can only turn a failing install into a second attempt — a job that installed Qt successfully takes the identical path it does today. The image changes add two packages and pin the compiler that every other Linux job already uses.

### DIFF
--- a/.github/static/DockerUbuntu
+++ b/.github/static/DockerUbuntu
@@ -4,6 +4,18 @@ FROM yaraslaut/static_qt:$VERSION
 #FROM qt_static:$VERSION
 ARG ARCH="x86_64"
 
+# Two additions over the plain build-dependency list, each answering a failure this job hit:
+#
+# libssl-dev -- src/net does find_package(OpenSSL REQUIRED) for the TLS decorator behind its
+# ITlsContext seam, so its absence fails the configure step outright rather than at link time. This
+# image installs its packages by hand instead of through scripts/install-deps.sh, whose Ubuntu list
+# does carry libssl-dev, so the two lists have to be kept in step and were not.
+#
+# g++-14 -- Ubuntu 24.04's default g++ is 13.3, whose libstdc++ has no std::ranges::to; that is C++23
+# and arrived in GCC 14. Every other Linux job already builds with GCC 14 or Clang, so this image was
+# the one place still on 13.3, and vtbackend/Color.cpp failed there with "'to' is not a member of
+# 'std::ranges'". The compiler is then named explicitly at configure time below, because installing
+# it does not make it the default.
 RUN apt-get update && apt-get install -y \
     lsb-release software-properties-common gnupg \
     wget gcc g++ \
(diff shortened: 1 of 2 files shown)


Yaraslaut:
Thanks for fixing it :)

PR #2067, opened 2026-08-16
build: link the threading library where the headers demand it

christianparpart:
With #2063 in, the FreeBSD build compiled for the first time since 2026-07-29 — and reached the link step, where it stopped:

```
[ 34%] Linking CXX executable crispy_test
ld: error: undefined symbol: pthread_create
```

## Why

`crispy-core`'s **public** headers — `BufferObject.hpp`, `LogSink.hpp`, `ReadSelector.hpp`, `FileDescriptor.hpp`, `Environment.hpp` — expose `std::mutex`, `std::condition_variable` and `std::thread`. Threading is part of that library's INTERFACE, so every consumer instantiates it and needs the threading library on its link line.

Neither `crispy-core` nor `net` declared `Threads::Threads` at all. Of the ten first-party modules, only `vtbackend` did:

| module | uses threading | linked `Threads::Threads` |
|---|---|---|
| crispy | 6 files | ✗ |
| vtpty | 6 files | ✗ |
| net | 3 files | ✗ |
| vthost | 3 files | ✗ |
| vtrasterizer | 2 files | ✗ |
| vtbackend | 4 files | ✓ |

**No Linux job could notice**: glibc has carried `pthread_create` in libc since 2.34, so the omission is free there. FreeBSD keeps those symbols in `libthr`, where a missing `-pthread` is a link error — and this is the first time in seven weeks a FreeBSD build got far enough to link anything.

## The fix

Declared PUBLIC on `crispy-core`, which carries it to `vtpty`, `vthost` and `vtrasterizer` — all of them reach threading through that same header set, so one declaration covers them correctly rather than four copies.

`net` gets its own, because it deliberately does not depend on crispy (per its README, so the module stays consumable without dragging `crispy` along) and therefore cannot inherit it.

## Verification

- `crispy_test`: **1613 assertions in 109 test cases** — pass
- `net_test`: **633 assertions in 121 test cases** — pass
- configures and links clean

## Note

This is the fourth layer of the FreeBSD build: `<stop_token>` → `environ` → this. Each was a real, pre-existing defect that no other platform was positioned to catch. The job stops at the first error, so I cannot promise it is the last — but the failures have moved from configure, through compile, to link, which is the right direction.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -448,7 +448,13 @@ jobs:
   # }}}
   # {{{ FreeBSD
   freebsd:
-    if: github.ref == 'refs/heads/master'
+    # TEMPORARY, REVERTED BEFORE MERGE: also run on this branch. The FreeBSD build has not compiled
+    # since src/coro landed and is being repaired one layer at a time (<stop_token>, then environ,
+    # now the threading library); with the job gated on master, each layer costs a merge and a full
+    # master run to discover the next. Restore this line to
+    #     if: github.ref == 'refs/heads/master'
+    # once the job is green here.
+    if: github.ref == 'refs/heads/master' || github.head_ref == 'fix/freebsd-link-threads'
     runs-on: ubuntu-latest
     # Unversioned deliberately: the VM's release is whatever vmactions/freebsd-vm defaults to, which
     # has moved on to 15.1-RELEASE. This job was named "FreeBSD 13" long after that stopped being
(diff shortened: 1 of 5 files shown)


christianparpart:
**Temporarily running the FreeBSD job on this branch** (commit `813ae5d7`), to be reverted before merge.

That job is gated on `master`, so every repair layer has cost a merge plus a full master run just to learn what the next layer is — and the VM takes ~20 minutes to reach the failure. Three layers have gone that way already (`<stop_token>` → `environ` → threading). Running it here turns the loop into a push.

The gate goes back to:

```yaml
    if: github.ref == 'refs/heads/master'
```

once the job is green here, and that revert is the last commit before merge.

PR #2093, opened 2026-08-22
daemon: resolve a profile's shell, sandbox policy, and presentation — not just its emulation settings

### LINKED ISSUE #2092: contour client / daemon silently ignores most of a profile's non-emulation settings (shell, ssh, colors, presentation, escape_sandbox)
## Summary

`contour daemon` (and therefore every `contour client` session, since it hosts through the
daemon) only resolves a profile's *emulation* settings via `config::resolveEmulationSettings()`
(`ContourApp::daemonAction()`, `src/contour/cli/ContourApp.cpp`). That function returns a bare
`vtbackend::Settings`, which structurally cannot carry most of what a `TerminalProfile` actually
configures. Everything beyond that narrow slice silently reverts to a hardcoded default —
**on the very first daemon spawn**, not just when an already-running daemon is reused.

A local `contour terminal` launch does not have this problem: `TerminalSession::createSettingsFromConfig()`
and `SessionFactory::createPty()` layer the rest of the profile on top of the same
`emulationSettings()` call. The daemon path never calls either.

## Reproduction

```
contour client --profile myprofile
```

where `myprofile` sets any of: a custom `shell:`, a non-default `status_line`, a custom cursor
style, a non-default `color_scheme`, `ssh:`, or `escape_sandbox: false`.

Expected: behaves like `contour terminal --profile myprofile`.
Actual: silently falls back to defaults for all of the below.

## Confirmed gaps

1. **Shell** (`ContourApp::daemonAction`, `src/contour/cli/ContourApp.cpp:707`) — hardcoded to
   `vtpty::Process::loginShell()`. The profile's `shell:` block (program/arguments/workingDirectory/env)
   is never read at all.

2. **Presentation settings** (`TerminalSession::createSettingsFromConfig`,
   `src/contour/session/TerminalSession.cpp:209-289`) — ~15 fields the daemon never sets: cursor
   shape/blink/display, status line type/position/indicator text, smooth/momentum scrolling,
   screen transition style/duration, mouse wheel scroll multiplier, `copy_last_mark_range_offset`,
   highlight timeout/word, `sync_window_title_with_host_writable_status_display`, tab naming mode.

3. **Color palette** (`TerminalSession.cpp:284`) — the profile's `colors:` scheme never reaches
   `Terminal::_colorPalette`, which is semantically load-bearing (default fg/bg, indicator status
   line colors, fold-marker colors, indexed-color lookups) — not just a rendering skin. The
   client-side remote-mirroring path (`src/contour/remote/`) has no palette handling of its own to
   compensate.

4. **SSH profiles** (`SessionFactory::createPty`, `src/contour/session/SessionFactory.cpp:45` vs.
   `vthost::makeShellPtyFactory`, `src/vthost/Daemon.cpp:81-94`) — a profile with `ssh.hostname` set
   opens a plain local shell under the daemon instead of a `vtpty::SshSession`. `DaemonConfig` has
   no field to carry SSH config at all.

5. **`escape_sandbox`** (`vthost::makeShellPtyFactory`, `src/vthost/Daemon.cpp:92`) — hardcoded to
   `true`, ignoring a profile's `escape_sandbox: false`.

6. **`CONTOUR_PROFILE` env var** (`vthost::hostedShellEnvironment`, `src/vthost/HostedShell.hpp:40`)
   — never set for daemon-hosted shells, unlike a local session
   (`ContourGuiApp.cpp:938`). Lower severity/confidence — the same header already documents a
   sibling gap (`TERM`/`COLORTERM`/`TERM_PROGRAM`) as a deliberately deferred decision, so the
   maintainers may want to fold this in with that rather than fix in isolation.

## Root cause

All six share the same structural cause: `src/vthost` is (deliberately) Qt-free and has no access
to the full `contour::config::TerminalProfile` model — only to `vtbackend::Settings` via
`config::resolveEmulationSettings()`. There is currently no Qt-free path that resolves *all* of a
profile's non-GUI-rendering fields (shell, ssh, escape_sandbox, and the presentation/color fields
above) for a caller like `daemonAction()` to hand to `DaemonConfig`.

## Suggested fix direction

Factor the logic currently duplicated between `TerminalSession::createSettingsFromConfig()` +
`SessionFactory::createPty()` (local sessions) into a shared, Qt-free "resolved session config"
step that both the local-session path and `ContourApp::daemonAction()` can call, then extend
`DaemonConfig` (`src/vthost/Daemon.hpp`) to carry what it's currently missing (full `shell`
ExecInfo already exists as a field but is never populated correctly; needs `ssh`,
`escapeSandbox`, and the daemon needs to pass the complete `vtbackend::Settings` including
presentation/color fields).

cc @contour-terminal/contour maintainers

Yaraslaut:
## Summary

Fixes #2092: `contour daemon` (and therefore every `contour client` session, since it hosts
through the daemon) resolved only a profile's VT-emulation settings and hardcoded the shell and
sandbox policy, silently ignoring the profile's `shell:`, `escape_sandbox:`, and presentation/color
configuration — on the very first spawn, not only when an already-running daemon is reused.

## Root cause

`ContourApp::daemonAction()` called `config::resolveEmulationSettings()`, whose return type
(`vtbackend::Settings`, restricted to history/terminal-id/images/reflow/etc.) structurally cannot
carry a profile's shell or presentation fields. Everything beyond that narrow slice — the shell
program/args/env, `escape_sandbox`, cursor style, status line, scrolling, color scheme — was only
ever applied by `TerminalSession::createSettingsFromConfig()`, reachable by a local GUI session
alone.

## Fix

- **`contour::config`**: add `ResolvedSessionConfig` and `resolvedSessionConfig()`/
  `resolveSessionConfig()` — the `resolveEmulationSettings()`-shaped counterparts that also resolve
  the shell (with `CONTOUR_PROFILE` set, matching a local session), the sandbox policy, and the
  presentation/color fields `createSettingsFromConfig()` used to apply GUI-side only. The
  load-resolve-refuse sequence is shared between `resolveEmulationSettings` and
  `resolveSessionConfig` via a new `loadAndFindProfile()` helper rather than duplicated.
- **`vthost`/daemon**: `daemonAction()` now calls `resolveSessionConfig()` and fills
  `DaemonConfig::settings`/`shell`/`escapeSandbox` from it, replacing the hardcoded
  `vtpty::Process::loginShell(false)` / `escapeSandbox=true`. `DaemonConfig` gains `escapeSandbox`
  (default `true`, preserving prior behavior for a bare `DaemonConfig{}`), and
  `makeShellPtyFactory()` takes it as a parameter.
- **GUI session**: `TerminalSession` now calls the shared `resolvedSessionConfig()` instead of its
  own copy, and `updateColorPreference()` calls the now-shared `preferredColorPalette()` — so the
  two paths cannot drift apart on what a profile means again.

### Deliberately out of scope

- **SSH profiles** (`profile.ssh`) are not addressed here — the daemon has no hostkey-verification
  policy to open an `SshSession` under (unlike the GUI's own TODO stub for this). This needs its
  own design decision and is left as an open item on #2092.
- A daemon's shell working directory, when the profile leaves it unset, now falls through to
  whatever directory the daemon process itself started in (matching local-session behavior) rather
  than the previously hardcoded `homeDirectory()` — a local session never had that fallback either,
  so this brings the daemon in line rather than introducing a new divergence.

## Testing

- New Catch2 cases in `Config_test.cpp` (`resolveSessionConfig` carries shell/sandbox/presentation/
  color fields; defaults match a profile that sets nothing; refuses an unknown profile) and
  `Daemon_test.cpp` (`DaemonConfig` defaults to escaping the sandbox).
- Full local verification on this Windows checkout (`clangcl-debug` preset):
  - `vthost_test`: **4083 assertions / 326 test cases, all passed**.
  - `contour_gui_test` `[config]` tag: **829 assertions / 138 test cases, all passed**.
  - `contour_gui_test` `TerminalSession*`: **798 assertions / 105 test cases, all passed**
    (confirms the local-session refactor is behavior-preserving).
  - Full `contour_gui_test` suite: **11066/11067 assertions passed** — the one failure
    (`SessionInput_test.cpp:266`, a Kitty-keyboard-protocol modifier-encoding mismatch) reproduces
    identically on a clean `master` checkout and is unrelated to this change. 3 display tests skip,
    needing a real display (`CONTOUR_TEST_DISPLAY=1`), which this environment doesn't have.
- Ran the `/simplify` review (reuse/simplification/efficiency/altitude passes) on the diff and
  applied its two in-scope, low-risk findings: deduped the load/lookup/error sequence between the
  two resolvers, and cleaned up the pre-existing `tabNamingMode` string-parsing helper (now uses
  `crispy::toLower` and a small array scan instead of a hand-rolled `tolower` loop and three
  sequential `if`/`else if` checks). A deeper finding (folding the status line's `Indexing=`
  attribute into `StatusLineDefinitions::Tabs`'s real `StringInterpolation`-based parser instead of
  a separate ad-hoc string scan) is a legitimate but separable improvement, not attempted here.

Closes #2092 (partially — SSH support tracked as a remaining open item on that issue).

## Review follow-ups (since the original review)

- **CI fix — clang-tidy.** `HarnessOptions` gained a `vtworkspace::Layout` member on this branch, so
  `NativeHarness`'s by-value constructor parameter stopped being a trivially-copyable bundle and
  `modernize-pass-by-value` / `performance-unnecessary-value-param` began firing on
  `NativeSession_test.cpp`. Moved into the member.
- **A startup-layout pane's `arguments` without a `command` are no longer dropped silently.** Such a
  pane engaged an override whose `program` was empty, and the arguments were then never copied into
  it. Arguments name what a command is run with, so the pane is malformed; it now engages no
  override at all (the daemon's configured shell runs, as before) and logs what it ignored, the way
  the neighbouring `profile` override already reports itself. Covered by a new test.
- **Comment corrections** above `realizeStartupLayout`'s seeder: it described the override condition
  without its `arguments` clause, and claimed to *mirror* the local path's overlay rule while
  deliberately differing on directories — the daemon has no separate `cwd` channel, so an override
  is the only way a pane's directory can reach the session at all.
- **Rebased onto master** (`86ec2ec8`), so CI runs against current head.

Local verification after these changes: `vthost_test` **4133 assertions / 335 test cases, all
passed**.


christianparpart (dismissed):
LGTM, and you can merge it whenever you have also manually tested it. :-)

christianparpart (dismissed):

christianparpart:
@Yaraslaut can you fix the failing test as part of it, please?

PR #2097, opened 2026-08-24
fix(session): honor capture_buffer permission via requestPermission gate

### LINKED ISSUE #2089: capture_buffer permission is never consulted: allow/deny are ignored and "remember" does nothing
## What is wrong

`TerminalSession::requestCaptureBuffer()` emits the permission dialog signal directly instead of going through the shared `requestPermission()` gate. The config-driven call has been commented out since at least 2023 (`a65f9f57` only renamed members *inside* the comment):

```cpp
// src/contour/session/TerminalSession.cpp:817
void TerminalSession::requestCaptureBuffer(LineCount lines, bool logical)
{
    if (!_display)
        return;

    _pendingBufferCapture = CaptureBufferRequest { .lines = lines, .logical = logical };

    emit requestPermissionForBufferCapture();
    // _display->post(
    //     [this]() { requestPermission(_profile.permissions.captureBuffer, GuardedRole::CaptureBuffer); });
}
```

`requestPermission()` (`TerminalSession.cpp:763`) is what reads the configured `Permission` and the remembered per-session answer. Skipping it means none of that is consulted.

## What breaks

Three user-visible things, most important first:

1. **`capture_buffer: deny` does not deny.** A profile that explicitly refuses buffer capture still pops the dialog, and the user can approve it. The setting reads as a security control and is not one.
2. **`capture_buffer: allow` does not allow.** A profile that pre-approves capture is still interrupted by a dialog on every request.
3. **"Remember my answer" does nothing for this role.** `executePendingBufferCapture()` *writes* `_rememberedPermissions[GuardedRole::CaptureBuffer]` (`TerminalSession.cpp:832`), but only `requestPermission()` ever *reads* that map. The answer is stored and never consulted, so the user is asked again every time.

The option is documented and parsed as if it worked: `capture_buffer` in `ConfigDocumentation.hpp:2174`, loaded at `Config.cpp:2137`, held at `Config.hpp:434`.

## Why it has gone unnoticed

`TEST_CASE("TerminalSession: config-driven permissions resolve without asking the user")` (`src/contour/session/TerminalSession_test.cpp:2028`) *does* set `captureBuffer = Permission::Deny` — but then calls `session->executePendingBufferCapture(false, false)` directly, which is the executor **after** the gate. It never calls `requestCaptureBuffer()`, so the one function containing the bug is not on the path under test, and the assertion is `CHECK_NOTHROW` rather than a check that the configuration was honoured.

The shared gate is covered. The call site that bypasses it is not.

## How to fix

Mirror what the `screenshot` role does (`TerminalSession.cpp:860`, added in #2088):

```cpp
void TerminalSession::requestCaptureBuffer(LineCount lines, bool logical)
{
    if (!_display)
        return;

    _pendingBufferCapture = CaptureBufferRequest { .lines = lines, .logical = logical };

    _display->post([this]() {
        requestPermission(_profile.permissions.value().captureBuffer, GuardedRole::CaptureBuffer);
    });
}
```

Two traps, both of which the commented-out line walks into:

- **`.value()` is required.** `permissions` is a `ConfigEntry<PermissionsConfig, …>` (`Config.hpp:711`), not a bare struct. The commented line predates that wrapper, so uncommenting it verbatim does not compile.
- **The `post()` is load-bearing, not decoration.** `requestCaptureBuffer` is a `Terminal::Events` hook reached from the terminal/parser thread. `requestPermission()`'s `Allow` branch runs `executeRole` synchronously, which reaches `executePendingBufferCapture()` → `primaryScreen().captureBuffer()` and `flushInput()`. Those belong on the GUI thread, like every other guarded role's executor. Dropping the hop would run them on the terminal thread.

Note that the `emit` disappears — `requestPermission()` emits `requestPermissionForBufferCapture()` itself on the `Ask` branch (`TerminalSession.cpp:791`). Leaving both in place would raise two dialogs.

## How to verify

Extend the existing test at `TerminalSession_test.cpp:2028` so it drives `requestCaptureBuffer()` rather than the executor:

- with `capture_buffer: deny` — nothing reaches `mockPtyOf(session).stdinBuffer()`, and no dialog signal is emitted;
- with `capture_buffer: allow` — the capture reply appears in the PTY buffer without a dialog;
- with `capture_buffer: ask` and a remembered answer seeded through `executePendingBufferCapture(allow, /*remember=*/true)` — a second request resolves from memory instead of asking again.

`FakeDisplaySurface` runs posted work immediately by default (`runPostsImmediately`), so the assertions are about the effect rather than the thread hop.

---

Found while implementing OSC 533 (#2088), whose `screenshot` permission does consult its configuration and can serve as the reference for the shape above. Not fixed there to keep that PR to its own scope.

VedantMadane:
## Summary

fix(session): honor capture_buffer permission via requestPermission gate

## Changes

- TerminalSession.cpp: requestCaptureBuffer uses requestPermission like other roles
- TerminalSession_test.cpp: drive requestCaptureBuffer for deny/allow configs

Fixes #2089

### DIFF
--- a/src/contour/session/TerminalSession.cpp
+++ b/src/contour/session/TerminalSession.cpp
@@ -874,9 +874,11 @@ void TerminalSession::requestCaptureBuffer(LineCount lines, bool logical)
 
     _pendingBufferCapture = CaptureBufferRequest { .lines = lines, .logical = logical };
 
-    emit requestPermissionForBufferCapture();
-    // _display->post(
-    //     [this]() { requestPermission(_profile.permissions.captureBuffer, GuardedRole::CaptureBuffer); });
+    // Consult config + remembered answers via the shared gate (same shape as other roles).
+    // post() keeps executeRole / captureBuffer on the GUI thread (#2089).
+    _display->post([this]() {
+        requestPermission(_profile.permissions.value().captureBuffer, GuardedRole::CaptureBuffer);
+    });
 }
 
 void TerminalSession::executePendingBufferCapture(bool allow, bool remember)
--- a/src/contour/session/TerminalSession_test.cpp
+++ b/src/contour/session/TerminalSession_test.cpp
@@ -2193,8 +2193,25 @@ TEST_CASE("TerminalSession: config-driven permissions resolve without asking the
 
     // A font-change action with a pre-allowed permission applies without asking (no crash, no dialog).
     CHECK_NOTHROW((*session)(actions::ResetFontSize {}));
-    // Requesting a capture with a config-denied permission resolves to the deny path.
-    CHECK_NOTHROW(session->executePendingBufferCapture(false, false));
+
+    // Drive the real gate for CaptureBuffer (not only the post-gate executor).
+    // Deny must not write a capture reply into the PTY stdin buffer (#2089).
+    auto const stdinBeforeDeny = mockPtyOf(*session).stdinBuffer().size();
+    CHECK_NOTHROW(session->requestCaptureBuffer(vtbackend::LineCount(10), false));
+    CHECK(mockPtyOf(*session).stdinBuffer().size() == stdinBeforeDeny);
+}
+
+TEST_CASE("TerminalSession: capture_buffer allow resolves without dialog path side effects",
+          "[contour][session][permission]")
+{
+    contour::test::TestApp testApp;
+    auto const profileName = registerProfile(testApp.app(), "perm-allow", [](contour::config::TerminalProfile& p) {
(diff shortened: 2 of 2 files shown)


christianparpart:
nice catch, many thanks, @VedantMadane

VedantMadane:
Closing as superseded by #2091 (merged), which already routes CaptureBuffer through the shared permission gate (postPermissionRequest / 
equestPermission), answers every outstanding capture, and covers deny/allow/ask + remember with tests. Issue #2089 is closed.

Rebase onto current master leaves an empty diff against upstream — no unique changes remain. Thanks for the parallel fix!

Issue #732, opened 2022-06-11
Implement "Set warning-bell volume (DECSWBV), VT520."


christianparpart:
Implement this VT sequence.

```
CSI Ps SP t
       Set warning-bell volume (DECSWBV), VT520.
         Ps = 0  or 1  -> off.
         Ps = 2 , 3  or 4  -> low.
         Ps = 5 , 6 , 7 , or 8  -> high.
```

buggylintermachine:
Hey since this is similar to #733 I'd like to pick this up too

buggylintermachine:
Bumping this message for visibility. Not sure what the ETA usually is for responses but just wanted to make sure that the updates are being seen.

PR #2100, opened 2026-08-30
vtbackend: implement DECSWBV, the VT520 warning-bell-volume sequence

buggylintermachine:
Adds BellVolume {Off, Low, High} (vtbackend/core/Primitives.hpp) and Terminal::setWarningBellVolume, and wires CSI Ps SP t up through Functions.hpp and impl::DECSWBV in Screen.cpp: Ps 0/1 selects Off, 2-4 Low, 5-8 High, anything else is ApplyResult::Invalid, and the parameterless form (minimumParameters = 0) is accepted rather than rejected.

RIS restores the factory volume in Terminal::hardReset(), and the GUI frontend scales the configured bell gain by the selected volume in TerminalSession::bell() (Off silences it, Low halves it, High leaves it unchanged).

Covered by three Screen_test.cpp cases pinning the Ps mapping, the parameterless default, and the RIS reset.

Implements #732.


buggylintermachine:
amended the commit with the changelog changes

buggylintermachine:
@Yaraslaut Please let me know if there are any issues. If not I'd would like to see if the workflow catches any issues.

buggylintermachine:
@christianparpart @Yaraslaut If you guys are busy are there any other contributors or maintainers I can ping to look into this PR? No rush at all for the PR. I just wanna make sure the PR isn't forgotten.

src/vtbackend/screen/Screen.cpp
```diff
@@ -3831,6 +3831,30 @@ namespace impl
...
+                    case 6:
+                    case 7:
+                    case 8: terminal.setWarningBellVolume(BellVolume::High); break;
+                    default: return ApplyResult::Invalid;
+                }
+                return ApplyResult::Ok;
+            }
+            else
+                return ApplyResult::Invalid;
+        }
```
Yaraslaut, line 3856: This case is not covered in unit tests, can we please add it

Yaraslaut (commented):
Thanks, looks good, one coverage is missing, i guess we need to make sure that ill-formed sequence will be properly handled

src/vtbackend/screen/Screen.cpp
buggylintermachine: Thanks for the catch. I will add this in later today.

buggylintermachine: Turns out that DECSWBV is registered with maximumParameters = 1 which means that Functions::select() does not take anymore than 1 param. In order to test the case I make a multi param call that bypasses the regular flow so it hits the switch with 2 params. 

DECSCUSR also has a similar gap. I can resolve that in a follow up PR. 

I also added another  DECSWBV case to test the behavior when default is hit  (Ps = 9).

Yaraslaut: thanks, for the investigation, would be great to also ensure behavior in the DECSCUSR

Yaraslaut (approved):
Thanks :)

buggylintermachine:
@christianparpart Hey! Just wanted to ping again for visibility so that this PR is on your radar.

christianparpart:
I'll review it today, sorry for the delay - apply anything on top (if needed at all), and then merge. Thank you very much already, @buggylintermachine :-)

buggylintermachine:
I'll get started on the margin bell volume changes as well.

christianparpart (approved):

PR #2107, opened 2026-09-08
Four render-latency fixes: fontconfig sort, synchronized-output freeze, column reflow, APC parsing

Yaraslaut:
Four performance fixes found with the Tracy instrumentation from #2106, each with a before/after on
the same workload. They are independent and reviewable one at a time.

## 1. Fontconfig sorted every installed font, per codepoint, inside the frame

Text in a script the system has no font for rendered at roughly **3 fps**.

`resolveByCoverage()` asks the locator which font covers a run the configured fallback chain could
not render, and `FontconfigLocator::resolve()` answered with `FcFontSort`, which ranks *every*
installed font — 379 on this machine. The coverage cache is keyed on a single codepoint, so an
unsupported block paid that sort **once per codepoint**: 20,992 times for CJK alone, each on the
render thread inside the frame, every one returning nothing.

The fix probes the union of every installed font's charset, built once. Exact rather than heuristic —
a codepoint absent from that union is in no font's charset, so the sort it replaces could not have
found a covering font either — which is why it can sit *in front of* the query rather than beside it.

Identical call counts before and after (22,939 shapes, 21,095 resolutions — the same work, no longer
the same cost):

| | before | after |
| --- | --- | --- |
| frames over budget | 90.9% | 9.3% |
| median frame | 288.9 ms | 12.3 ms |
| `resolveByCoverage` | 7192 ms @ 341 µs | 9.5 ms @ 0.4 µs |

## 2. Synchronized output could freeze the display indefinitely

While a DECSET 2026 block is open nothing asks for a repaint, so an application that opened one and
then stalled, blocked or **died** before closing it froze the display with no recovery path — not a
slow frame, no frames. One capture showed 1008 opens against 825 closes.

Rendering resumes once a block outlives `Settings::synchronizedOutputTimeout` (150 ms). The mode
register is untouched, so DECRQM still reports what the application set.

It is a throughput fix as well as a robustness one, which a mean block length hides — what a cap acts
on is the tail:

| | before | after |
| --- | --- | --- |
| frame rate | 13.6 fps | **29.7 fps** |
| inter-frame gap p99 | 498.8 ms | 170.5 ms |
| gaps over 250 ms | 147 | **0** |

Three details that were each a way to get this wrong, and are worth a reviewer's attention: the
deadline uses `steady_clock`, **not** `_currentTime`, which only advances when a frame is painted and
would be frozen in exactly the state the deadline exists to break; `isRenderingSuppressed()`
*replaced* all three bare reads of the flag rather than sitting beside them (the first version gated
only the `screenUpdated` path, and the new test caught that the screen still did not repaint); and it
keys on the atomic flag rather than the mode register, which the render thread must not read.

## 3. Column changes scaled with `history_limit`, not with content

A column change rebuilt the scrollback ring, and the ring is sized to `HistoryLimits::capacity`
rather than to what the grid holds — so the work was proportional to the **configured** scrollback
whatever the grid contained. Contour negotiates eight or nine column changes at startup, and a window
drag pays one per column step.

```
                                     before      after
startup column work                  34.58 ms    0.02 ms
grow+shrink on 8,000 lines,
  history_limit 1,000                 1,710 µs   1,302 µs
  history_limit 100,000              22,025 µs  13,474 µs
```

A missing reserve, then skipping the reflow rebuild when nothing in the used range is wrapped (and,
for shrink, when nothing would be cut), then walking the used range instead of the whole ring.

Two traps, both of which cost real time: `LineSoA::usedColumns` is a hint, not an authority —
`resizeLineSoA` *clamps* it to the new width, so a line whose content was already cut still reports a
width that fits; using it in the shrink guard truncated content and failed 13 `[grid]` cases.
`Grid::verifyState()` is inside `#ifdef CONTOUR_VERIFY_STATE`, **OFF** in `clang-release` and **ON**
in `clang-asan`, so grid tests in release check behaviour but not the ring's invariants — every
change here was verified under ASan too.

## 4. APC payload was parsed one byte at a time

`parseFragment()` selected the bulk scanner only for `DCS_PassThrough`, and `parseBulkText()` returns
`FallbackToFSM` for every non-Ground state — so each byte of an APC body walked the state machine for
a table lookup, an action dispatch and a `putAPC(char)`. Sixel had a bulk path; the kitty graphics
protocol, which sends its images as APC bodies, did not.

The new scanner stops below 0x20, at 0x7F and at 0x80 and above — a strict subset of the bytes the
table maps to `APC_Put` — so no byte changes meaning, only the route it takes. BEL, ESC and 8-bit ST
still terminate through the state machine, which is the only place that tells an 8-bit ST from a
UTF-8 continuation byte.

On an image-heavy workload carrying 1.4 GB of payload, `parseFragment` self time falls from
**7.55 s to 1.06 s** (46.6% → 6.6% of the run).

## Also here

- **The per-cell Tracy zones move behind `CONTOUR_TRACY_CELL_ZONES`, off by default.** Three of them
  fire once per cell per renderer per frame — 22.3 million in a 16-second capture. Cheap to record
  (~20 ns), ruinous to read back: a one-minute capture reached 345 MB and took the profiler over five
  minutes and 5.8 GB of RSS to open. Gating them gives 16 MB, 2.7 M events, 0.3 GB and an instant
  export. They earned their cost once, by showing per-cell work is not a bottleneck.
- **`readFromPty` is split into `pty.waitOne` and `pty.readSome`.** One zone over both meant its
  26.8 µs mean could be read as "blocked waiting" or "the syscall is expensive", which call for
  opposite fixes. Split: `waitOne` is 0.51 µs at the median, so the poll is nearly free and its
  7.55 s is genuine idle waiting; `readSome` is the real per-byte cost at 1.31 ns/byte. That ruled out
  both a non-blocking-read-first change and a larger read buffer — 35% fewer syscalls at 64 KB cost
  the same total time, because the cost is the tty layer copying bytes.
- **The Tracy stub lint gate is removed.** `CONTOUR_TRACY` defaults to OFF, so every CI and packaging
  job already fails to compile an undefined macro, naming it just as precisely; `TracyStub_test.cpp`
  covers the stub's semantics. It duplicated the build, and recognising Tracy's vocabulary by the
  *shape* of a name had already reported contour's own `FramePass` and `FrameDrawItem` types as
  undefined Tracy macros.
- **`docs/internals/profiling.md`** gains the measurement rules these fixes were found with — self
  time vs per-event and what each answers, the pipeline-latency join and its two traps, paired-run
  discipline, and why our frame count is not the frame rate under synchronized output.

## Testing

`ctest --test-dir out/clang-release` green, and the reflow changes verified under ASan/UBSan
specifically for `verifyState()` coverage. New test: `Terminal.SynchronizedOutputTimesOut`.

Note that `-E e2e` filters by test *name*, not label; `-LE e2e` is the one that excludes the label.


christianparpart:
@Yaraslaut this still needs an actual changelog entry. I think this is worth mentioning to the human reader's audience.

Issue #2108, opened 2026-09-09
Kitty graphics: implement the animation protocol (a=f / a=a / a=c) — clients retransmit whole frames without it, 15x the bytes


Yaraslaut:
## Summary

Contour parses the kitty graphics animation actions (`a=f` frame transmit, `a=a` animate, `a=c`
compose) and then declines them:

```cpp
// src/vtbackend/screen/Screen.cpp:5538
case Action::Frame:
case Action::Animate:
case Action::Compose:
    // Animation is not implemented. Say so rather than silently accepting frames that will
    // never be shown.
    replyKittyGraphics(command, "ENOTSUP:animation not supported");
    return;
```

Declining is the right behaviour for an unimplemented feature — it is honest, and it is why clients
fall back cleanly instead of showing garbage. But the fallback is expensive: a client that wanted to
send **frame deltas** instead re-transmits the **whole image on every frame**.

Measured against kitty 0.48.2 with the same client, the same demo and the same terminal geometry
(96 rows x 173 cols), reading notcurses' own byte accounting:

| | bitmaps sent | bytes received | synchronized updates |
| --- | --- | --- | --- |
| kitty 0.48.2 | 553 | **393.33 MiB** | 0 (0.00%) |
| contour (this branch) | 580 | **5.80 GiB** | 583 (98.81%) |

**15x the data for the same number of images.** The synchronized-update column is a consequence
rather than a second problem: notcurses wraps contour's full-frame retransmissions in DEC 2026
because they are large enough to tear, and sends kitty's small deltas with no synchronization at all.

## Why the difference is this large

notcurses' own escape templates show the two paths (`strings libnotcurses-core.so.3`):

```
_Gf=32,s=%d,v=%d,i=%d,p=1,a=t,%s                        full transmit, raw RGBA
_Ga=f,x=%d,y=%d,s=%d,v=%d,i=%d,X=1,c=%d,r=%d,%s;        frame delta at (x,y), size s x v
_Ga=a,i=%d,c=2,q=2                                      animate
_Ga=c,x=%d,y=%d,X=%d,Y=%d,w=%d,h=%d,i=%d,r=1,c=2,q=2;   compose
```

The `a=f` form carries an offset and a size, so it updates **only the rectangle that changed** in an
image the terminal already holds. For video-like content most of the frame is unchanged, which is
where the 15x comes from. Note also that the payload is uncompressed 32-bit RGBA in both cases —
notcurses never sends `o=z` — so this is not a compression question.

[...]

PR #2109, opened 2026-09-09
Implement the kitty graphics animation protocol (a=f, a=a)

Yaraslaut:
Implements the kitty graphics animation protocol, closing the gap measured in #2108.

## The gap

Contour parsed `a=f`, `a=a` and `a=c` and then declined all three. Declining is honest, but the
fallback is expensive: a client that wanted to send **frame deltas** re-transmits the **whole image
every frame**. Measured against kitty 0.48.2 with the same client, demo and geometry (96x173):

| | bitmaps sent | bytes received |
| --- | --- | --- |
| kitty 0.48.2 | 553 | **393.33 MiB** |
| contour (before) | 580 | **5.80 GiB** |

## What is implemented

- **`a=f`** builds or edits a frame, compositing the transmitted rectangle onto the frame being
  edited, or onto a copy of the one `c=` names when creating a new one. `c=` is the source only for a
  frame that does not yet exist — preferring it once `r=` names an existing frame silently discards
  earlier blocks of a frame composed from several rectangles, which is the natural shape when every
  block comes from one template. Source-over blending, or a plain overwrite for `X=1`.
- **`a=a`** applies play state, loop count, per-frame gap, and `c=` to make a frame current.
- **The control data**, whose keys are overloaded in a way that is silently wrong if missed: under
  `a=f`, `c`/`r`/`z` are a base frame, the frame being edited and a gap; under `a=a`, `s` and `v` are
  the play state and loop count, not the pixel dimensions, so "loop three times" would otherwise be
  read as a 3-pixel-tall image. The action can appear anywhere in the control data, so the
  reinterpretation runs after the whole block is decoded. `X=` and `Y=` were not decoded at all.

`a=c` (compose) and self-running playback (`a=a,s=2`/`s=3`) report `ENOTSUP`. Nothing advances frames
on a timer yet, and answering OK would leave a client believing an animation is running while the
screen holds one frame forever — worse than the ENOTSUP it used to get, which at least made it fall
back. A playback refusal does not discard a frame selection made by the same command.

### Frames are found in the grid, not tracked

Making a frame current replaces the image's fragments **wherever they are**, rather than redrawing at
a remembered position. This is the design point the review rounds converged on: a placement *is* the
fragments in the cells, and the cells move — they scroll in both directions, are inserted and
deleted, erased, and reflowed by a resize. Any recorded coordinate has to be kept in step by every
one of those paths, and a stale one stamps the frame over live text. The replacement is built from
the fragment already on screen, so a new frame inherits the placement's geometry, alignment and gap
colour rather than whatever the cursor carries at the time — which would otherwise write the current
hyperlink into every covered cell.

### Bounds

All reachable from the wire, so all checked: `MaxAnimationFrames` caps the frame list, since each
frame is a full copy of the image and frames are created by *naming* a 1-based number; frames are
charged against `MaxStoredImageBytes` by distinct image, because a gap in the numbering aliases frame
1 and charging every slot would refuse legitimate frames; a frame number beyond the next is refused
rather than fabricating the ones in between; the rectangle must fit the image and the payload must
match its declared dimensions; and `f=`, `t=` and `o=` are honoured rather than silently
reinterpreted.

### Protocol details worth a reviewer's eye

- **Animation control is legal before any `a=f`.** Every stored image already has a root frame — the
  protocol has no command to send frame 1 separately — and the spec relies on this to set the root
  frame's gap. Answering ENOENT tells a client its image is gone and provokes the full
  re-transmission this feature exists to avoid.
- **A missing frame answers `ENOENT`, not `EINVAL`.** The spec reserves EINVAL for a rectangle out of
  bounds. The distinction decides whether a client re-transmits or gives up.
- **`d=f`/`d=F`** reclaims frame memory while keeping the picture on screen, which is the verb's
  purpose; **`d=i`/`d=I` with no `i=`** names id 0, which no image can have, and so deletes nothing.

## Measured result: no change for notcurses, and why

**This does not by itself speed up `notcurses-demo`**, and I have not claimed otherwise anywhere in
the branch. notcurses selects the animated path by *terminal identity and version*, not by
capability — `apply_contour_heuristics()` never calls `setup_kitty_bitmaps()`, so contour reaches
kitty graphics through the generic `a=q` probe and lands in the static path. Filed upstream as
https://github.com/dankamongmen/notcurses/issues/2941. This PR is the precondition for that change,
not a substitute: without it, notcurses recognising contour would produce frames we reject.

## Testing

- `ctest --test-dir out/clang-release -E e2e` — 19/19, **125,137 assertions across 1,645 cases**
- `[kitty]` under ASan/UBSan — 54 cases, which matters because the compositing does raw pointer
  arithmetic

The suite grew from 26 `[kitty]` cases to 54. Beyond the happy paths, the added cases cover what four
review rounds found: alpha blending (including a faint source over an opaque base, where a rounding
mismatch used to wrap a near-white result to black), a frame transmitted without `r=`, editing the
frame that is on screen, editing the root frame twice, repeated `c=` on a composed frame, a frame
advance after the screen scrolls, an advance after the screen is erased, frame deletion, a delete
naming no id, an unimplemented delete target, format and compression mismatches, a frame number
beyond the next, and a hyperlink opened after placement.

Refs #2108


christianparpart (dismissed):

Yaraslaut:
Converting to draft. A third review round (max effort) reproduced eight defects, several of which are
design-level rather than local bugs. Recording them here so the work is not lost.

**The normal spelling of the feature is broken.** `a=f` without `r=` — which kitty documents as "a new
frame is created", and the natural streaming form — *destroys the base frame*. `renderKittyImage()`
default-inserts an empty `_kittyAnimations` entry through `operator[]`, so `existingFrames` is 0 rather
than 1 and `target` becomes 1, overwriting frame 1 with the delta. The following `a=a,c=2` then answers
`EINVAL:no such frame` and the image is gone. Behaviour depends on whether the image was ever
*displayed*, because `a=t` seeds no entry and `a=T` does. Every test I wrote passes an explicit `r=2`,
so none of them catches it.

**`a=d,d=F` destroys every image in the terminal.** The spec defines `d=f`/`d=F` as "delete animation
frames"; it falls into `deleteKittyGraphics()`'s catch-all, which runs `_kittyImages.clear()` and wipes
every placement on the page — and still replies OK. `d=n/N`, `d=p/P`, `d=c/C`, `d=z/Z` and `d=r/R`
share that branch. This target only became reachable because this diff makes frames real.

**Editing the live frame in place is invisible.** The lowest-bandwidth loop the protocol offers — keep
editing the current frame, no further `a=a` — replies OK and changes nothing on screen, because
replacing `frames[target-1]` never refreshes the placement. It also desynchronises `placedImage` from
the current frame, so `placementIsStillOnScreen()` answers about the wrong Image.

**Deletes stop working once a frame is current.** The cells then hold the frame Image, not the base, so
`removeKittyPlacements()`'s identity test skips them: `a=d,d=i` leaves the image on screen, and `d=I`
erases both maps while the grid keeps the frame alive — outside `storedKittyBytes()` and outside every
later delete. A transmit/frame/advance/delete loop grows resident memory without bound, which is
exactly what `MaxStoredImageBytes` exists to prevent.

Also reproduced: `a=f,r=128` fabricates 126 alias frames in one escape and exhausts
`MaxAnimationFrames` while the byte quota sees nothing; `placementIsStillOnScreen()` never checks
`ImageFragment::offset()`, so a partial scroll still stamps frames over text — the corruption its own
comment claims to prevent; `frameAt()` short-circuits frame 1 to the stored image, so reads and writes
of frame 1 permanently disagree; and `a=p` re-places the root frame rather than the current one.

Two further points worth separating from the bug list:

- **`a=a` before any `a=f` must be legal.** In kitty every transmitted image already has a root frame,
  and the spec's own example sets its gap with `a=a,i=7,r=3,z=48`. Contour answers ENOENT, and my test
  `control_without_frames_is_ENOENT` locks that divergence in — a test asserting the wrong contract is
  worse than no test.
- **`s=2`/`s=3` returning ENOTSUP short-circuits `c=`**, so `a=a,c=2,s=3` does neither half. It also
  moves the client's fallback signal to the wrong command: `a=f` now succeeds, so a client uploads
  every frame before the one command that displays them refuses.

**Assessment.** These are not independent bugs to patch one at a time. They share a root: I bolted
frames onto a model where a placement *is* the fragments in the cells and images are immutable, so
there is no object that owns "this placement shows frame N of image M". Every defect above is a
consequence — the delete path, the quota, the redraw guard, the root-frame aliasing. The previous round
demonstrated the risk of patching around it: my fix for a blending finding introduced a worse bug than
the one it corrected.

Recommend redesigning around a placement that references an animation, rather than continuing to fix
in place. The decode layer (the first commit) is sound and independently useful.

Refs #2108

metainfo.xml
```diff
@@ -107,6 +107,7 @@
     <release version="0.7.1" urgency="medium" type="development">
       <description>
         <ul>
+          <li>Adds the kitty graphics animation protocol: <code>a=f</code> transmits a frame, compositing the rectangle it carries onto the frame being edited or onto a copy of the one <code>c=</code> names, and <code>a=a</code> selects which frame is shown. Previously both were parsed and then declined, so a client that wanted to send frame deltas re-transmitted the whole image every frame instead — 5.80 GiB against kitty's 393 MiB for the same 553 images of one <code>notcurses-demo</code> run. Frames advance only when the application selects one: self-running playback (<code>a=a,s=2</code>/<code>s=3</code>) and frame composition (<code>a=c</code>) report ENOTSUP, and the loop count and per-frame gap are recorded but not yet acted on (#2108)</li>
```
christianparpart, line 110: changelog entry is waaaay too long.

christianparpart (changes requested):
as much as I am not in favor of the spec at all, it is what it is, and now, here we are - i see the greater good, I guess :-)

But please adapt the changelog. that's not easily understandable with a wall of text.

PR #2111, opened 2026-09-17
Fix both documented ways of running a command reporting an error

### LINKED ISSUE #2102: executing a specific command has no error free path
### Contour Terminal version

0.7.0-master-e4241cdc

### Installer source

GitHub: Source code

### Operating System

Arch Linux

### Architecture

x86-64 / amd64

### Shell

tcsh

### GPU & Renderer

_No response_

### Other Software

_No response_

### Steps to reproduce

Run contour as `contour terminal execute echo hello`
Note this works but you get an error:
```
[error] Deprecated option "execute" used. Only supported for compatibility with very old KDE desktops.
```
Also note that the resulting terminal gets the:
```
Shell terminated too quickly.
The window will not be closed automatically.
```
message twice.

Run contour as `contour terminal echo hello`
Note this works but you get an error:
```
[error] Do not know what to do with `echo` will use it as a program
```
and still the double message about terminating too quickly.

### Expected Behavior

There should be a way of telling contour which command you want to run in your new window without it reporting an error.

Or maybe the help should make how to do this more obvious in case its user error ;-)

### Actual Behavior

both documented paths produce an error message

### Configuration excerpt

```yaml

```

### Additional notes

not serious - it works so this is just cosmetic

Yaraslaut:
`contour terminal echo hello` — the documented way to run a command in a new window — answered with `[error] Do not know what to do with \`echo\` will use it as a program`, and then showed "Shell terminated too quickly." twice. Both are cosmetic, and both are the terminal telling the user something untrue.

## Changes

- The leading token of `contour [terminal] PROGRAM ARGS...` is now classified by asking the same question the spawn asks. It was probed on disk only (`fs::exists` + `is_regular_file`/`is_directory`), while the spawn resolves it through `execvp()`, which searches `$PATH` — so any program named the way programs are normally named fell into the else-branch that reports it. The PATH question goes through `platform::isReachableProgram()`, which the external launcher already asked for exactly this (and which carries the Windows `%PATHEXT%` nuance); it moves to `ExternalLauncher.hpp` beside `isOpenable()`, for the same reason that one is shared. Only a token that is nowhere is still reported.
- The decision itself is extracted as the pure, injectable `classifyVerbatimCommand()` rather than an `if`/`else` ladder inside `loadConfig()`, so it is unit-testable against a temporary directory and a stub PATH.
- `TerminalSession::onClosed()` is idempotent for an armed notice. One exit legitimately reaches it from two callers — the session's exit watcher, and `TerminalDisplay`'s check for a device already closed when the first frame was about to be drawn, which is the norm for a command that exits immediately — and the existing `_onClosedHandled` de-duplication sits *below* the notice branches' own `return`, so each caller wrote the notice. This also closes the same hole for a device that never started, whose notice `reportDeviceStartFailure()` has already written.

Fixes #2102


Yaraslaut:
Addressing the two review comments.

### `ContourGuiApp.cpp:137` — `owner_exec` is not "this user may execute it" — **accepted, fixed**

Correct, and it was worse than the comment says: on Windows `std::filesystem` reports *every* readable file as owner-executable, so `contour terminal notes.txt` was run as a program instead of opening a shell beside the file. That is what the Windows CI job was failing on.

The check now asks Qt instead, which tests the exec bit against the current user on POSIX and the suffix against `%PATHEXT%` on Windows — the same answer `platform::isReachableProgram()` already gives for a token naming a path.

The suggested *mechanism* — routing regular files through the injected reachability predicate — is not safe, and the new test pins why. `platform::isReachableProgram()` sends a token carrying no separator to `QStandardPaths::findExecutable()`, which searches `PATH` and **not** the current directory (measured: with an executable `probe.sh` in the cwd, `findExecutable("probe.sh")` returns `""` while `QFileInfo("probe.sh").isExecutable()` returns true). So `contour terminal run.sh`, run from the directory holding `run.sh`, would open a shell beside the script instead of running it. Patching the branch to the suggested form makes both the new test and the existing absolute-path one fail.

On the requested non-owner-executable test: it cannot be written portably. POSIX checks the owner class first, so for a file the test process owns, the group and other exec bits grant it nothing — a fixture would need a second uid. The new predicate handles that case correctly by construction, since it asks whether *this user* may execute the file rather than reading one bit.

### `ContourGuiApp.cpp:121` — sandbox PATH vs host PATH under Flatpak — **valid, filed as #2112**

The mechanism is right: with `escape_sandbox: true` the spawn prepends `flatpak-spawn --host` and the program resolves on the host, while the probe asks the sandbox.

The consequence is smaller than stated, though. `Unknown` does not refuse the command — it logs and runs the token as a program anyway, so a host-only program still starts correctly and the only damage is one untrue line in the error log. This PR narrows that message rather than introducing it.

Answering the right question means probing the host (`flatpak-spawn --host command -v …`) during CLI parsing and threading `escape_sandbox` into a predicate that has no notion of profiles today — a design decision needing a real Flatpak build to verify, so it is tracked separately as #2112 rather than widening this PR.

